Issue #8, opened 2026-08-12
Codex CLI is not detected when installed via NVM on macOS


dxcore35:
### Environment

- OpenMausBot 0.1.3
- macOS
- Node.js 22.21.1 managed by NVM
- `@openai/codex` 0.147.0, installed with `npm i -g @openai/codex`

### Steps to reproduce

1. Install Codex with NVM’s Node: `npm i -g @openai/codex`.
2. Confirm it works in Terminal: `codex --version` returns `codex-cli 0.147.0`.
3. Open OpenMausBot and select Codex.

### Actual behavior

OpenMausBot reports `` `codex` CLI not found ``.

The OpenMausBot main process is launched with only:

```text
PATH=/usr/bin:/bin:/usr/sbin:/sbin
```

That excludes NVM’s global-bin directory (`~/.nvm/versions/node/v22.21.1/bin`), so the Codex driver’s default `codex` command cannot be resolved.

### Expected behavior

A working global Codex CLI should be detected, including when Node is installed through NVM.

### Workaround

I was able to make it work by setting the Codex instance `cli` value in `~/.openmausbot/config.json` to an absolute, executable wrapper that invokes both the NVM Node binary and Codex’s `bin/codex.js` by absolute path. Simply setting `cli` to the NVM `codex` script is insufficient because that script uses `#!/usr/bin/env node`, and OpenMausBot’s PATH also cannot resolve `node`.

### Suggested fix

For child CLI processes, include common user-managed Node locations (or obtain the user’s login-shell PATH), and/or provide a documented UI setting for an absolute Codex executable path. An absolute CLI path should also work when its Node runtime is NVM-managed.

milind-soni:
Fixed on main in 547ebfa — same root cause as #12, with an nvm twist: nvm only exists in interactive shells, so a Finder-launched app never sees `~/.nvm/versions/node/*/bin` where your npm-installed `codex` lives.

The harness now explicitly includes every nvm node version's bin dir (newest node first) plus your login shell's real `PATH` when locating agent CLIs — verified on an nvm machine. Ships in v0.1.5; until then `pnpm dev:server` from a terminal works, or set `OMB_EXTRA_PATH=$(dirname $(which codex))` before launching. Leaving open until the fixed build is published.

milind-soni:
[...]

PR #14, opened 2026-08-12
Find the user's CLIs: resolve the login shell PATH at harness startup

### LINKED ISSUE #12: `claude` CLI not found, macOs
### Before filing

### Area

Chat UI (src/)

### What happened

`claude` CLI not found, macOs

### Version and install

v0.1.3 (0.1.3) 

### Logs

```text

```

alianrock:
## What changed

`server/login-path.ts` (new) asks the login shell for the PATH it would give a terminal, and `server/index.ts` merges that into `process.env.PATH` once at startup — before `registry.load()` probes any driver.

- `-l -i -c` because the profile files hold Homebrew and the rc files hold nvm/pyenv, and only interactive shells read the latter.
- The answer comes back between sentinels, not as "the output" — rc files print whatever they like.
- It reads `printenv PATH` rather than `$PATH`, so shells with a list-valued PATH (fish) still return a delimiter-joined string.
- 5s timeout; if the shell can't answer, a short list of common install dirs (`~/.local/bin`, `/opt/homebrew/bin`, `~/.bun/bin`, `~/Library/pnpm`, …) stands in.
- The merge only appends and dedupes, so PATH can never come back smaller than it started. `ensureUserPath()` never throws.
- Windows returns early — it inherits a usable PATH from the shell that started it, and has no `$SHELL -lic` to ask.

No `electron/` changes: in the packaged app the harness is the utilityProcess that owns every CLI spawn, so fixing PATH there covers `claude`, `codex` and `grokAgent` at once, and keeps this off the files the in-flight Windows ports are editing.

## Why

A packaged app is launched by launchd, not by a shell, so it inherits `PATH=/usr/bin:/bin:/usr/sbin:/sbin`. Every driver spawns its CLI by bare name — `spawn(config.cli, …)` with `{ ...process.env }` — so a working install in `~/.local/bin`, Homebrew, nvm, pnpm or bun is invisible to the harness. The probe then reports ``​`claude` CLI not found`` on a machine where `claude` runs fine in Terminal.

Fixes #12
Refs #8 — the NVM report is the same root cause (that one also needs `node` on PATH for the `#!/usr/bin/env node` shim, which the login PATH supplies).

## How it was verified

macOS 15 (arm64), Node 24, `claude` 2.1.224 installed at `~/.local/bin/claude`.

`pnpm typecheck` and `pnpm test` pass (62 passed, 1 skipped — the Windows-only case).

Reproduced the exact launchd environment and compared before/after:

```
env -i HOME="$HOME" SHELL=/bin/zsh PATH=/usr/bin:/bin:/usr/sbin:/sbin OMB_PORT=8801 node server/index.ts
curl -s http://127.0.0.1:8801/api/instances
```

before (`main`):

```
claude   unavailable  `claude` CLI not found
```

after:

```
claude   available    2.1.224 (Claude Code)
```

`grok` and `codex` correctly stay `unavailable` — they really aren't installed on this machine.

Probe cost measured at **177 ms**, once per harness boot.

Also ran the probe script directly against every shell on the box:

| shell | result |
| --- | --- |
| zsh, bash, dash, ksh, sh | returns a delimiter-joined PATH |
| tcsh, csh | no sentinels → falls back |

`-i` earns its place: zsh returns 8 directories `-l` alone does not, including `/opt/homebrew/bin` and `~/.npm-global/bin` — where most people's `claude` and `codex` actually live.

### Known limit

When the shell can't answer (csh/tcsh, or an rc file slower than the 5s timeout) the static fallback covers `~/.local/bin`, Homebrew, bun, pnpm and volta, but not the versioned directories of nvm/asdf/mise. A csh user whose CLI came from nvm is still stuck. Resolving those would mean directory scanning plus a "which version" policy per version manager, which felt well above the altitude of this fix — happy to add it if you'd rather cover that case.

New tests in `server/login-path.test.ts` cover the merge (appends, dedupes, never loses an existing entry, idempotent), sentinel extraction out of chatty rc output, the fallback path when the shell throws or times out, the Windows early return, and one case that runs the probe against the host's real login shell — the whole fix rests on that actually working.

## Screenshots (UI changes)

None — server-side only.

## Checklist

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -14,6 +14,7 @@ import type { RuntimeEvent } from "./contracts.ts";
 import { BUILT_IN_DRIVERS } from "./drivers/builtIn.ts";
 import { EventBus } from "./harness/bus.ts";
 import { ProviderRegistry } from "./harness/registry.ts";
+import { ensureUserPath } from "./login-path.ts";
 import { Store, type Message } from "./store.ts";
 
 const PORT = Number(process.env.OMB_PORT || process.env.OGB_PORT || 8799);
@@ -29,6 +30,9 @@ const MIME: Record<string, string> = {
   ".woff2": "font/woff2",
 };
 
+// before anything probes or spawns a CLI: the packaged app is launched by
+// launchd, so without this the drivers only ever see /usr/bin:/bin:/usr/sbin:/sbin
+await ensureUserPath();
 ensureDirs();
 const cfg = loadConfig();
 const registry = new ProviderRegistry(BUILT_IN_DRIVERS);
--- a/server/login-path.test.ts
+++ b/server/login-path.test.ts
@@ -1,0 +1,115 @@
+// The bug this guards against is silent and total: with the launchd PATH
+// every driver reports its CLI missing on a machine where that CLI runs
+// fine in Terminal. So the contract is (a) the merge never loses an entry
+// the process already had, (b) a shell that hangs, dies or chatters can't
+// take the harness down or leave PATH worse than it was, and (c) the
+// probe really does read the login shell — proven against a real one.
+import { delimiter } from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { ensureUserPath, extractPath, loginShellPath, mergePath, type ProbeRunner } from "./login-path.ts";
+
+const LAUNCHD_PATH = ["/usr/bin", "/bin", "/usr/sbin", "/sbin"].join(delimiter);
+const posixOnly = process.platform === "win32" ? it.skip : it;
+
+const originalPath = process.env.PATH;
+afterEach(() => {
+  process.env.PATH = originalPath;
+});
+
+/** A shell that answers with `path`, wrapped in the noise rc files print. */
+const fakeShell =
+  (path: string): ProbeRunner =>
+  async () =>
+    `Welcome to your shell!\n__OMB_PATH_BEGIN__${path}\n__OMB_PATH_END__`;
+
+describe("mergePath", () => {
+  it("appends the extra entries and keeps the current ones first", () => {
+    expect(mergePath("/usr/bin:/bin", "/opt/homebrew/bin:/Users/x/.local/bin")).toBe(
+      "/usr/bin:/bin:/opt/homebrew/bin:/Users/x/.local/bin",
+    );
+  });
+
+  it("drops duplicates and empty segments", () => {
+    expect(mergePath("/usr/bin::/bin", "/bin:/usr/bin:/opt/homebrew/bin:")).toBe("/usr/bin:/bin:/opt/homebrew/bin");
+  });
+
+  it("survives an empty side", () => {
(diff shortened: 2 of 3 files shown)


milind-soni:
This is a genuinely well-built fix — and I owe you an apology for the awkward timing: a fix for the same root cause shipped in [v0.1.5](https://github.com/milind-soni/openmausbot-releases/releases/tag/v0.1.5) (`server/env-path.ts`, commit 547ebfa) right around the time you opened this, which is why #12 closed under you mid-flight. Parallel invention — you diagnosed launchd's bare PATH exactly right.

Comparing the two, several of your pieces are better than what shipped, and I'd like them in the tree:

- **Awaiting the shell before the first driver probe** — env-path.ts probes in the background, so the very first model-picker paint can race the shell answer; your `await ensureUserPath()` before `registry.load()` makes it deterministic.
- **`printenv PATH` between BEGIN/END sentinels** — fish-safe and immune to chatty rc files (env-path matches to end-of-line, which is weaker).
- **The injectable `ProbeRunner`** — your tests exercise hang/noise/failure cases mine can't reach.

What env-path.ts has that this PR doesn't: the synchronous well-known-dirs pass (works even when the shell probe fails or times out — covers nvm's every-version bin dirs newest-first, volta, bun, asdf), and the `OMB_EXTRA_PATH` escape hatch.

Rather than shipping two overlapping PATH systems, would you be up for rebasing this as a refactor **of** `env-path.ts` that folds in your three improvements (deterministic await at startup, sentinel/printenv probe, injectable runner) while keeping the sync known-dirs + `OMB_EXTRA_PATH`? That version I'd merge fast — CI on three OSes will hold you honest. If you'd rather not do another round, say so and I'll do the consolidation myself with a `Co-authored-by:` credit — your call, but either way this work is landing, not closing.

milind-soni (approved):

milind-soni (commented):
Current main already has server/env-path.ts wired through every Claude/Codex/ACP spawn, with known install directories, nvm discovery, a login-shell probe, and tests for issues #8/#12. This PR now duplicates that implementation and would add up to a 5s startup block. I am leaving it unmerged as superseded by the existing env-path solution.

milind-soni:
Thank you — you identified a real bug: an app launched from the GUI gets a bare PATH and can't find CLIs installed by nvm, Homebrew, volta and friends.

Closing as superseded: this is fixed on main in `server/env-path.ts`. `augmentedPath()` merges `OMB_EXTRA_PATH`, the inherited PATH, and the usual install locations (`~/.local/bin`, `~/.claude/local`, Homebrew, volta, bun, asdf, deno, nvm versions newest-first), and — the part your PR was really about — it runs a login-shell probe in the background (`env-path.ts:107-116`: `$SHELL -l -i -c 'printf "__OMB_PATH__%s" "$PATH"'`) and folds the result into the cache for the next spawn, without ever blocking a turn. Every driver spawn and every `--version` probe goes through it, and `server/env-path.test.ts` pins the behaviour.

Sorry it sat while a fix for the same problem landed separately. If your version handles a case main misses, a small PR against current main would be welcome.

PR #21, opened 2026-08-12
Reliability, privacy & CI hardening

### LINKED ISSUE #20: Reliability, privacy & CI hardening
While reading through the codebase I noticed a cluster of reliability, privacy and CI hardening items that are cheap to fix and independent of any feature work. Grouping them here; a PR is on the way.

### Reliability / correctness
- [ ] **Non-atomic persistence → silent data loss.** `bots.json`, `messages-*.json` and `config.json` are written straight onto the final file (`server/store.ts`, `server/config.ts`). An interrupted write truncates the JSON, and the next boot silently treats the parse failure as empty state. Fix: temp-file + `fsync` + `rename`.
- [ ] **`readBody` keeps buffering past the limit.** Over 1 MB it rejects but keeps concatenating the incoming body (`server/index.ts`); it should stop buffering and return `413` (and `400`, not `500`, on invalid JSON).
- [ ] **Bot can stay `busy` forever after a config reload.** `reloadProviders` disposes in-flight turns whose completion handlers are already detached (`server/index.ts`), leaving the bot stuck. Clear `busy` the way a restart does.
- [ ] **UTF-8 corruption on stream boundaries.** Agent stdout is accumulated with `buf += chunk` (`claude.ts`, `codex.ts`, `acp/core.ts`), which splits multibyte characters across reads. Fix: `stdout.setEncoding("utf8")`.
- [ ] **Message-map key collision between concurrent bots.** `toolMessageByItem` / `askMessageByRequest` key on the provider item/request id alone (`server/index.ts`); two bots acting at once can collide and patch each other's messages. Scope the keys by `threadId`.
- [ ] **No timeout on codex JSON-RPC.** A wedged `app-server` can accept stdin and never reply (`server/drivers/codex.ts`), hanging the turn with the bot busy. Add a request timeout.
- [ ] **`@mention` has no trailing word boundary.** `@New Bottle` matches a bot named `New Bot` (`server/store.ts`).
- [ ] **Stale closure in the OAuth poll.** `PluginsPanel` reads `status[slug]` from the closure captured at connect time, so it always polls the full 6 times (`src/components/PluginsPanel.tsx`).

### Privacy
- [ ] **PostHog `autocapture` leaks conversation fragments.** Autocapture ships the `$el_text` of clicked elements; the sidebar and option cards render model output and message previews, so fragments of private conversations go to a third party (`src/lib/analytics.ts`). Turn autocapture off, keep only the named product events.

### CI hardening
- [ ] **Mutable action tags + no minimum permissions.** `ci.yml` uses `actions/checkout@v4` etc. and declares no `permissions`. Pin to commit SHAs and add `permissions: contents: read`.

Happy to split these into separate PRs if you'd prefer.

KesleyDavid:
Closes #20.

A batch of reliability, privacy and CI hardening fixes found while reading through the code. All independent of feature work; typecheck is clean and the full suite passes (87 tests, +5 new).

Reviewable as three commits:

### `reliability:` atomic persistence, bounded requests, and lifecycle fixes
- **Atomic writes** — `bots.json` / `messages-*.json` / `config.json` now go through a temp-file + `fsync` + `rename` helper (`server/atomic.ts`), so an interrupted write can never truncate the JSON into silent empty state on next boot.
- **`readBody`** stops buffering and returns `413` past 1 MB (it previously kept concatenating after rejecting), and returns `400` instead of `500` on invalid JSON.
- **`reloadProviders`** clears `busy` on bots caught mid-turn, which `disposeAll` would otherwise strand busy forever.
- **UTF-8** — agent stdout is decoded across chunk boundaries with `setEncoding("utf8")` (claude/codex/acp), fixing corruption of multibyte characters that straddle two reads.
- **Message-map keys** scoped by `threadId` so two bots acting at once can't patch each other's messages.
- **codex JSON-RPC timeout** (60s) so a wedged `app-server` can't hang the turn.
- **`@mention`** requires a word boundary at the end of a name (`@New Bottle` no longer matches `New Bot`).
- **`PluginsPanel`** polls OAuth status off the fresh fetch result instead of a stale closure.
- **`perm:open-settings`** uses an own-property lookup for the pane key.

### `privacy:` disable PostHog autocapture
Autocapture ships the `$el_text` of clicked elements; the sidebar and option cards render model output and message previews, so it would send fragments of private conversations to a third party. Named product events are unchanged.

### `ci:` pin actions to SHAs + least-privilege permissions
Pin `checkout` / `action-setup` / `setup-node` to their v4.4.0 commit and add `permissions: contents: read`.

Tests: `server/atomic.test.ts` (new) and an added `@mention` boundary case. Happy to split into separate PRs if you prefer, and to rebase around the in-flight Windows PRs since a few of these touch the same files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,10 @@ on:
     branches: [main]
   pull_request:
 
+# least privilege: CI only reads the checkout, it never writes back
+permissions:
+  contents: read
+
 concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
@@ -25,9 +29,9 @@ jobs:
       # tests never launch Electron; skip the binary download
       ELECTRON_SKIP_BINARY_DOWNLOAD: "1"
     steps:
-      - uses: actions/checkout@v4
-      - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v4.4.0
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: 24
           cache: pnpm
--- a/electron/main.mjs
+++ b/electron/main.mjs
@@ -151,9 +151,10 @@ ipcMain.handle("perm:open-settings", (_event, pane) => {
     screen: "Privacy_ScreenCapture",
     speech: "Privacy_SpeechRecognition",
   };
-  return shell.openExternal(
-    `x-apple.systempreferences:com.apple.preference.security?${panes[pane] ?? "Privacy"}`,
-  );
+  // own-property lookup only — a renderer-supplied "__proto__"/"constructor"
+  // would otherwise resolve up the prototype chain to a truthy object
+  const anchor = Object.hasOwn(panes, pane) ? panes[pane] : "Privacy";
+  return shell.openExternal(`x-apple.systempreferences:com.apple.preference.security?${anchor}`);
 });
 
 ipcMain.handle("speech:start", (event) => {
--- a/server/atomic.test.ts
+++ b/server/atomic.test.ts
@@ -1,0 +1,45 @@
+import { existsSync, mkdtempSync, readdirSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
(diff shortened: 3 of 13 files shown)


milind-soni (changes requested):
There are several worthwhile hardening fixes here, but this branch now conflicts with current main and is too broad to merge safely as one unit. Please rebase and split it into focused PRs: atomic state/config writes; thread-scoped item/request keys; Codex RPC timeout plus UTF-8 stream decoding; and analytics/CI privacy hardening. Preserve the newer provider, branching, Linux, and process-lifecycle work while rebasing, and add focused regression coverage for the request-size error path and Codex timeout.

milind-soni (approved):
Approved after integrating this onto current main and resolving the two conflicts. During integration I also fixed malformed JSON handling (the original done flag could leave invalid requests unresolved), kept the newer provider-reload interruption cleanup, made atomic temp names collision-safe, and added regression coverage. The integrated tree passes typecheck, all 178 tests, Electron syntax validation, and the production build.

PR #33, opened 2026-08-12
Fit the app icon to the macOS icon grid

guilimasp:
The icon tile fills the whole 1024 canvas, so in the Dock it renders visibly larger than every other app. macOS draws app icons on an 824×824 body inside the 1024 canvas, with a 185.4 corner radius.

before and after

- `build/icon.svg` — scale the tile into the grid; the rasters (`icon-1024.png`, `icon.iconset`, `icon.icns`) are regenerated from it
- `electron/resources/app-icon.png` — same treatment, because `app.dock.setIcon()` applies it at runtime and it overrides the bundle `.icns`, so fixing only the `.icns` leaves the Dock unchanged

The artwork itself is unchanged. `pnpm typecheck && pnpm test` pass.


milind-soni (approved):
Reviewed source and regenerated assets: 824x824 macOS grid scaling is correct, all iconset resolutions and ICNS are valid, runtime Dock PNG matches, and Windows ICO is untouched.

PR #35, opened 2026-08-12
Dictate in the user's language, not always en-US

guilimasp:
`speech-helper.swift` pins the recognizer to `en-US`, so dictation only works for English speakers — everyone else gets their speech transcribed into English words that sound roughly similar.

It now walks `Locale.preferredLanguages` and uses the first locale with an available recognizer, falling back to `en-US`. A locale the system has no model for reports `isAvailable == false`, so this can only match today's behavior or improve on it.

One caveat: the system language is not always the language someone dictates in, so a per-user setting would cover that case properly. 

`pnpm typecheck && pnpm test` pass, and the helper compiles with `swiftc`.

### DIFF
--- a/electron/resources/speech-helper.swift
+++ b/electron/resources/speech-helper.swift
@@ -24,8 +24,15 @@ func fail(_ message: String) -> Never {
 
 SFSpeechRecognizer.requestAuthorization { status in
   guard status == .authorized else { fail("speech-not-authorized") }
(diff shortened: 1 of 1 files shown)


milind-soni (approved):
Reviewed against current main: locale selection preserves en-US fallback. Swift typecheck, project typecheck, and all 159 tests pass in the combined tree.

PR #36, opened 2026-08-12
Tell the agent about the connected apps in the system prompt

guilimasp:
When a Composio key is configured the harness mounts the Connect MCP server and pre-allows its tools, but the persona never mentions it — the computer and agents integrations each get a line, Composio gets none.

The result is that the model does not link a request like "can you read my email?" to the `mcp__composio__*` tools it is holding. It answers that it has no access, or starts hunting for an unrelated integration to authorize, while a working Gmail connection sits one tool call away.

Asked the same question before and after this line: before, the bot said it had no access and suggested authorizing a different connector; after, it searched the toolkits, ran the tool and answered from the mailbox.

One line, in the same shape as the two beside it. `pnpm typecheck && pnpm test` pass.

## Summary by CodeRabbit

* **New Features**
  * Connected Composio apps are now available through supported providers as HTTP MCP integrations.
  * Supported providers advertise whether Composio integrations are available.
  * Connected-app guidance is shown only when the integration is active.

* **Bug Fixes**
  * Prevented Composio integrations from being mounted for providers that do not support them.

* **Tests**
  * Added coverage for integration mounting, authorization, API key forwarding, and capability reporting.


milind-soni (changes requested):
The intent is good, but this hint is currently gated only on integrations.composio. startTurn sets that integration for every provider whenever a key exists, while server/drivers/claude.ts is the only driver that mounts the Composio MCP server; Codex, Grok, and ACP drivers ignore it. That means this change would tell non-Claude bots to call tools they do not have. Please gate both the integration/prompt through an explicit capability (preferred), or at minimum restrict this prompt to the Claude driver until other drivers mount Composio.

guilimasp:
You are right, and thank you — gating on `cfg.composio.key` would have told Codex, Grok and every ACP bot to call tools they were never handed. Took the preferred route.

- New `capabilities.composioMcp`, documented next to `agentsMcp` and `computerMcp` and following the same rule: the flag means *this driver mounts the integration*, not *the user has connections*.
- `claude.ts` declares it (it is the only driver that mounts the Connect MCP server today). Nothing else does.
- `startTurn` now gates `integrations.composio` on the capability, so a non-Claude bot gets neither the tools nor the hint. The prompt line hangs off `integrations.composio` rather than the key, the way the `agents` hint already hangs off `integrations.agents` — so the two can never drift apart.

Tests: `claude.test.ts` asserts it both mounts the server (pre-allowing `mcp__composio`) and claims the capability; `acp.test.ts` asserts ACP does not claim what it cannot honour. Any future driver that mounts Composio flips one flag and picks up the prompt with it.

`pnpm typecheck && pnpm test` pass. Rebased onto current main.

milind-soni (approved):
The requested provider-capability gating is now implemented: Composio configuration and prompt guidance are exposed only through adapters that declare composioMcp support. I refreshed the branch on current main while preserving both ACP/Droid test coverage, then verified the full test/build suite and the hosted macOS, Ubuntu, Windows, and Linux package-smoke matrix.

milind-soni:
Heads-up after this merged: the prompt line named `COMPOSIO_EXECUTE_TOOL`, which doesn't exist on the endpoint the driver mounts. I checked `tools/list` against connect.composio.dev/mcp with the repo's own key — there are seven tools, and calling that name returns `-32602: Tool COMPOSIO_EXECUTE_TOOL not found`. The executor is `COMPOSIO_MULTI_EXECUTE_TOOL`, and arguments come from `COMPOSIO_GET_TOOL_SCHEMAS`, so search→execute alone can't build a valid call.

Fixed in #126 (merged): the hint now says SEARCH_TOOLS → GET_TOOL_SCHEMAS → MULTI_EXECUTE_TOOL. The idea was right — a bot that doesn't know it has connected apps tells the user it has no access — it just needed the real tool names.

PR #38, opened 2026-08-13
Paste images into the composer

guilimasp:
The composer only takes text, so anything visual — a screenshot of a failing build, a chart, a photo of a whiteboard — has to be described in words, or saved to a file and referenced by path, which only works when the bot happens to have filesystem tools.

Pasting an image now stages it as a thumbnail above the input (hover for an ✕ to drop it), sends with or without accompanying text, and renders in the user's own bubble. The Claude driver turns the attachments into Messages-API image blocks on the stream-json prompt; a turn with no images sends the same plain string it does today, and a driver that cannot carry images ignores them and sends the text alone.

Sizing happens before anything leaves the browser: 1568px on the long edge (past that the providers downscale anyway), re-encoded as PNG, or JPEG when the PNG would be large — GIFs pass through untouched so animation survives. `server/attachments.ts` re-validates independently of the client (known image types, at most 6 per message, 5MB of base64 each) and drops what fails rather than throwing away the user's message.

**Composer, before and after a paste:**

composer before

composer after

**The turn itself — pasted chart, and the bot reading it:**

conversation

New behavior is covered by `server/attachments.test.ts` (wire validation and the block shape) and `server/drivers/claude-images.test.ts` (the driver contract against the scripted fake CLI, `skipIf(win32)` like its neighbours). `pnpm typecheck && pnpm test` pass, and there are no new dependencies.

Two things deliberately left out, happy to add either if you'd rather they ship together: the `+` button is still inert — this is only about paste — and only the Claude driver forwards images today.

### DIFF
--- a/server/attachments.test.ts
+++ b/server/attachments.test.ts
@@ -1,0 +1,48 @@
+// Pasted images: what the wire is allowed to carry, and what the Claude CLI
+// is handed for it.
+import { describe, expect, it } from "vitest";
+
+import { sanitizeImages, userContent } from "./attachments.ts";
+
+const png = { mime: "image/png", data: "aGVsbG8=" };
+
+describe("sanitizeImages", () => {
+  it("keeps well-formed images", () => {
+    expect(sanitizeImages([png])).toEqual([png]);
+  });
+
+  it("drops anything that is not a supported image", () => {
+    expect(sanitizeImages([{ mime: "application/pdf", data: "aGk=" }])).toEqual([]);
+    expect(sanitizeImages([{ mime: "image/png", data: "not base64!" }])).toEqual([]);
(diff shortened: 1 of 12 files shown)


guilimasp:
Closing this one — it went conflicting when the chat overhaul landed, and rebasing it means rewriting the image wiring across 12 files against code that has moved a long way. Not worth carrying it red in your queue while it waits.

For files that live on disk, #69 covers the same ground from a different angle: dropping a file on the window attaches it by path, which every driver can open without an attachment protocol. What that does not cover is an image straight from the clipboard, which is what this PR was for.

Happy to redo the clipboard-image path against current `main` if you want it — say the word and I'll open a fresh one.

PR #40, opened 2026-08-13
Support Shift+Enter for newlines in the composer

guilimasp:
## What

The message box was a single-line `<input>`, so there was no way to write a multi-line message. This swaps it for an auto-growing `<textarea>`:

- **Enter** still sends (guarded against IME composition so it doesn't fire mid-composition)
- **Shift+Enter** inserts a newline
- The field grows with its content up to ~160px, then scrolls internally
- The attach/mic buttons anchor to the bottom and the pill corners are relaxed to `rounded-3xl` so the multi-line state looks clean
- The `@mention` picker behavior is unchanged (Enter still picks while it's open)

## Testing

`pnpm typecheck` and `pnpm test` (92 tests) both pass.

https://claude.ai/code/session_01YAPBLjFSFsENaGGBvrkwJJ

## Summary by CodeRabbit

* **New Features**
  * Replaced the single-line message field with a multiline text area.
  * Text area height expands with content up to a maximum size.
  * Use Shift+Enter to add line breaks and Enter to send messages.

### DIFF
--- a/src/components/Composer.tsx
+++ b/src/components/Composer.tsx
@@ -26,7 +26,7 @@ export function Composer({ bot }: { bot: Bot }) {
   const [caret, setCaret] = useState(0);
   const [highlight, setHighlight] = useState(0);
   const [dismissedAt, setDismissedAt] = useState<number | null>(null); // Esc'd this @
-  const inputRef = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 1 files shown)


milind-soni:
LGTM

PR #42, opened 2026-08-13
Windows: bind the permission broker to a named pipe (every approval silently becomes a deny)

E4B-labs:
> **Stacked on #41.** The diff below shows #41's commit until that merges; this PR's own change is the second commit ([`server/drivers/claude.ts` + `claude.test.ts`](https://github.com/milind-soni/OpenMausBot/pull/42/files)). #41 and this PR are the only two in the Windows series that depend on each other, and only for the test unskip — see "Why stacked" at the bottom.

## What's broken

**Every permission prompt on Windows silently ends as a deny.** The user is never asked, nothing appears in the UI, and there is no error anywhere to explain it — the turn just quietly loses the action ~15 minutes later with the timeout note.

## Root cause

`createPermissionBroker` binds a **unix domain socket** in `DATA_DIR` and hands that path to the permission MCP proxy the CLI spawns. Windows has no unix domain sockets, so `server.listen(socketPath)` fails immediately.

And the failure was invisible:

```js
server.on("error", () => {});
```

So: the listen fails, nothing is logged, the proxy cannot connect, the ask never reaches `onAsk`, no `request.opened` event is emitted, and the CLI's request sits until the 15-minute broker timeout fires and resolves it as `deny` with `source: "timeout"`. From the user's side an approval prompt simply never appears and the agent reports it was blocked.

## The fix

`net.createServer` binds a **named pipe** under exactly the same API, both ends included — so `permissionSocketPath()` returns `\\.\pipe\omb-perm-<pid>-<tag>` on win32 and the existing socket path everywhere else. Nothing in the broker or the proxy protocol changes.

The pid is in the pipe name deliberately. The Windows pipe namespace is global and flat, so unlike socket files it is not isolated by `DATA_DIR`; without the pid, two harnesses running concurrently would collide on `omb-perm-<tag>`.

The error listener now logs instead of discarding. A broker that cannot come up is worth one line on stderr — that swallowed handler is the reason this failure mode was invisible rather than obvious, on any platform.

## Tests

`permissionSocketPath` is exported so the test that raises an ask over the socket asks the driver for the path instead of rebuilding it by hand. It then follows the platform automatically rather than hardcoding the POSIX shape.

With #41 applied, the whole Claude driver suite runs unskipped on Windows, including the broker round-trip.

## How this was tested

- `pnpm typecheck` and `pnpm test` on Windows 10, on this branch: **89 passed / 3 skipped**, up from 52 passed / 33 skipped on `main`. The 3 remaining skips are POSIX-only assertions that are correct to skip.
- The broker round-trip test (`brokers a permission ask into request.opened and answers over the socket`) now runs on Windows and passes: proxy connects to the pipe, raises an ask, `request.opened` is emitted, the answer comes back over the same connection.
- macOS/Linux behaviour is unchanged — the `win32` branch is the only new path, and the socket path off win32 is byte-identical to before.

## Why stacked

The Claude driver's tests need a spawnable fake CLI (#41) *and* a bindable broker (this PR) before they can run on Windows — either one alone still fails. Rather than leave the win32 unskip unowned, it rides here, in the later of the two.

The other Windows fixes in this series are independent of both and branch straight from `main`.

### DIFF
--- a/server/comms.test.ts
+++ b/server/comms.test.ts
@@ -6,8 +6,9 @@
 // session/new mcpServers → agents-proxy → /api/internal/ask-bot →
 // askBotAndWait → bus fold. The internal endpoints' auth is pinned too.
 //
-// The fake CLI is a shebang script, so the e2e half is POSIX-only (same
-// gating as acp.test.ts); the mention-resolution unit tests run everywhere.
+// The fake CLI is a shebang script — POSIX-only until resolveCliSpawn
+// turned it into `node <script>` on Windows too, so the e2e half now runs
+// everywhere alongside the mention-resolution units.
 import { spawn, type ChildProcess } from "node:child_process";
 import { chmodSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
@@ -21,7 +22,6 @@ const SERVER_DIR = dirname(fileURLToPath(import.meta.url));
 const FAKE_CLI = join(SERVER_DIR, "testing", "fake-acp-cli.ts");
 const PORT = 18800 + Math.floor(Math.random() * 10_000);
 const BASE = `http://127.0.0.1:${PORT}`;
-const posixOnly = describe.skipIf(process.platform === "win32");
 
 describe("mentionedBots", () => {
   const peers = [
@@ -46,7 +46,7 @@ describe("mentionedBots", () => {
   });
 });
 
-posixOnly("comms e2e (fake ACP fleet)", () => {
+describe("comms e2e (fake ACP fleet)", () => {
   let child: ChildProcess;
   let home: string;
   let stderr = "";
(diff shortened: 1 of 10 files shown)


milind-soni (commented):
Current main already uses server/procs.ts::brokerSocketPath: a unique \\.\pipe\openmausbot-perm-* named pipe on Windows and a filesystem socket on POSIX, and Claude passes that path through the existing broker/proxy flow. This PR is now superseded by the landed implementation.

milind-soni:
The remaining permission-broker fix has been integrated into `main` via #83, after #41/#81. I kept the centralized named-pipe helper, added the harness PID there to prevent cross-instance collisions, surfaced listen failures, and enabled the full Claude approval round-trip suite on Windows. The final integration passed Windows, macOS, Ubuntu, and the packaged Ubuntu smoke test. Thank you for isolating the silent-deny failure and supplying the cross-platform test path.

PR #45, opened 2026-08-13
Add message editing with conversation branches

guilimasp:
## What

Editing a user message forks the conversation, ChatGPT/Claude style. Hover a message for a pencil icon, or press **ArrowUp** in an empty composer to edit your last message. Enter resends, Esc cancels, and edited messages get `‹ 2/2 ›` controls to flip between versions — each version keeps its own continuation, so nothing is lost.

## How

- **Tree model** — messages carry a `parentId` and each thread stores an `activeLeafId`; the visible chat is the root→leaf path. Existing flat `messages-<threadId>.json` files migrate transparently on load (rows chain in array order, last row becomes the leaf).
- **New endpoints** — `POST /api/bots/:id/messages/:msgId/edit` (fork + rerun the turn) and `POST /api/bots/:id/active-branch` (switch versions, no new turn). `GET /api/bots` now includes `activeLeafId`.
- **Provider context after a rewind** — the provider's native session still holds the abandoned branch, so resume cursors are invalidated and the surviving path is replayed inline for cursor-resuming drivers (claude/codex/gemini); the grok driver keeps receiving it via `transcript` as before.

## No double generation, by construction

The classic failure mode of edit-in-place chat UIs is two conversation tails generating at once. This PR prevents it server-side:

- An edit interrupts the in-flight turn and **waits for it to settle** before forking (the timeout outlasts every driver's own cancel window, ACP's 5s included).
- Straggler events from a dead turn are dropped at the bus (turnId fence + busy guard), so they can never fold into the new branch.
- Edits serialize per bot; sending while busy stays 409; switching versions during a live turn is refused.

## Testing

- Store unit tests: fork/switch/persistence/legacy-file migration.
- API tests against the real server: validation and no-mutation on failure.
- New `server/branching.test.ts` e2e on the fake ACP CLI: real turn → edit → reply lands on the new branch → version switching; plus an edit against a **hung** turn proving the interrupt path never runs two turns at once.
- `pnpm typecheck`, `pnpm test`, `pnpm build` all green; also smoke-tested in the browser against the dev server.

https://claude.ai/code/session_01634YWxEfsreuTSoDLCSy7b

### DIFF
--- a/server/branching.test.ts
+++ b/server/branching.test.ts
@@ -173,7 +173,7 @@ posixOnly("conversation branching e2e (fake ACP fleet)", () => {
   );
 
   it(
-    "editing mid-turn interrupts the old turn instead of running two at once",
+    "refuses to rewind a live thread, then edits cleanly once it is stopped",
     async () => {
       const created = (await api("POST", "/api/bots")).body.bot;
       await api("PATCH", `/api/bots/${created.id}`, {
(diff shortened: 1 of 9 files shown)


milind-soni:
Looks great! Thanks for the contribution

PR #53, opened 2026-08-13
feat: add Hermes ACP driver support

longwaywandrer:
- Add Hermes ACP driver (server/drivers/acp/hermes.ts)
- Register HermesDriver in builtIn.ts
- Add hermes instance to default fleet in config.ts
- Bind harness to 0.0.0.0:8799 and Vite to 0.0.0.0:5199 
- Downgrade Node engine requirement >=24 to >=22 for server compat

Verified end-to-end: a Hermes bot spawned hermes-acp and replied through the harness

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally
- [ ] Server behavior changes come with tests (see CONTRIBUTING.md → Tests)
- [ ] No `dist-server/` edits (it's build output)
- [ ] macOS-only code is platform-gated; no `shell: true` / cmd.exe string-building
- [ ] No secrets in logs, responses, events, or argv

## Summary by CodeRabbit

- **New Features**
  - Added support for the Hermes assistant, including its default model and authentication flow.
  - Included Hermes in the default available instance configurations.

- **Accessibility**
  - The application and development server can now be accessed through all network interfaces.
  - Startup logs display the accessible network address.

- **Compatibility**
  - Lowered the minimum supported Node.js version to 22.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "main": "electron/main.mjs",
   "engines": {
-    "node": ">=24"
+    "node": ">=22"
   },
   "packageManager": "pnpm@10.33.0",
   "scripts": {
--- a/server/config.ts
+++ b/server/config.ts
@@ -91,6 +91,7 @@ export function instanceConfigs(cfg: AppConfig): InstanceConfigMap {
           claude: { driver: "claudeAgent" },
           codex: { driver: "codex" },
           computer: { driver: "boxAgent" },
+          hermes: { driver: "hermes" },
         };
   for (const entry of Object.values(map)) {
     entry.environment = {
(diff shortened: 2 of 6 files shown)


milind-soni (changes requested):
Do not bind the unauthenticated harness and Vite server to 0.0.0.0: the public API includes config writes, bot control, connector actions, and Box commands. Please keep both listeners loopback-only and remove the unrelated Node engine downgrade; then add focused Hermes ACP auth/contract tests for the driver itself.

milind-soni:
Closing this because it is no longer a safe or current integration path. Alongside the three-line Hermes driver, this branch changes the app and Vite servers from loopback-only to `0.0.0.0` (exposing a local control surface on the network), lowers the supported Node version, and targets provider/config contracts that have since changed. Hermes support should come back as a focused driver against current contracts, with tests and without unrelated network-binding changes.

PR #55, opened 2026-08-13
feat: add Reasonix ACP driver

y0usaf:
Adds [Reasonix](https://github.com/y0usaf/reasonix-flake) — a cache-first DeepSeek coding agent with a documented [ACP v1](https://agentclientprotocol.com/) implementation (`reasonix acp`, NDJSON JSON-RPC 2.0 over stdio).

## Change

One support object on the generic ACP runtime (`server/drivers/acp/core.ts`), plus a registration line and a default-fleet entry:

- `server/drivers/acp/reasonix.ts` (new) — 44 lines
- `server/drivers/builtIn.ts` — +1 import, +1 registration
- `server/config.ts` — +1 default-fleet entry

## Details

- **Models**: `deepseek-flash-0731` (default), `deepseek-pro`
- **Auth**: lenient (`authFailure: "continue"`). The only advertised method is the terminal `reasonix-setup` flow, which a GUI host cannot drive, so turns proceed on Reasonix ambient login (`~/.reasonix/.env`).
- **Verified live** against reasonix 1.24.1: `initialize` advertises `protocolVersion: 1` and the `reasonix-setup` terminal auth method.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Reasonix AI provider through the ACP v1 protocol.
  * Reasonix is now available as a default instance option.
  * Added support for selecting Reasonix models when starting a session.
  * Automatically detects existing Reasonix credentials, including platform-specific credential locations.
  * Supports custom Reasonix home-directory configuration.
  * Sessions can continue when terminal authentication is unavailable.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -108,6 +108,7 @@ export function instanceConfigs(cfg: AppConfig): InstanceConfigMap {
           codex: { driver: "codex" },
           antigravity: { driver: "antigravityAgent" },
           opencodeGo: { driver: "opencodeGo" },
+          reasonix: { driver: "reasonix" },
           computer: { driver: "boxAgent" },
         };
   for (const entry of Object.values(map)) {
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -20,6 +20,7 @@ import { GrokAgentDriver } from "./grok.ts";
 import { GeminiAgentDriver } from "./gemini.ts";
 import { KimiAgentDriver } from "./kimi.ts";
 import { DroidAgentDriver } from "./droid.ts";
+import { ReasonixDriver } from "./reasonix.ts";
 
 const FAKE_CLI = join(dirname(fileURLToPath(import.meta.url)), "..", "..", "testing", "fake-acp-cli.ts");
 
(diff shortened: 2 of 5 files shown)


milind-soni (changes requested):
The generic ACP adapter makes this small, but the branch currently conflicts and the provider-specific credential/auth behavior has no tests. Please rebase and add focused coverage for REASONIX_HOME/Windows fallback plus valid, empty, comment-only, quoted-empty, and export-prefixed .env values before adding Reasonix to the default fleet.

y0usaf:
@coderabbitai review

y0usaf:
Tests added — 16 covering REASONIX_HOME, Windows fallback, and the .env grammar (valid, empty, comment-only, quoted-empty, export-prefixed). Branch is clean against main now, no conflict.

milind-soni:
Reviewed against current main. The driver shape is right and the ACP core does inherit most of the contract — but the credential lookup diverges from its siblings in a way that's actually test-pinned.

`reasonixHome()` falls back to `os.homedir()` — the *server process's* home. The ACP core builds the child env as `{...process.env, ...input.environment, PATH}` and hands that to `isAuthenticated`, so an instance configured with its own `HOME` gets checked against the wrong directory. Both neighbours do it the other way deliberately: `kimi.ts` uses `env.KIMI_CODE_HOME || join(env.HOME || homedir(), ".kimi-code")` and `droid.ts` uses `env.FACTORY_HOME_OVERRIDE || env.HOME || homedir()`, and there's a test named "kimi checks KIMI_CODE_HOME before the child HOME" in `server/drivers/acp/acp.test.ts` that pins exactly this precedence.

Mirroring that one line would make this consistent with the rest of the fleet, and it also needs a rebase (currently conflicting). Worth finishing — a new ACP harness is cheap to add now that the core exists.

PR #56, opened 2026-08-13
feat: add OpenAI-compatible provider manager

carbongotfound:
## Summary

Adds a reusable, chat-only OpenAI-compatible provider path without changing the existing Claude, Codex, Grok, Gemini, or xAI integrations.

- Adds one streaming `/chat/completions` driver with transcript replay, cancellation, a bounded timeout, safe HTTP errors, malformed-SSE tolerance, and provider-reported token usage.
- Adds configuration presets for OpenAI, OpenRouter, DeepSeek, Groq, Together, Fireworks, Mistral, Qwen, Kimi, MiniMax, Ollama, LM Studio, vLLM, and multiple custom OpenAI-compatible endpoints.
- Adds local write-only credential storage, safe renderer-facing provider snapshots, live `/models` discovery, cached-model recovery, and manual model fallback.
- Adds a Provider Manager in App Settings and marks API instances as `Chat only` in the model picker.

## Security and reliability

- API keys are never returned through provider APIs or renderer state.
- Remote endpoints must use HTTPS; plain HTTP is restricted to loopback addresses for local runtimes.
- API instances do not advertise MCP, computer, Composio, or peer-agent capabilities.
- Discovery failures retain cached models and surface safe status guidance.

## Validation

- `pnpm typecheck`
- `pnpm test` — 70 passed, 39 skipped
- `pnpm build`
- Manual Vite + harness UI verification of App Settings, provider presets, keyless local setup, and chat-only guidance.

## Limitations

- This is deliberately chat-only: a permission-aware function/tool bridge is deferred.
- The direct xAI driver remains intact; its migration to the shared driver is intentionally deferred.
- This slice records provider-reported tokens only. It does not claim usage-reduction benchmark results.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring OpenAI-compatible API providers.
  * Manage standard or custom providers, API keys, model lists, labels, and default models.
  * Discover and refresh available models automatically.
  * Added provider status, error reporting, and provider removal options.
  * Provider changes now refresh available instances automatically.
* **UI Improvements**
  * OpenAI-compatible instances are marked with a “Chat only” badge.
* **Bug Fixes**
  * Improved handling of authentication, rate-limit, timeout, interruption, and unavailable-model scenarios.
  * Restricted provider endpoints to secure HTTPS or local connections.


milind-soni (changes requested):
The provider manager is valuable, but several correctness issues need fixing first. Saving one API provider currently reloads the entire fleet and can terminate unrelated in-flight turns; rewound turns can send the transcript twice; stale or hand-edited preset IDs can crash display-name resolution; and bracketed IPv6 loopback URLs are rejected. Please make reload targeted and no-op unchanged saves, deduplicate rewind history, harden preset fallback and loopback normalization, then address the remaining secret-safe error, accessibility, and test-isolation findings.

carbongotfound:
Addressed the provider review set: saves now no-op unchanged payloads and reload only the affected api instance instead of disposing the fleet; bracketed IPv6 loopback endpoints and stale preset IDs are safe; rewound history is not sent twice; and discovery guidance is accurate when no cached model exists. Verified with typecheck and focused provider/driver/API tests.

carbongotfound:
Correction to the prior update: the focused suite caught a malformed array spread before completion. Fixed in this follow-up commit; typecheck and focused provider, driver, and API tests now pass.

PR #57, opened 2026-08-13
feat: compact replay context and track task usage

carbongotfound:
## Summary

Adds a bounded, deterministic context-packing path for transcript-replay providers and makes each completed bot turn auditable.

- Replaces the fixed `slice(-40)` replay history with a 12k-character context budget.
- Preserves the original user goal and the most recent exchange verbatim; compacts older user constraints and assistant outcomes into a structured record without calling another model.
- Adds a `transcriptReplay` adapter capability so accounting is shown only when a provider actually receives replayed history (or a native session has been rewound).
- Persists per-turn model calls, tool calls, computer-action count, provider-reported tokens/cost, duration, and context metrics.
- Adds a compact header indicator with a detailed hover breakdown.

## Benchmark

`pnpm bench:context` uses a stable offline long-running-support fixture:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Messages sent | 32 | 3 | 29 compacted |
| Prompt characters | 155,801 | 7,837 | 95.0% reduction |
| Estimated input tokens | 38,951 | 1,960 | 95.0% reduction |

The benchmark measures deterministic prompt payload only. It verifies structural preservation of the original goal and latest exchange, but does **not** claim provider latency or model task-success improvements.

## Validation

- `pnpm typecheck`
- `pnpm bench:context`
- `pnpm build`
- Manual Vite + harness browser check: page loaded, meaningful content and key controls rendered, no Vite overlay or browser errors.

## Limitations

- Provider-reported token/cost values depend on what each existing driver emits; estimated context tokens are explicitly labelled as estimates.
- Semantic caching, model routing/escalation, and computer/browser observation optimizations remain separate follow-up slices.

## Summary by CodeRabbit

* **New Features**
  * Added an expandable task-usage summary showing duration, model and tool activity, token usage, context metrics, and provider cost.
  * Added automatic transcript compaction for long conversations while preserving goals, recent exchanges, and key decisions.
  * Added run status tracking for active and completed tasks, including success, errors, and stop reasons.
  * Added a deterministic context-packing benchmark command.

* **Bug Fixes**
  * Preserved completed task-usage data across application restarts while clearing incomplete runs.


milind-soni (changes requested):
The usage UI and bounded replay direction are good, but the compactor does not yet guarantee the exact original goal it claims to preserve, especially for a large goal or when no middle entries are omitted. Run accounting also records a model call before the adapter accepts the turn, and restart cleanup can remove activeRun only in memory without persisting the normalization. Please reserve budget for the raw first goal, count calls only after acceptance or turn start, persist cleanup once when needed, and add focused regression tests.

carbongotfound:
Addressed the actionable review feedback in d232be4. Context packing now keeps the original first user instruction verbatim (including oversized goals), records a model call only after sendTurn accepts the request, and persists stale active-run cleanup after restart. Added regressions for oversized goals and durable restart cleanup. Verified with: corepack pnpm exec vitest run server/context.test.ts server/store.test.ts server/index.test.ts (30 passed) and corepack pnpm typecheck. The merge also retains current upstream screen-polling behavior; the existing generated upstream dist-server/index.js whitespace warning was left unchanged.

milind-soni:
Thank you for this, and for following up on the review feedback at the time — keeping the original first instruction verbatim while packing context was the right call. Re-checked today: the branch is 1,156 commits behind `main`, and the two things it set out to do have since landed by other routes — compact replay context through `server/turn-context.ts` and the context work now being split from #759 onto current `main` (#1038 first), and task usage through the per-thread token-usage events every driver now emits. Nothing here would merge as-is any more, so closing it with thanks rather than leaving it open. If there is a piece you still miss on `main`, a small PR against current `main` would be welcome.

PR #58, opened 2026-08-13
feat: add structured computer observation guards

carbongotfound:
## Summary

Adds a small, provider-neutral observation policy to the existing cloud-computer MCP proxy. It prefers structured Chrome state before visual observations and avoids resending unchanged screenshots.

- adds safe structured Chrome state via the local DevTools target list (page title and URL with query/fragment removed)
- adds bounded navigation verification for `open_url` and an explicit `wait_for_navigation` tool
- avoids repeat captures when no computer action has made the screen dirty, and suppresses an image when freshly captured pixels are unchanged
- supports bounded screenshot crops and exposes per-session observation/action/retry/verification metrics
- adds deterministic unit tests and an offline benchmark fixture

## Why

Cloud-computer turns previously treated screenshots as the primary observation mechanism. Many navigation and load checks can be answered using structured browser state, and stable screens should not consume another vision payload. The policy keeps an action-triggered fresh capture so it does not claim a deterministic action succeeded without new evidence.

## Overlap avoided

This does not duplicate open PR #37. PR #37 changes `server/index.ts` so a final screen frame is posted only after a turn uses the computer. This PR leaves that path alone and improves the existing cloud-computer MCP observation/action contract. It also does not change the separate local CUA architecture or permission broker.

## Architecture

`ObservationCoordinator` is independent of the MCP transport and models. The proxy uses it for metrics, change detection, cropping, and bounded verification; future computer providers can reuse the same policy without provider-specific prompting hacks.

Structured state is best-effort and deliberately safe: Chrome URLs are limited to HTTP(S) and query/fragment data is removed before it reaches an agent. A failed structured check retains visual fallback rather than reporting success.

## Validation

- `pnpm typecheck`
- `pnpm bench:observation` — deterministic fixture reduced model-facing screenshot observations from 7 to 3 (57.1%); captured 3 frames and reused 4 stable reads
- `pnpm build`
- MCP protocol smoke: new tools advertised and `observation_metrics` returned a safe zeroed snapshot
- Local Vite UI smoke: normal workspace rendered, non-empty page, no Vite error overlay

## Limitations

- The 57.1% figure is from an offline deterministic fixture, not a real-world token, latency, or task-success claim.
- This slice uses Chrome's target-list endpoint only; DOM/accessibility-tree inspection and structured element actions are follow-up work.
- Live cloud-Box/DevTools integration could not be exercised without a configured cloud computer, so the proxy behavior is covered by unit/protocol tests rather than an external desktop run.

## Summary by CodeRabbit

* **New Features**
  * Improved computer observation with reduced duplicate screenshot transmission and more reliable action-triggered captures.
  * Added bounded screenshot cropping and structured browser state reporting.
  * Added tools for browser state, navigation waiting, and observation metrics.
  * Added support for verified browser navigation with remote debugging.

* **Bug Fixes**
  * Improved browser target parsing, URL handling, frame-change detection, and verification tracking.

* **Tests**
  * Added coverage for observation scheduling, crop validation, browser parsing, and reliability metrics.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "build": "tsc -b && tsc -p tsconfig.server.json && vite build",
     "typecheck": "tsc -b && tsc -p tsconfig.server.json",
     "test": "vitest run",
+    "bench:observation": "node --experimental-strip-types scripts/bench-observation.ts",
     "test:watch": "vitest",
     "preview": "vite preview",
     "build:server": "tsc -p tsconfig.server.build.json",
--- a/scripts/bench-observation.ts
+++ b/scripts/bench-observation.ts
@@ -1,0 +1,18 @@
+import { ObservationCoordinator } from "../server/computer-observation.ts";
+
+// Stable fixture: old strategy sends a full screenshot after every read. The
+// new policy captures after each deterministic action but only sends an image
+// when pixels change; repeated reads without actions reuse the observation.
+const frames = ["initial", "initial", "after-click", "after-click", "after-submit", "after-submit", "after-submit"];
+const oldSent = frames.length;
+const coordinator = new ObservationCoordinator();
+for (const [index, frame] of frames.entries()) {
+  if (index === 0 || index === 2 || index === 4) coordinator.noteAction();
+  if (coordinator.needsCapture()) coordinator.observeFrame(frame, null);
+}
+const metrics = coordinator.metrics;
+const reduction = ((1 - metrics.screenshotsSentToModel / oldSent) * 100).toFixed(1);
+console.log("Computer observation benchmark (offline deterministic fixture)");
+console.log(`Screenshot observations sent to model: ${oldSent} → ${metrics.screenshotsSentToModel} (${reduction}% reduction)`);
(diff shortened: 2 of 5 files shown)


milind-soni (changes requested):
The observation coordinator is a good direction, but the current implementation can report incorrect verification and leak URL credentials. Navigation compares against a redacted URL, an invalid expected URL can succeed, crop bounds ignore image height, crop conversion failures are suppressed, and change detection hashes cropped payloads instead of the canonical frame. Please fix those paths, strip username and password before any URL is exposed, and extend the deterministic tests and benchmark to cover suppression, crop failures, navigation query changes, and every public metric.

milind-soni:
Integrated through #92 after rebasing the contribution onto current main and preserving commit 322e569 as a merge parent. The integration fixes crop-bound validation, canonical full-frame hashing, fail-closed crop conversion, credential-safe navigation checks, loopback-only Chrome DevTools with a private per-user profile, accurate observation metrics, redirect reporting, and deterministic suppression coverage. Final validation passed: 215 tests (7 skipped), production build, Electron syntax checks, the observation benchmark, macOS/Ubuntu/Windows CI, packaging smoke, and all reviewer threads. Thank you for the foundation here.

PR #60, opened 2026-08-13
feat: add durable task checkpoints and resume

carbongotfound:
## Summary

Adds durable, user-addressable task checkpoints on top of OpenMausBot's existing transcript branches and provider-native resume cursors.

- captures a checkpoint automatically when a bot turn starts
- persists active branch, model selection, last message pointer, status, and reason in the existing local `bots.json`
- marks a running checkpoint interrupted after a harness restart instead of pretending the task is still live
- marks checkpoints completed, interrupted, or failed from existing turn lifecycle paths
- adds list and resume endpoints, restoring the saved branch and reusing the existing provider cursor where available
- protects resume against a changed model, missing provider, completed checkpoint, missing branch, or a currently busy bot

## Architecture

This reuses the existing transcript tree and `resumeCursors`; checkpoints are lifecycle pointers, not a second task engine or a copy of conversation data. They are local-first, capped at 20 per bot, and safe across legacy bot records that do not yet contain checkpoints.

## Overlap avoided

PR #51 owns routines and command palette. This PR does not introduce scheduling, templates, or routine UI. It confines itself to bot turn persistence and explicit resume HTTP endpoints.

## Validation

- `pnpm typecheck`
- `pnpm build`
- focused store/harness tests covering persistence, restart recovery, lifecycle updates, checkpoint listing, and invalid resume requests

## Limitations

- Resume currently uses the API; a dedicated checkpoint UI belongs with the later task dashboard/usage UI to avoid duplicating PR #51 and #57 surfaces.
- It cannot revive a process that a provider itself does not support resuming; in that case the persisted transcript is still replayed by the existing driver behavior.

## Summary by CodeRabbit

* **New Features**
  * Added durable checkpoints for bot tasks, preserving conversation progress and model selection.
  * Added checkpoint listing and resume capabilities.
  * Resumed tasks can accept an optional instruction.
  * Checkpoint status updates now reflect completion, interruption, failure, or restart recovery.

* **Bug Fixes**
  * Resuming a nonexistent checkpoint now returns a clear not-found response.


milind-soni (changes requested):
Checkpoint persistence is useful, but resume can currently restore an older transcript branch while reusing the provider cursor from the newer branch, which reintroduces abandoned context. Mark the bot rewound after setActiveLeaf so the cursor is discarded and the restored path is replayed. Also capture the current activeLeafId and lastMessageId when interruption happens, and strengthen tests for the model-selection snapshot and transcript pointers.

carbongotfound:
Addressed the checkpoint review while resolving against current upstream. Interruption now records the current active leaf and last message; resume marks the bot rewound after restoring the checkpoint leaf so the provider cursor is discarded and only the restored path is replayed. The existing checkpoint API/store tests pass (28 passed) and pnpm typecheck passes.

PR #61, opened 2026-08-13
feat: add local-first MCP manager

carbongotfound:
## Summary by CodeRabbit

* **New Features**
  * Added MCP server management in settings, including stdio and HTTP servers.
  * Servers can be enabled, disabled, removed, refreshed, and inspected for health and available tools.
  * Added bot-specific server assignments and configuration status visibility.
  * Configured MCP integrations are now available during supported assistant sessions.
* **Bug Fixes**
  * Added validation for unsafe or invalid remote server URLs.
  * Sensitive server details and process output are protected from exposure.


milind-soni (changes requested):
The MCP manager should not merge with the current security and protocol gaps. Stored or broadcast remote URLs can contain userinfo credentials; stdio inspection buffers unbounded child output; initialize skips the required notifications/initialized message; sanitized names can collide across servers; and updates can change routed IDs. Please fix those issues, accept bracketed IPv6 loopback safely, return generic API errors rather than raw exception messages, and add protocol, collision, output-limit, secret-redaction, and stable-ID tests.

carbongotfound:
Addressed core MCP review findings: bracketed IPv6 loopback HTTP is accepted, stdio inspection output is bounded and sends notifications/initialized before tools/list, and configured MCP IDs now receive collision-free names in both Claude and ACP adapters. Also merged current upstream computer capability changes. Typecheck and focused MCP/ACP tests were run before this update.

carbongotfound:
Completed the remaining MCP renderer/API feedback: unexpected MCP write failures are now mapped to a generic server error, the renderer does not display raw exception text, and update/remove/inspect clear stale errors before a new operation. Full verification passed: pnpm typecheck, pnpm test (73 passed, 40 platform-skipped), and pnpm build.

carbongotfound:
Finished the last two current threads: create IDs are route-safe and update IDs immutable; remote MCP URLs with userinfo credentials are rejected before persistence/snapshots. Added coverage for bracketed IPv6, credential rejection, ID validation, and immutable updates. Verified with typecheck and server/mcp.test.ts.

gianito:
User of the packaged macOS app (0.1.41) here — +1 on shipping HTTP in this PR rather than treating it as a later slice.

I already run several MCP servers against Cursor that OpenMausBot cannot mount today: remote HTTPS (Axonaut, Recraft, Xello, Couac) and loopback HTTP on 127.0.0.1 (Beeper, SuperWhisper, local Couac broker). Composio’s catalog does not include those products, so Connected apps cannot replace them. #397 looks like the right stdio path for Things/Fantastical-style servers, but it explicitly leaves remote HTTP out of scope. This PR is the one that would cover the HTTPS + loopback case.

Happy to test a build against those endpoints if useful.

milind-soni:
Closing this early MCP-manager implementation in favor of the custom MCP manager shipped in #697. We now maintain one registry, UI, validation path, and provider bridge. This PR also proposed per-bot assignment and remote HTTP support, which are not claimed as completed by #697; those are better scoped as follow-ups on the shipped registry rather than merging a parallel manager. Thanks for the initial design and security work.

PR #62, opened 2026-08-13
feat: add selective reusable skills

carbongotfound:
## Summary by CodeRabbit

* **New Features**
  * Added skill management in Settings, including creation, enabling, disabling, and deletion.
  * Added controls to assign enabled skills to individual bots.
  * Skills now enhance bot prompts only when explicitly assigned and enabled.
  * Skill instructions are protected from configuration and API responses.
* **Bug Fixes**
  * Deleting a skill automatically removes it from assigned bots.
  * Added validation and normalization for skill data and assignments.


milind-soni (changes requested):
Selective skills are worth keeping, but IDs and assignments need hardening. Updates can replace a skill ID, creates accept IDs that cannot be addressed by the route, and bot patches retain unknown and duplicate skill IDs. Please keep existing IDs immutable, validate or generate route-safe create IDs, deduplicate and filter assignments against known skills, serialize UI mutations, add persistent form labels, and cover these API invariants.

carbongotfound:
Addressed the server-side correctness findings in the latest commit: existing route IDs now remain immutable during updates; supplied create IDs are restricted to the route-safe character set (or generated); source is type-validated; and bot skill assignments are deduplicated and filtered to known skills. Focused skills/API tests passed (15), and typecheck passes. I did not claim the UI mutation/accessibility comments resolved: the one-line component needs a separate careful refactor to add labels and a shared pending state without obscuring its error behavior.

carbongotfound:
Completed the remaining UI review items: the form now has persistent labels; add/toggle/remove share one in-flight state and disable every mutation control; and refresh failures are reported separately after a successful write. Verified with pnpm typecheck and focused skills/API tests.

milind-soni:
Closing this early skills-library implementation as superseded by the current reviewed skill system (including #656 and #833). Main already has per-bot skill discovery, enable/disable/remove, bounded prompt selection, and review-gated creation/updates backed by SKILL.md rather than another config-embedded library. Thanks for the original selective-loading design. We should extend the current skill system instead of maintaining a parallel registry and settings manager.

PR #63, opened 2026-08-13
feat: hand off video turns to local watch-skill

carbongotfound:
## Summary by CodeRabbit

- **New Features**
  - Added support for analyzing referenced videos from YouTube, Vimeo, Twitch, Loom, and common video-file links.
  - When available, video references can be indexed to provide timestamped evidence and support follow-up questions.
  - Added optional local video-analysis integration support.

- **Tests**
  - Added coverage for detecting supported video links and ignoring unrelated URLs.


carbongotfound:
@coderabbitai review

milind-soni (changes requested):
The local-video handoff is useful, but this needs fixes before merge. The host matcher currently treats any page on YouTube/Vimeo/Twitch/Loom as a video; the synchronous  probe can block each matching turn for two seconds; ACP drops the integration environment; and URL-less follow-up turns lose the mounted tool/index. Please make the probe async, validate provider-specific video URLs and trailing punctuation, forward env values, persist the active video per thread, and expand the detector/integration tests.

carbongotfound:
Addressed the actionable feedback in the latest branch commit. Detection now accepts only provider-specific video URLs (plus direct video files), strips trailing prose punctuation, and includes Vimeo/Twitch/Loom and negative-page coverage. hasWatchSkill is asynchronous and cached instead of synchronously blocking every matching turn. ACP now forwards watch-skill env as {name,value} entries with a regression test. The last valid video URL is persisted per bot/thread, so URL-less follow-ups remount watch-skill and use its durable local index after a harness restart. Verified: corepack pnpm exec vitest run server/watch-skill.test.ts server/drivers/acp/acp.test.ts (7 passed; 10 POSIX-only ACP tests skipped on Windows) and corepack pnpm typecheck.

milind-soni:
Thanks for the video-analysis integration. We now have a user-configured local MCP path (Plugins → MCP servers), so tools such as watch-skill should plug into that existing registry instead of adding a special-case auto-mount in each provider and a bot-wide active-video field. The automatic URL detection here is not already shipped; this is a simplification decision. Closing without merging, with the video workflow still possible through an explicitly configured MCP server.

PR #64, opened 2026-08-13
feat: surface objective turn evidence

carbongotfound:
## Summary by CodeRabbit

* **New Features**
  * Added activity updates summarizing whether turn-related actions succeeded or failed.
  * Success and failure summaries now reflect the outcomes of multiple actions within a turn.
  * Agent-reported completions are distinguished from independently verified results.
  * Suppressed summaries for purely conversational turns without actionable evidence.

* **Bug Fixes**
  * Prevented late notifications after a completed turn from creating duplicate activity events or altering turn state.


carbongotfound:
@coderabbitai review

milind-soni (changes requested):
The evidence distinction is good, but the current fold can lose or recreate evidence incorrectly.  drops successful-action counts whenever the overall turn fails, and late ACP notifications can recreate per-turn state after settlement/cancellation. Please preserve succeeded/failed counts independently of , ignore notifications after the session settles, and add exact-payload tests for failed turns with successful and mixed tool outcomes.

carbongotfound:
Addressed the requested evidence fixes in the latest commit. Successful and mixed tool counts are now preserved even when the overall turn fails; 	urnOk only affects the summary status. Codex now ignores notifications after settlement, with a fake ACP regression case. Focused verification and Codex tests pass (4 passed, 9 platform-gated skips), and pnpm typecheck passes.

carbongotfound:
Addressed the remaining cancellation race: interrupt now marks the turn stopping before kill, and both notifications and server requests are ignored while stopping or settled. The late-event regression now asserts the expected initial command and explicitly rejects late command activity. Verified with typecheck and focused Codex/verification tests (POSIX spawn cases remain skipped on Windows).

milind-soni:
Thanks for being careful to distinguish action success from actual verification. We already show individual tool outcomes in the transcript, and the current Codex stream handler stops consuming notifications once its turn settles. We are not adding a second per-turn accumulator/completion chip just to count successful tools: stronger verification should report observed results, not another success counter. Closing this older implementation as a scope/simplification decision, not claiming independent task verification is complete.

PR #65, opened 2026-08-13
feat: add reusable task templates

carbongotfound:
## Summary by CodeRabbit

* **New Features**
  * Added Task Templates management in Settings.
  * Create, view, launch, and delete reusable task templates.
  * Templates can define names, titles, descriptions, instructions, and computer modes.
  * Bots created from templates inherit the configured details and greeting.
  * Added validation, safe template previews, and clear error handling.

* **Bug Fixes**
  * Improved handling of duplicate client IDs and missing templates.


carbongotfound:
@coderabbitai review

milind-soni (changes requested):
Task templates are worth keeping, but the creation path is not ready yet. Client-supplied IDs can contain characters that make a template impossible to delete and can also collide; renaming the bot only after createBot means persistence and greeting side effects use the generated identity. Please always generate IDs server-side, pass the template name into initial bot creation, add accessible labels to the form controls, and cover the identity and ID behavior in tests.

carbongotfound:
Addressed the template creation review. Template IDs are now always generated by the harness, so client IDs cannot collide or produce undeletable routes. Store.createBot accepts initial template identity before persistence and greeting creation, and the template form controls now have accessible names. Added API coverage for server-owned IDs, independent deletion, and correct initial greeting. Focused tests pass (28 passed) and pnpm typecheck passes.

milind-soni:
Thanks for building and testing this. Reusable bot setups now have a home in the existing Teams/library and bot-package import flow. We want to improve that path rather than introduce a second Task Templates catalog, settings page, and API family, so we are closing this implementation as a scope/simplification decision. Improvements to creating and reusing templates within the existing library are still welcome; this is not a claim that every option here is already implemented.

PR #66, opened 2026-08-13
feat: add an honest execution timeline

carbongotfound:
## Summary by CodeRabbit

* **New Features**
  * Added a collapsible task timeline above chat messages.
  * Displays recent task activity, including tools, screens, results, and user follow-ups.
  * Shows event timestamps and running, completed, or failed statuses.

* **Tests**
  * Added coverage for timeline activity, persisted events, failures, pending states, and task transitions.

### DIFF
--- a/server/task-timeline.test.ts
+++ b/server/task-timeline.test.ts
@@ -18,4 +18,21 @@ describe("execution timeline", () => {
       { kind: "result", state: "complete" },
     ]);
   });
+
+  it("shows a pending activity as running until its completion patch arrives", () => {
+    const start = { id: "tool", role: "bot" as const, kind: "activity" as const, tool: { name: "browser.search" }, at: 2 };
+    expect(timelineEvents([start])).toMatchObject([{ label: "browser.search", state: "running" }]);
+    expect(timelineEvents([{ ...start, tool: { name: "browser.search", ok: true } }])).toMatchObject([
+      { label: "browser.search", state: "complete" },
+    ]);
+  });
+
+  it("does not call every later user reply a new task", () => {
+    const events = timelineEvents([
+      { id: "task", role: "user", kind: "text", text: "Research this", at: 1 },
+      { id: "reply", role: "bot", kind: "text", text: "What should I compare?", at: 2 },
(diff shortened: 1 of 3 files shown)


milind-soni (changes requested):
The timeline is useful, but it currently presents an in-progress tool as completed: activity messages are created when a tool starts with no outcome yet, while timelineEvents maps every non-false outcome to the green complete state. Please represent pending actions as running or observed until the completion patch arrives, test that transition, and avoid describing every user reply, including answers and continuations, as a new task start.

carbongotfound:
Addressed the timeline review. A newly-created activity now renders as running until its completion patch sets 	ool.ok; failures remain failed. Only the first user message is called Task started; later replies are conservatively labelled User input. Added transition and follow-up regression coverage. Focused timeline tests pass (3 passed) and pnpm typecheck passes.

milind-soni:
Reviewed against current main. The branch is **135 commits behind** — it merges with only a one-line lucide-icon conflict, and after resolving, typecheck and the suite are green, so mechanically it's fine.

The problem is that everything it was verified against has moved. The PR body certifies "62 passed, 39 skipped"; main's suite is now 335. Droid ACP, Kimi, opencode, resumable hydration, group conference calls and the Chief of Staff all landed after this branch's merge base, and the timeline this adds renders alongside them without knowing they exist.

A rebase onto current main plus a re-run of your own verification would make this reviewable again — the underlying idea (an execution timeline that doesn't overstate what happened) is worth having, especially now that turns fan out across more surfaces than when you wrote it.

milind-soni:
Rebased onto current main and adapted the timeline to the current chat contract, pinned-banner layout, connector messages, Windows drag-region lint rules, and accessible state labels.

Verified with 1,674 passing tests (12 skipped), broker/updater/viewer/packaged-server smoke, typecheck, touched-file lint, Electron checks, and the production UI build. @coderabbitai full review

milind-soni (approved):
Rebased concerns are resolved: pending tools remain running, later user replies are not mislabeled as new tasks, and current-main compatibility/accessibility checks are covered. Full local and platform suites are green.

PR #67, opened 2026-08-13
fix: keep each thread's composer draft when switching bots

guilimasp:
Follow-up to my #39, and the half of it I should have seen at the time.

#39 keyed the composer by bot id so a half-written message stops following you into the next conversation — the wrong-bot send is real. But remounting also throws the draft away: leave a thread mid-sentence, come back, and the box is empty. I even wrote in that PR that per-bot drafts would work too and picked the smaller change; having lived with it, dropping the text is the part users notice.

So: keep the draft instead of discarding it.

- New `src/lib/drafts.ts` — the unsent text per bot (and per room) in `localStorage`, plus a `useDraft(id)` hook that is a drop-in for `useState("")`.
- `Composer.tsx` — one line and an import. Remounting still isolates threads from each other and still resets the mention picker and dictation exactly as #39 arranged; only the text now survives the trip, including across a restart. Sending or emptying the box clears the entry.

Storage is best-effort: a full quota, a locked-down origin, or a garbled value reads as "no drafts" rather than costing a keystroke.

`pnpm typecheck && pnpm test` pass (101 tests). Also ran it in a packaged build: type to one bot, switch away, come back — the text is waiting; send empties it; quitting and reopening keeps it.

## Summary by CodeRabbit

* **New Features**
  * Draft messages are saved automatically while composing.
  * Drafts are preserved separately for each bot or group conversation.
  * Empty drafts are removed automatically when no text remains.

* **Bug Fixes**
  * Improved resilience when saved draft data is unavailable, malformed, or inaccessible.
  * Draft content now remains available when switching between conversations and returning to a previous thread.

### DIFF
--- a/src/components/Composer.tsx
+++ b/src/components/Composer.tsx
@@ -3,6 +3,7 @@ import { useEffect, useMemo, useRef, useState } from "react";
 import { ArrowUp, Clock, Mic, Square, X } from "lucide-react";
 import { useStore, type Bot, type Group } from "@/state/store";
 import { cn } from "@/lib/cn";
+import { useDraft } from "@/lib/drafts";
 import { MausAvatar } from "./Avatar";
 import { normalizeState } from "@/lib/mascot";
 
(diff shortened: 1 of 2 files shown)


milind-soni (approved):
Reviewed against current main: draft keys, storage failure handling, and composer remount behavior are sound. All platform checks pass; combined-tree build and all 159 tests pass.

PR #68, opened 2026-08-13
feat: keep a long paste as an attachment chip

guilimasp:
Paste a stack trace, a log, or a page of code into the composer today and it buries the input: the textarea grows to its cap, the send button drifts down, and whatever you had written scrolls out of sight. The message is still fine — the composer is not.

Past ~900 characters or 12 lines, a paste becomes a chip above the input instead: the first lines fading out, the size beside them, an `x` to drop it. Shorter pastes are untouched and still land as text, so nothing about ordinary typing changes.

On send the chips fold back into the message as `<pasted-text index="…">` blocks. Tagged rather than fenced on purpose: pasted code and markdown carry fences of their own, and nesting them loses the boundary. Nothing crosses the server or a driver — the provider still receives one plain prompt.

Two new files (`src/lib/composer-attachments.ts`, `src/components/ComposerAttachments.tsx`) and ~25 lines in `Composer.tsx`.

#69 stacks on this branch and reuses the same chips for files dropped onto the window.

`pnpm typecheck && pnpm test` pass (101 tests). Exercised in the dev UI against a real thread: a 40-line paste collapses into the chip instead of the input, the chip removes cleanly, and short pastes still type through.

## Summary by CodeRabbit

* **New Features**
  * Long pasted text is now captured as removable attachment chips above the message composer.
  * Attachments show a text preview, size summary, and “Pasted” label.
  * Remove individual pasted attachments before sending or queuing.
  * Pasted content is included in sent or queued messages and cleared afterward.
  * The send control appears when attachments are present, while the microphone stays hidden whenever the composer contains content.
  * Short pastes continue to use the default behavior.

### DIFF
--- a/src/components/Composer.tsx
+++ b/src/components/Composer.tsx
@@ -1,9 +1,16 @@
 import { track } from "@/lib/analytics";
-import { useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { ArrowUp, Clock, Mic, Square, X } from "lucide-react";
 import { useStore, type Bot, type Group } from "@/state/store";
 import { cn } from "@/lib/cn";
 import { MausAvatar } from "./Avatar";
+import { ComposerAttachments } from "./ComposerAttachments";
+import {
+  composeMessage,
+  isLongPaste,
+  pasteAttachment,
+  type Attachment,
+} from "@/lib/composer-attachments";
 import { normalizeState } from "@/lib/mascot";
 
 /** The active @mention query at the caret: the text between an `@` that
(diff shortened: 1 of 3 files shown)


guilimasp:
Thanks — all three were real, fixed in 4d10418:

- **Attachment-only send.** The send control keyed off `text.trim()`, so a chip with nothing typed showed the mic instead of send; Enter worked, the mouse had no way through. Both controls now read one `hasContent` predicate.
- **Remove button.** `hidden` until `group-hover` took the only way to drop a chip out of the keyboard's reach. It is `opacity-0` with `focus-visible:opacity-100` now — same look on hover, back in the tab order.
- **Size.** `String#length` counts UTF-16 units, so the chip read about a third under on accented text. Measured in UTF-8, and measured once at paste time rather than on every re-render — a chip re-renders on each keystroke, and encoding half a megabyte there would be felt.

#69 is rebuilt on top of the fixed base.

milind-soni (commented):
This feature is useful and the latest fixes look sound, but #67 has now landed in the same Composer state. Please rebase onto current main and integrate attachments with useDraft without dropping persisted text; after that I can run the combined build/tests and merge it.

milind-soni:
Integrated through #89 on current main, preserving the original commits as a merge parent. The integration keeps the newer per-conversation draft, approval, and dictation behavior; persists paste chips independently for each bot/room; avoids reserializing large payloads per keystroke; preserves selection replacement; and adds focused draft/composition tests. Windows, macOS, Linux, and packaged-app checks all passed.

PR #69, opened 2026-08-13
feat: drop files onto the chat to attach them

guilimasp:
**Builds on #68** and carries its commit as the first of the two here — GitHub would not let me base a PR on a branch that does not exist in this repo. Merge #68 first and this diff shrinks to its second commit; review that one alone if it helps.

Handing a bot a file today means copying the path out of Finder by hand. Dropping one on the window did something worse: Electron navigated the window to the file and the app went blank.

Dropping a file anywhere on the window now attaches it. An overlay says so while you drag, the file becomes a chip beside any pasted text, and on send it folds into the message as `<attached-file path="…" />`.

**By path rather than by content**, deliberately. Every driver here is an agent that can open a file itself, so a 200 MB video, a PDF and a CSV all work the same way, in one line of prompt, including for providers with no attachment protocol at all. Only the preload can name a dropped file — Electron 32 removed `File.path` — so the bridge grows `getPathForFile` over `webUtils`.

A drag out of a web page carries no file on disk: small text lands as a pasted chip instead, and anything else says so rather than attaching a path that does not exist.

`pnpm typecheck && pnpm test` pass (101 tests). Behaviour I exercised in the dev UI with synthetic drag events — the overlay, the pathless fallback, the notice — plus a packaged build to confirm the preload bridge ships. A real Finder drag is the one path I could not automate; worth a spot check on your side.

## Summary by CodeRabbit

* **New Features**
  * Added support for attaching files and long pasted text to messages.
  * Attachments appear as removable chips with previews or file details.
  * Added drag-and-drop file handling with notices for unsupported files.
  * Messages now include attached files and pasted content when sent.
  * Added accessible keyboard controls for managing attachments.

### DIFF
--- a/electron/preload.cjs
+++ b/electron/preload.cjs
@@ -1,6 +1,6 @@
 // Renderer bridge. contextIsolation stays on; the renderer only ever sees
 // this narrow surface (window.ogb), never Node or ipcRenderer itself.
-const { contextBridge, ipcRenderer } = require("electron");
+const { contextBridge, ipcRenderer, webUtils } = require("electron");
 
 contextBridge.exposeInMainWorld("ogb", {
   /** Host platform ("darwin" | "win32" | "linux") — for platform-aware UI. */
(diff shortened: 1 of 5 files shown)


milind-soni (changes requested):
Drag-and-drop attachments are useful, but this branch is stacked on #68 and now conflicts with the per-thread draft state merged in #67. Please rebase after #68 is updated, store attachment chips with the same bot or room draft so switching conversations cannot lose or misapply them, and serialize file paths safely instead of interpolating raw quotes or markup into an XML-like attribute. Add tests for paths containing quotes and newlines, queued sends, conversation switching, pathless browser drops, and remove or send behavior.

milind-soni:
Merged via #91—thank you! I rebased the contribution through an integration merge so your original commits remain in the main history.

The integration also persists file chips per conversation, safely encodes unusual paths (including quotes and line breaks), preserves multi-file drop order, handles pathless browser text drops deterministically, and adds regression coverage. Windows, macOS, Ubuntu, and packaging checks all passed.

PR #82, opened 2026-08-14
Settings as a modal with sections, and a local computer you can set up

milind-soni:
### Settings, restructured

App settings were a single long right-hand panel. They're now a **modal with a section list** — General · Connections · Local computer — so new settings have somewhere to go that isn't "make the column longer."

### The new section: a computer your bots can borrow

A Linux desktop in a container on the user's own machine. Free, disposable, and isolated from their desktop and files — the middle ground between the cloud box (needs a paid plan) and "This Mac" (points the agent at the desktop the human is using).

It runs the **same X11 desktop our computer tools already speak** (`xdotool` + `scrot`), so the tool layer doesn't change — only where the commands run. The image is Anthropic's MIT-licensed computer-use demo image, which ships exactly the binaries our shell strings call.

**It's a checklist that watches itself, not a wall of instructions.** A read-only status endpoint (`GET /api/local-computer`) reports:

- which container runtime is installed (docker / podman / apple container)
- whether its **daemon is actually up** — installed ≠ running
- whether the image has been pulled
- whether the container is running, stopped, or missing

The panel polls while you work through the steps in a terminal and ticks each one off. Verified live against a real Docker install that was *installed but stopped* — precisely the case a naive `which docker` would call "ready" and then fail two steps later.

### Deliberate choices

- **We never tell anyone to install Docker.** Its licence requires payment above 250 employees or $10M revenue, and for all government use. The steps suggest Colima or Podman (both free) and we adapt to whatever is already installed.
- **Nothing here installs or runs anything.** The endpoint only inspects; the commands are shown for the user to run.
- Commands are generated from the same constants the status check uses, so instructions can't drift from what we look for.

### Notes

- Built in a worktree off current `main`, which surfaced two drifts from freshly merged community PRs: `ApiKeyRow` now carries its own labels, and a `server/local-computer.ts` already exists (the refactored "This Mac" descriptor reader) — mine is `container-computer.ts` to avoid clobbering it.
- Selecting this as a bot's computer is the **next** PR; this one gets the machine ready and tells you honestly whether it is.

## Summary by CodeRabbit

* **New Features**
  * Added a settings modal with sections for profile preferences, API connections, updates, and local computer setup.
  * Added local computer configuration for detecting and setting up a containerized Linux desktop.
  * Added runtime, daemon, image, and container status details with refresh and screen-view controls.
  * Added setup, stop, removal, and command-copy options for the local computer environment.
* **Improvements**
  * Profile name and email changes now save automatically.
  * Added update checking and installation progress indicators.
  * Settings can be dismissed with Escape or by clicking the backdrop.

### DIFF
--- a/server/container-computer.ts
+++ b/server/container-computer.ts
@@ -1,0 +1,110 @@
+// A computer your bots can use, running on this machine in a container.
+//
+// The cloud box needs a paid plan, and the "This Mac" path points an agent
+// at the desktop the human is using. A container sits between them: free,
+// disposable, and isolated from the user's files — and it runs the same
+// X11 desktop our computer tools already speak (xdotool + scrot), so the
+// tool layer doesn't change, only where the commands run.
+//
+// This module is deliberately read-only. It reports what is installed and
+// never installs anything on someone's machine.
+import { execFile } from "node:child_process";
+import { promisify } from "node:util";
+
+import { augmentedPath } from "./env-path.ts";
+
+const run = promisify(execFile);
+
+/** The desktop image. Anthropic's computer-use demo image is the one whose
+ * toolchain matches our shell exactly — it ships xdotool, scrot and
+ * imagemagick — and it is MIT, with an arm64 build as well as x86. */
+export const IMAGE = "ghcr.io/anthropics/anthropic-quickstarts:computer-use-demo-latest";
+export const CONTAINER = "openmausbot-computer";
+
+/** Runtimes we can drive, best-known first. Docker is listed because most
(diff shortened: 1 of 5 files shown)


milind-soni:
Integrated through #88 on the current main branch, with the original commit preserved as a merge parent. The integration addresses all three review threads (runtime-less commands, loopback-only VNC ports, and modal focus containment), restores updater download/install behavior, corrects Apple container command syntax, adds cross-platform setup guidance, detects already-public container ports, and adds focused runtime/network tests. All Windows, macOS, Linux, and packaged-app checks passed.

PR #87, opened 2026-08-14
Local VM: a bot can drive a Linux desktop on this machine

milind-soni:
> Stacked on #82 (the settings modal + setup panel). Review that one first.

The cloud box needs a paid plan, and "This computer" points the agent at the desktop the human is using. **Local VM** is the third option: a Linux desktop in a container on this machine — free, disposable, and separate from the user's own desktop and files.

### It cost almost no new code

The computer tools were never really coupled to the cloud: every tool runs a shell command and reads a file back. So only two functions learned a second transport — commands go through the runtime's `exec` instead of a REST endpoint, and a frame is read straight out of the container instead of over HTTP. Every tool schema, the coordinate scaling, `computer_batch` and the frame dedup are untouched.

### Verified end to end, on a real container

Pulled the image (4.25 GB unpacked, **native arm64**), started it with the exact command the settings panel prints, and drove the **real `computer-proxy.ts`** over MCP:

| | local container | cloud box (measured earlier) |
|---|---|---|
| screenshot | **72 ms** | ~4,000–5,000 ms |
| click, with the fused frame | 546 ms | ~5,900 ms |
| open_url + frame | 125 ms | ~3,500 ms+ |
| 3-action batch | 833 ms | ~6,000 ms |

Frames come back as 21 KB JPEGs. There is no network hop at all, which is where the difference comes from.

**Keystrokes provably reach applications**, not just the shell: the bot launched `xterm`, typed `echo LOCALVM_PROOF_9271 > /tmp/proof.txt` with synthetic keys, pressed Return, and reading that file back returned `LOCALVM_PROOF_9271`. Also confirmed the image ships the three binaries our shell strings call (`xdotool`, `scrot`, `convert`) and reports a 1024×768 display.

### Details

- **Runs on:** Cloud box · Local VM · This computer · Off
- Both drivers ask one helper what env the proxy needs, so the transport decision lives in one place — with a test asserting box credentials can never leak into a container turn
- Choosing Local VM before it's ready says *which* step is missing (no runtime / not started / image not pulled / container stopped) instead of failing inside the agent's turn
- The box-only Computer engine explains itself rather than throwing
- 8 new tests; **186 pass** overall

### DIFF
--- a/server/computer-proxy.ts
+++ b/server/computer-proxy.ts
@@ -26,9 +26,18 @@
 //     type, Enter) in one round trip with one frame at the end.
 //
 // stdout is the MCP channel — never console.log here.
+import { spawn } from "node:child_process";
 const BOX_API = process.env.OGB_BOX_API ?? "https://ascii.dev/api/box/v1";
 const boxId = process.env.OGB_BOX_ID ?? "";
 const token = process.env.OGB_BOX_TOKEN ?? "";
+// The same desktop can live in a cloud box or in a container on this
+// machine. Only the two functions below know the difference — every tool
+// above them just runs shell and reads a file back.
+const container = process.env.OGB_CONTAINER ?? "";
+const runtime = process.env.OGB_RUNTIME || "docker";
+const local = Boolean(container);
+// the desktop images run X on :1; a cloud box uses :0
+const CONTAINER_DISPLAY = process.env.OGB_DISPLAY || ":1";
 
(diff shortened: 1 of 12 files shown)


milind-soni:
Superseded by #99, which re-applies this on top of the rewritten container-computer.ts instead of forcing a rebase — and keeps that module's loopback port check, which #99 now enforces when attaching a bot.

PR #98, opened 2026-08-14
feat: add Kimi Code driver over ACP stdio

koeseo:
## What

Adds Moonshot's **Kimi Code CLI** (`kimi`) as a new agent provider, riding the generic ACP runtime in `server/drivers/acp/core.ts` — the same path as Grok Build and Gemini. One support object (`server/drivers/acp/kimi.ts`), one registration line, one default-fleet instance.

## Why ACP (not Kimi's `--output-format stream-json`)

Verified empirically against kimi-code 0.29.1: the CLI's own stream-json mode emits bare `{"role":"assistant","content":…}` lines — no tool events, no permission protocol. `kimi acp` however is a full ACP server: `initialize` reports `loadSession: true` (session resume works), `mcpCapabilities` http+sse, and a `session/new` → `session/prompt` roundtrip streams `agent_thought_chunk` + `agent_message_chunk` and settles with `end_turn`. So Kimi gets inline permission cards, thinking-stream rendering, and session continuity for free via the existing core.

## Per-harness quirks encoded

- **Auth:** the only advertised authMethod is `{id: "login", type: "terminal"}` (device-code flow) — it can't be driven over ACP, so `pickAuthMethod` returns `null` and turns ride the ambient login from a prior `kimi login` (`~/.kimi-code/credentials/kimi-code.json`), `authFailure: "continue"`.
- **Model selection:** `-m` is a global commander option and must precede the `acp` subcommand (verified on 0.29.1). Catalog mirrors the CLI's own aliases (`kimi-code/k3` default, `k3-256k`, `kimi-for-coding`, `kimi-for-coding-highspeed`).
- **Billing guard:** `MOONSHOT_API_KEY` / `KIMI_API_KEY` are stripped from the child env so a subscription CLI can't silently flip to pay-as-you-go (mirrors the claude/grok drivers).

## Verified

- `pnpm typecheck` clean, `pnpm test` 259 passed / 8 skipped on current main.
- Live end-to-end through the harness API: bot created with `modelSelection {instanceId: "kimi", model: "kimi-code/k3"}`, turn sent, streamed reply with thought chunks, `turn.completed ok` — persona injection via `buildPromptText` confirmed working.
- `/api/instances` snapshot: `kimi → available, authenticated: true, version 0.29.1`.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Kimi Code assistant.
  * Added Kimi as a built-in assistant option with model selection.
  * Added platform-specific installation and sign-in guidance.
  * Kimi authentication is detected automatically when valid credentials are available.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -99,6 +99,7 @@ export function instanceConfigs(cfg: AppConfig): InstanceConfigMap {
       ? cfg.instances
       : {
           grok: { driver: "grokAgent" },
+          kimi: { driver: "kimiAgent" },
           claude: { driver: "claudeAgent" },
           codex: { driver: "codex" },
           antigravity: { driver: "antigravityAgent" },
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -6,7 +6,7 @@
 //
 // The fake CLI is a shebang script Windows cannot exec directly —
 // resolveCliSpawn turns it into `node <script>`, so these run everywhere.
-import { chmodSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { chmodSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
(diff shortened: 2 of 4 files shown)


milind-soni:
Reviewing now. Looks pretty good

kargnas:
Follow-up implemented in #143: Kimi models, defaults, effort levels, and capabilities are now discovered from the installed provider JSON instead of a hardcoded list, then validated and applied through ACP.

PR #99, opened 2026-08-14
Local VM: let a bot drive the Linux desktop on this machine

milind-soni:
Replaces #87, which conflicted after `container-computer.ts` was rewritten. Rather than force a rebase, I re-applied the transport **on top of** the newer module — its platform-aware runtime detection, Apple `container` support and loopback-only port checking are better than what I had, and one of those checks now does real work here.

### What it adds

The setup panel prepares a Linux desktop; this connects a bot to it.

The tools were never really coupled to the cloud — each one runs a shell command and reads a file back. So only two functions learned a second transport: commands go through the runtime's `exec`, and the frame is read straight out of the container instead of over HTTP. There is no network hop at all, which is why it's faster than the cloud path.

- **Runs on:** Cloud box · **Local VM** · This computer · Off
- Both drivers ask one helper what env the proxy needs, so the transport decision lives in exactly one place — with a test that box credentials can never travel with a container turn
- **A container whose desktop ports are open beyond this machine is refused** as a bot's computer. The panel already detected that state; now it's enforced where it matters
- Picking Local VM before it's ready names the missing step (no runtime / not started / image not downloaded / not started / ports unsafe) instead of failing mid-turn, and never touches the cloud provisioner

### Measured on a real container

| | Local VM | Cloud box |
|---|---|---|
| screenshot | **60–88 ms** | ~4–5 s |
| click with fused frame | 557 ms | ~5.9 s |
| 3-action batch | 829 ms | ~6 s |

Driven through the real `computer-proxy.ts` over MCP, against main's current toolset (including the new `browser_state` / `observation_metrics` tools). Earlier verification also proved synthetic keystrokes reach applications: the bot typed a command into `xterm` and the file it created read back correctly.

### A bug this found

**This desktop image cannot resume.** `docker start` on a stopped container exits immediately — `Xvfb is already running on display :1`, because the X lock survives the stop. Our panel was telling users to run exactly that. It now tells them to recreate the container, with the reason recorded in the code.

263 tests pass; typecheck and production build clean.

## Summary by CodeRabbit

* **New Features**
  * Added Local VM as a computer option for supported bots and providers.
  * Added VM readiness checks, setup diagnostics, and a link to Local computer settings.
  * Local VM sessions now use the configured container runtime and display.

* **Bug Fixes**
  * Improved handling of stopped local computers by recreating them when necessary.
  * Cloud computer behavior and credentials remain isolated from local VM sessions.
  * Added clearer messaging when a VM is unavailable or requires setup.

### DIFF
--- a/server/computer-proxy.ts
+++ b/server/computer-proxy.ts
@@ -26,6 +26,7 @@
 //     type, Enter) in one round trip with one frame at the end.
 //
 // stdout is the MCP channel — never console.log here.
+import { spawn } from "node:child_process";
 import {
   normalizeBrowserUrl,
   normalizeCrop,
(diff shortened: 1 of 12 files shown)


milind-soni:
Superseded by #100, which replaces the custom Local VM transport with the direct Cua Driver/MCP integration and includes the legacy-container recreate safeguard.

PR #110, opened 2026-08-14
feat: Telegram gateway — chat with your bots from your phone

koeseo:
## What

A **Telegram gateway** — chat with your bots from your phone. OpenMausBot keeps the messaging-app shape, but only on the Mac; the thing that makes the original compelling is that the bots live in your pocket. This closes that gap without building a mobile app.

It is an optional standalone process (`pnpm gateway:telegram`) that rides the existing harness HTTP+SSE contract — **zero changes to the app or server**, zero new dependencies (plain `fetch` long polling against api.telegram.org).

## How it works

```
Telegram ⇄ gateway (long polling) ⇄ harness 127.0.0.1:8799 (HTTP + SSE)
```

- `/start` / `/bots` shows the roster as inline buttons; pick a bot, then just type. Replies stream in live — the Telegram message grows by throttled edits (4096-char chunking rolls into fresh bubbles).
- Permission asks become inline keyboards — `✅ Allow / ❌ Deny` wired straight to `/api/bots/:id/respond`. Questions offer the agent's choices as buttons, or free-text via reply. Asks from **every** bot come through (named); replies stream only for the active one, other bots' finished turns arrive as a `📬 <name> finished` nudge.
- `/stop` interrupts the active bot. Timed-out asks keep the harness's fail-closed behavior — the gateway then tells you it was resolved without you.
- **Security:** the first chat that sends `/start` becomes the owner (persisted in `~/.openmausbot/telegram-gateway.json`); every other chat is refused. This thing can approve shell commands, so it is deliberately single-user.

Token via `TELEGRAM_BOT_TOKEN` env or `{"telegram": {"token": "…"}}` in `~/.openmausbot/config.json`. Docs in `docs/telegram-gateway.md`.

## Verified

- `pnpm typecheck` clean; `pnpm test` 267 passed / 8 skipped (8 new gateway tests: chunking, keyboards, ask headers).
- **Live end-to-end** against a real Telegram bot + an 8-bot fleet: owner binding, roster keyboard, bot switching from both sides, streamed replies, a file-writing task driven entirely from Telegram, cross-bot `📬 finished` nudges, and SSE auto-reconnect straight through a harness restart.
- **The approval path is now proven live too.** A `rm -rf` on a throwaway directory is refused by `autoDecision` (destructive), so it reached the phone: gateway logged `sent permission card … → message 44`, the ✅ Allow tap came back through the callback, the harness recorded `answered=allow`, the command ran (`Bash ok=true`) and the directory is gone from the filesystem. Deny and the 15-minute broker timeout were exercised on earlier runs and behave as designed (nothing executed, keyboard cleared, user told it was resolved without them).
- Getting there took several tries for an instructive reason: routine tool permissions never reach a human at all — `auto-approve.ts` settles them server-side, and the agent CLIs classify most read-only commands as safe. Only genuinely destructive commands, secret-adjacent paths and questions surface. Worth knowing when testing this.

## Scope (v1)

Text turns, streamed replies, approvals/questions, interrupt. Not yet: rooms, voice, images/screens, multi-user — happy to follow up if this shape fits.

## Summary by CodeRabbit

- **New Features**
  - Added an optional Telegram gateway for interacting with the harness.
  - Supports streamed replies, bot selection, stopping active turns, and permission or question prompts.
  - Splits long messages automatically and reconnects after temporary service interruptions.
  - The first chat to use `/start` becomes the sole authorized chat.
  - Added a command to launch the Telegram gateway.

- **Documentation**
  - Added setup instructions and documented single-chat limitations, including no rooms, voice, images, or screens.

### DIFF
--- a/docs/telegram-gateway.md
+++ b/docs/telegram-gateway.md
@@ -1,0 +1,34 @@
+# Telegram gateway — chat with your bots from your phone
+
+OpenMausBot keeps the messaging-app shape — the Telegram gateway extends it off the Mac. It is an **optional standalone process** that rides the harness HTTP+SSE contract: nothing in the app or server changes, the gateway is just another client.
+
+```
+Telegram ⇄ gateway (long polling) ⇄ harness 127.0.0.1:8799 (HTTP + SSE)
+```
+
+## What you get
+
+- **Talk to any bot** from Telegram: `/start` (or `/bots`) shows your roster as buttons; pick one, then just type. Replies stream in live — the message grows by edits while the bot writes.
+- **Approve from your phone**: permission asks arrive as inline keyboards — `✅ Allow` / `❌ Deny` — wired straight to `/api/bots/:id/respond`. Questions offer the agent's choices as buttons, or free-text answers as a reply. Approvals from **every** bot come through (non-active bots are named); replies stream only for the bot you're talking to.
+- `/stop` interrupts the active bot's turn.
+
+## Setup
+
+1. Create a bot with [@BotFather](https://t.me/BotFather) and copy the token.
+2. Start the harness (`pnpm dev:server` or the app), then:
+
+```bash
+TELEGRAM_BOT_TOKEN=123456:ABC-… pnpm gateway:telegram
+```
+
+The token can also live in `~/.openmausbot/config.json` as `{"telegram": {"token": "…"}}`.
+
+3. Open your bot in Telegram and send `/start`.
+
+## Security model
+
+The **first chat that sends `/start` becomes the owner** — the binding persists in `~/.openmausbot/telegram-gateway.json`, and every other chat is refused. This gateway can approve shell commands, so it is deliberately single-user. To rebind, delete the state file and `/start` again. Timed-out asks keep the harness's fail-closed behavior (deny with a keep-moving note); the gateway then tells you it was resolved without you.
+
+## Scope (v1)
+
+Text turns, streamed replies, approvals/questions, interrupt. Not yet: rooms, voice, images/screens, multi-user.
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "companion": "node --experimental-strip-types companion/src/index.ts",
     "dev:server": "node --experimental-strip-types server/index.ts",
     "dev:desktop": "node scripts/prepare-cloudflared.mjs --current && electron .",
+    "gateway:telegram": "node server/gateway/telegram.ts",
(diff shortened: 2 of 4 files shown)


koeseo:
Review addressed in `dbda794` — all six findings, plus one I only found by running this for a day.

**From the review**
- **`flushStream`** edited the bubble with the *last* chunk and was never awaited — now settles head-then-rest, in order, and every caller awaits it. (Good catch; a long reply genuinely lost its head.)
- **Overflow** no longer re-joins the remainder into one oversized body.
- **Stale `activeBotId`** is repaired and persisted in `activeBot()`, so a deleted bot can't silently mute the whole reply stream.
- **Hot path:** the roster is cached and re-fetched only on a thread miss — it was one `GET /api/bots` per streamed token.
- **`respond` failures** now answer the callback, tell the user, and leave the keyboard usable for a retry, on both the button and free-text paths.
- **`saveState`** write errors are logged loudly — an unpersisted owner binding is a security hole, not an inconvenience.

**Found in live use:** `fetch` rejects with a bare `fetch failed` and hides the reason in `cause`, so an overnight laptop sleep produced hundreds of unreadable lines at a fixed 3s retry. Errors now name their cause, retries back off exponentially to a 60s ceiling, and recovery is logged — a quiet log now means a healthy gateway. Card dispatch is logged too, so "did it even go out?" is answerable from the log alone.

**The approval path is now proven end-to-end**, which the original PR text honestly flagged as untested: a `rm -rf` on a throwaway directory is refused by `autoDecision`, so it reached the phone — `sent permission card … → message 44`, ✅ tapped, harness recorded `answered=allow`, `Bash ok=true`, directory gone from the filesystem. Deny and the 15-minute broker timeout were exercised on earlier runs and behave as designed.

Worth knowing for anyone testing this: routine permissions never reach a human at all (`auto-approve.ts` settles them, and the agent CLIs treat most read-only commands as safe), so only genuinely destructive commands, secret-adjacent paths and questions surface as cards. That is the right default — it just makes the approval path easy to think broken.

`pnpm typecheck` clean, 271 tests pass.

milind-soni:
This looks great. There is an app thats coming up that pairs with the desktop app via tailscale. Will be a great addition for the whole mobile experience. Thanks for the contribution

yonnic:
Nooo, please do NOT add telegram or 3rd party chat gateway !
This is not hermes or openclaw agent !!

The core principles of grok bot and this app is the simple to use and opinionated approach.
We don't need yet another hermes

PR #112, opened 2026-08-14
fix(claude): detect sign-in via `claude auth status`, not the credentials file

### LINKED ISSUE #108: Claude login is falsely reported as missing when ~/.claude/.credentials.json is absent
## Summary

OpenMausBot reports Claude as “installed but not signed in” even though the Claude CLI is authenticated successfully.

## Environment

- OpenMausBot: 0.1.17 (signed macOS Apple silicon build)
- Claude Code: 2.1.224
- macOS (Apple silicon)

## Steps to reproduce

1. Install Claude Code 2.1.224 and complete its interactive login by running `claude`.
2. Verify that `claude auth status` returns:

```json
{
  "loggedIn": true,
  "authMethod": "oauth_token",
  "apiProvider": "firstParty"
}
```

3. Open OpenMausBot and select Claude.

## Actual behavior

The UI shows “Claude is installed but not signed in yet” and disables the Claude models.

## Cause

The packaged Claude driver checks only whether `~/.claude/.credentials.json` exists:

```js
const authenticated = existsSync(join(homedir(), ".claude", ".credentials.json"));
```

That file is absent for this valid current CLI login, while `claude auth status` reports `loggedIn: true`. The app therefore reports a false unauthenticated state.

## Expected behavior

Determine authentication by invoking and parsing `claude auth status` (for example, its `loggedIn` JSON field), or otherwise support the credential storage used by current Claude Code versions.

### LINKED ISSUE #125: Claude engine shows "not signed in" on macOS when login is Keychain-backed (existsSync check on .credentials.json)
**Environment**
- OpenMausBot 0.1.17, macOS (Darwin 25.5.0)
- Claude Code CLI 2.1.233, signed in via OAuth (`/login`)

**Symptom**

The setup screen permanently shows *"Claude is installed but not signed in yet. Run this once, then come back."* even though the CLI is fully authenticated — `claude -p "test"` succeeds from a terminal. Relaunching the app doesn't help.

Conversely, the reverse failure also occurs: if a **stale** `~/.claude/.credentials.json` exists (e.g. left over from an old install) while the live login is in Keychain, the app reports signed-in but every turn fails instantly with `Failed to authenticate: OAuth session expired and could not be refreshed`, surfaced as a bot reply.

**Root cause**

`server/drivers/claude.js` line 442 (in the packaged app):

```js
const authenticated = existsSync(join(homedir(), ".claude", ".credentials.json"));
```

On macOS, Claude Code stores OAuth credentials in the **Keychain** by default (`Claude Code-credentials` generic password item). `/login` does not create or update `~/.claude/.credentials.json`, so file existence and actual auth state are uncorrelated on Mac: fresh Keychain login → file absent → app says not signed in; stale file present → app says signed in while turns fail.

**Repro**

1. On macOS, sign in to Claude Code normally (`claude` → `/login`). Verify `claude -p "hi"` works.
2. Ensure `~/.claude/.credentials.json` does not exist (default for Keychain-backed logins).
3. Launch OpenMausBot → Claude card reports "installed but not signed in."

**Suggested fix**

Replace the existence check with a real probe — e.g. spawn the CLI headlessly and inspect the result (`claude -p` with a trivial prompt and short timeout, or parse the `is_api_error_message`/`authentication_failed` result the SDK already emits on the first turn). That reflects true auth state on every platform and also catches the stale-file case, which the existence check misclassifies in the opposite direction.

**Workaround for affected users**

`printf '{}' > ~/.claude/.credentials.json` — the CLI ignores the empty file and keeps using the Keychain, and the setup gate passes. (If you have a *stale* real file there instead, rename it away and re-run `/login`.)

ciscosurplus:
Fixes #108
Fixes #125

### Root cause

OpenMausBot inferred Claude authentication from credential storage. That is unreliable in both directions:

- macOS OAuth normally lives in Keychain, so a valid login can have no `~/.claude/.credentials.json`;
- a stale credentials file or Keychain item can exist after the live session is no longer usable.

Claude also exits `auth status` with code 1 when signed out while still returning valid `{"loggedIn":false}` JSON, so treating every non-zero exit as “command unsupported” recreates the stale-file false positive.

### What changed

- Ask the configured CLI through `claude auth status --json`.
- Treat a boolean `loggedIn` response as authoritative even when the process exits 1.
- Fail closed on empty, malformed, or unsupported-command output instead of falling back to credential presence.
- Use the same sanitized environment for auth probes and real turns, so an inherited API key cannot make setup claim a login that turns deliberately remove.
- Keep the original contributor’s cross-platform fake-CLI coverage and add regressions for signed-in, signed-out-with-exit-1, malformed/unsupported output, and environment hygiene.

### Validation

- `pnpm typecheck`
- `pnpm test` — 330 passed, 8 skipped; updater suite 11 passed
- `pnpm build`
- `pnpm check:electron`

Verified against the machine-readable auth command available in Claude Code 2.1.232/2.1.233.


TyBoyes:
Nice fix. One edge case worth handling: `claude auth status` exits 1 when logged out, not only when the subcommand is missing, so the fallback can misfire.

```
$ HOME=/tmp/empty claude auth status
{ "loggedIn": false, "authMethod": "none", "apiProvider": "firstParty" }
exit=1
```

A logged out user with a leftover `~/.claude/.credentials.json` is then reported as signed in, and the `loggedIn: false` payload is discarded before it is read.

Parsing stdout first and falling back only when there is no parseable JSON would separate the two cases, since a CLI predating the subcommand leaves stdout empty and writes to stderr. The fake CLI would then need to exit 1 in the `out` case to match the real contract.

milind-soni:
Reviewed this against current main — the bug you're fixing is **real and still live**: `server/drivers/claude.ts` still decides sign-in with `existsSync(~/.claude/.credentials.json)`, which is empty on a Mac where Claude Code stores credentials in the Keychain. So a signed-in user still gets a greyed-out engine.

Two things block merging it as written:

1. **The probe inherits the wrong environment.** `probeAuth` spawns `claude auth status` with the full `process.env`, but the turn path deliberately strips `ANTHROPIC_API_KEY` ("subscription users get billed pay-as-you-go if this leaks through"). `claude auth status` reports `{"loggedIn":true,"authMethod":"api_key"}` for any value of that variable **without validating it** — so a user with a stale key exported gets an enabled model picker and turns that then fail. The probe env has to mirror the turn env.
2. **Cost and failure mode.** This turns a ~45ms `--version` check into an extra full CLI spawn (~0.5s measured) on a path the UI re-probes, and every failure (non-zero exit, timeout, non-JSON) collapses back to the same `existsSync` check that has the original bug.

A cheaper fix for the macOS case specifically: probe the Keychain without reading the secret — `security find-generic-password -s "Claude Code-credentials"` (no `-w`, so it never prompts) — and fall back to the credentials file elsewhere. Happy to look again if you'd like to take it that way, or if you'd rather keep `auth status`, passing the same stripped env would resolve the correctness half.

milind-soni (approved):
Reviewed on current main. The refreshed branch treats the CLI JSON payload as authoritative even when `auth status` exits 1, removes the stale credential-storage fallback, uses the same sanitized environment as real turns, and adds deterministic regressions for signed-in, signed-out, malformed/unsupported output, and inherited API keys. Full local validation and the hosted macOS, Ubuntu, Windows, and Linux package-smoke matrix are green.

PR #115, opened 2026-08-14
feat(ui): slide the sidebar into a drawer below md

### LINKED ISSUE #114: Sidebar is unusable below ~400px: the fixed 320px column leaves 40px for the chat
## What happens

`src/components/Sidebar.tsx` declares the sidebar as `w-[320px] shrink-0`. On a
360px-wide viewport — a phone in portrait — that leaves **40px** for the
conversation. The app is not usable on a phone.

| Viewport | Left for the chat |
|---|---|
| Phone portrait (360px) | 40px |
| Phone landscape (780px) | 460px |
| Desktop | all of it |

## Why it is not just "the sidebar is wide"

`src/` contains **no media queries at all** — the four in the repo live in
`src/mascot-preview.css`, the mascot demo page. There is exactly one Tailwind
breakpoint in the whole app (`sm:grid-cols-` in `RoutinesPage.tsx`).

The rest of the CSS is in good shape: 24 widths bounded with `max-w-[…]` and 59
`w-full`. Content already adapts. The blocker is that one declaration.

## What I would like to do

Below `md` (768px), take the sidebar out of the flow and slide it in over the
chat as a drawer, behind a scrim, opened from a button in the top-left. Above
768px, nothing changes.

One implementation note that matters, because it is a real trap: the mobile
classes have to be scoped with `max-md:`, **not** applied at the base and
cancelled with `md:`. Tailwind v4 emits the native `translate` property, and any
value other than `none` makes the element a containing block for its `fixed`
descendants — so `md:translate-x-0` does not cancel anything, it reparents
`NewRoomPanel`'s overlay and the `+` menu backdrop on desktop. I hit this while
building it.

## Scope

I would like to split this in two, to keep each PR reviewable:

1. **Layout** — the drawer itself, its four dismissal paths, and reduced-motion
   support. Touches 6 files, ~100 lines.
2. **Overlay surfaces** — the settings modal (860px), approval cards (840px) and
   the Computer panel, which still overflow below the breakpoint.

I have the first one working, tested, and with before/after screenshots plus a
video of the animation. Happy to open it as a single PR instead if you prefer —
just say which you would rather review.

Nothing in the 17 existing issues covers mobile or responsive, so I wanted to
check the approach with you before sending code, as CONTRIBUTING asks.

NuCl34R:
Closes #114

## The problem

`Sidebar.tsx` declares the sidebar `w-[320px] shrink-0`. On a 360px viewport
that leaves **40px** for the conversation.

`src/` has no media queries — the four in the repo are in
`mascot-preview.css` — and one Tailwind breakpoint in total. The rest of the CSS
is fine (24 `max-w-[…]`, 59 `w-full`); content already adapts. This is the one
declaration in the way.

## What this does

Below `md`, the sidebar leaves the flow and slides in over the chat behind a
scrim, opened from a button in the top-left. Four ways out: tap the scrim, press
Escape, press the button, or pick a conversation.

Above `md`, nothing changes.

## Why `max-md:` and not `md:` cancellation

This is the part worth reviewing carefully.

My first version applied the mobile classes at the base and cancelled them with
`md:relative md:translate-x-0`. That is broken, and not obviously so.

Tailwind v4 emits the native `translate` property, and `translate-x-0` is
`translate: 0px 0px` — a real value, not `none`. Per spec, any `translate` other
than `none` makes the element a **containing block for its `fixed`
descendants**. So `md:translate-x-0` does not cancel anything; it reparents
them. Four `fixed` surfaces are JSX descendants of this `aside` (there are no
React portals in the codebase), and two of them broke **on desktop**:
`NewRoomPanel`'s `inset-0` overlay collapsed to the 320px column, and the `+`
menu backdrop stopped catching clicks over the chat.

Scoping with `max-md:` means none of these properties are emitted above the
breakpoint at all. The desktop guarantee is an absence, not a cancellation — and
a reviewer can verify it by reading the classes rather than reasoning about
state.

Neither `tsc` nor the test suite can catch that class of bug. It only exists at
render time.

## Verification

**Desktop is pixel-identical.** Rendered 1280×800 from a worktree on `main` and
from this branch, against the same harness server, and compared with
`ImageChops.difference` — `getbbox()` returns `None`.

`pnpm typecheck` and `pnpm test` pass (287 passed, 8 skipped).

No automated test is added: `vite.config.ts` declares `environment: "node"`, the
only test dependency is `vitest`, and the four tests under `src/` cover pure
logic without mounting components. Standing up a DOM harness would be a second
concern; per CONTRIBUTING, UI changes are verified with screenshots instead.

Manually checked: all four dismissal paths, focus returning to the button on
Escape, the sidebar's context menu still layering above the drawer, and the
animation stopping under `prefers-reduced-motion`.

## Screenshots

Attached in the comment below: before/after at 360px (drawer closed and open),
the desktop pair at 1280px, and a video of the drawer opening, dismissing on a
scrim tap, and reopening.

## Known limits, deliberately not addressed here

- **Below `md`, the aside is a containing block for its `fixed` descendants** —
  `translate` is what animates the drawer, so this is inherent to the approach.
  Opening the New Room dialog or the `+` menu *from the drawer* resolves their
  `inset-0` against the 320px column rather than the viewport. The dialogs still
  render and still dismiss (Escape, their own controls, tapping inside the
  column); what is lost is tap-outside-to-dismiss over the rest of the screen.
  Fixing it needs a portal, which this codebase does not have and which belongs
  with the overlay work.
- **The closed drawer stays in the tab order and the accessibility tree** below
  `md`. `inert`/`aria-hidden` keyed on the open state would break desktop, where
  the state is always closed but the sidebar must stay interactive; the CSS-only
  route (`visibility` in the transition) flips at 50% of the duration by spec,
  which would blank the drawer mid-slide.
- **Re-picking the conversation you are already on** from another view closes
  the drawer via `activeView`, but re-picking it from the chat view does not
  change any watched value. The scrim remains the way out.
- **Overlay surfaces** — settings modal, approval cards, Computer panel — still
  overflow below the breakpoint. That is the follow-up PR.
- Unrelated observation while touching that block: `animate-panel-in` is not
  covered by `prefers-reduced-motion`. Happy to open a separate issue.

## Notes

Commits are split so the reduced-motion change can be reviewed on its own. Tell
me if you would rather have it squashed.

## Summary by CodeRabbit

* **New Features**
  * Added a mobile menu button that opens the sidebar as a sliding drawer.
  * Added backdrop and Escape-key controls for closing the mobile sidebar.
  * Sidebar navigation now closes automatically after navigation or overlay changes.
  * Restored focus to the menu button after the sidebar closes.

* **Accessibility**
  * Disabled sidebar transitions when reduced-motion preferences are enabled.
  * Improved mobile header spacing for clearer navigation controls.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Loader2, Menu } from "lucide-react";
 import { StoreProvider, useStore } from "@/state/store";
 import { Onboarding } from "@/components/Onboarding";
@@ -17,9 +17,13 @@ import { NoEngines } from "@/components/NoEngines";
 
 function Shell() {
   const { state, dispatch } = useStore();
-  // Mobile-only drawer state. Above md the sidebar is always in flow and this
-  // value has no effect — the md: variants cancel every mobile class.
+  // Mobile-only drawer state. Above md, none of these properties are emitted
+  // at all — Sidebar scopes every mobile class with max-md: rather than
+  // cancelling them with md:, which would still emit a translate value and
+  // turn the aside into a containing block for its fixed descendants (see
+  // Sidebar.tsx's className comment).
   const [drawerOpen, setDrawerOpen] = useState(false);
+  const menuButtonRef = useRef<HTMLButtonElement>(null);
   const group = state.groups.find((g) => g.id === state.selectedId);
   const bot = group ? undefined : (state.bots.find((b) => b.id === state.selectedId) ?? state.bots[0]);
 
@@ -62,6 +66,16 @@ function Shell() {
     return () => window.removeEventListener("keydown", onKey);
   }, [state.bots, state.selectedId, dispatch]);
 
+  // Picking a conversation closes the drawer: on a phone the chat is what you
+  // asked for, and leaving the list up would hide it. Watching activeView too
+  // catches re-selecting the bot that is already current from another view —
+  // the reducer switches the view without changing selectedId. pluginsOpen
+  // and settingsOpen cover the same idea from a different trigger: close the
+  // drawer whenever an action opens something over the chat.
+  useEffect(() => {
+    setDrawerOpen(false);
+  }, [state.selectedId, state.activeView, state.pluginsOpen, state.settingsOpen]);
+
   return (
     <div className="flex h-full flex-col">
       {/* fixed-position popup, bottom-left — outside the layout flow */}
@@ -69,6 +83,7 @@ function Shell() {
       <div className="relative flex min-h-0 flex-1">
       <button
         type="button"
+        ref={menuButtonRef}
         aria-label="Open bot list"
         aria-expanded={drawerOpen}
         onClick={() => setDrawerOpen(true)}
@@ -83,7 +98,13 @@ function Shell() {
           className="absolute inset-0 z-30 bg-black/50 md:hidden"
         />
       )}
-      <Sidebar open={drawerOpen} onClose={() => setDrawerOpen(false)} />
+      <Sidebar
+        open={drawerOpen}
+        onClose={() => {
+          setDrawerOpen(false);
+          menuButtonRef.current?.focus();
+        }}
+      />
       {state.activeView === "routines" ? (
         <RoutinesPage />
       ) : noEngines ? (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { Loader2 } from "lucide-react";
+import { Loader2, Menu } from "lucide-react";
 import { StoreProvider, useStore } from "@/state/store";
 import { Onboarding } from "@/components/Onboarding";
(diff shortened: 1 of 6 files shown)


NuCl34R:
### Mobile, 360px

Before — the 320px column leaves 40px for the conversation:

before-360-sidebar-fills-screen

After, drawer closed — the chat gets the full width:

after-360-drawer-closed

After, drawer open — sliding over the chat behind the scrim:

after-306-drawer-open

### Desktop, 1280px — unchanged

desktop-1280-unchanged

Rendered from a worktree on `main` and from this branch against the same harness
server, then compared with `ImageChops.difference`: `getbbox()` returns `None`.
Identical down to the pixel.

### The drawer in motion


https://github.com/user-attachments/assets/5107d2a8-f223-48a3-9903-85ffd2936c52


Opening, dismissing on a scrim tap, then reopening.

### Context menu layering

The sidebar's own context menu still renders above the drawer:

after-360-context-menu-above-drawer

milind-soni:
Looks great. Merging it in

PR #117, opened 2026-08-15
feat: add Factory Droid driver over ACP stdio

TyBoyes:
## What changed

Adds Factory Droid as an engine: `server/drivers/acp/droid.ts` (`droidAgent`, over
`droid exec -o acp`), the registration in `builtIn.ts`, and a default-instance entry
in `config.ts`, matching how Kimi landed in #98.

Two optional hooks on `AcpSupport` in `acp/core.ts`, both no-ops for the existing
harnesses:

- `configureSession()` runs between `session/new` (or `session/load`) and the first
  `session/prompt`.
- `resolveModels()` supplies a per-instance model catalog, falling back to the
  driver's static list if it is absent or throws.

`windowsKnownDirs()` in `env-path.ts` gains `~/bin`, which is where the Windows
installer puts `droid.exe`.

## Why

Droid is the one ACP harness here that does not take session settings from argv.
`droid exec --help` says so under "Stream JSON-RPC Mode": *"CLI flags do not
configure JSON-RPC sessions: -m/--model, --auto, -r/--reasoning-effort ... are
still validated, but session settings come from JSON-RPC requests."* Confirmed
against 0.196.0: `droid exec -m claude-haiku-4-5-20251001 --auto high -o acp` still
reported `currentModelId` and `currentModeId` from the CLI's own config. So a model
that only reaches `spawnArgs` is accepted and then ignored, and the turn silently
runs whatever `~/.factory/settings.json` selected. `configureSession()` is the only
place the pick can land; a rejected setting fails the turn rather than quietly
running something else, with a message naming the engine, the setting and the
method. Autonomy is set the same way, and both are always explicit so a mode or
model pinned in settings.json can neither make a session yolo nor make `fullAuto`
ask.

`resolveModels()` exists because droid's catalog is half user-local: `custom:`
providers (Azure, a local LM Studio server, ...), `modelFavorites` ordering, and
`sessionDefaultSettings.model` all live in `settings.json`, and none of it can be
enumerated statically. Without it the picker shows a list that does not match what
`droid` itself offers.

Sign-in detection checks all three credential filenames droid can write
(`auth.v2.file`, `auth.v2.loginkeychain`, `auth.v2.keyring`): which one exists
depends on the account's `secure_auth_storage` flag, so checking only the first
reports a freshly signed-in macOS user as signed out. `FACTORY_HOME_OVERRIDE`
replaces the CLI's HOME, not its data root, so the `.factory` segment is kept on
both branches (verified: `FACTORY_HOME_OVERRIDE=/tmp/x` produces
`/tmp/x/.factory/`).

## How it was verified

`pnpm typecheck`, `pnpm test` (35 files, 295 passed / 8 skipped) and
`pnpm check:electron` green on macOS 15 (Apple silicon), against droid 0.196.0.

Against the real CLI:

- `initialize`: protocolVersion 1, `loadSession: true`, promptCapabilities
  image + embeddedContext, authMethods `device-pairing` + `factory-api-key`
- `session/new` advertises modes `normal | spec | auto-low | auto-medium |
  auto-high`; `normal` is "Auto-approves only read operations", `auto-high` is
  "Auto-approves all actions"
- `session/set_mode` and `session/set_model` both accepted; unknown values return
  `-32602` ("Invalid autonomy mode", "Model not recognized"), which is the failure
  the driver wraps
- A full turn in `normal` mode against a local provider: droid raised
  `session/request_permission` for a file write with `allow_once` /
  `allow_always` / `reject_once`, rejecting it left the file uncreated, and the
  turn settled `end_turn`
- Install one-liners fetched: the POSIX script is `#!/usr/bin/env sh` and installs
  to `~/.local/bin` (already covered by `knownDirs()`); the Windows PowerShell
  installer writes `droid.exe` to `%USERPROFILE%\bin`, hence the `env-path.ts` line
- `~/.factory/settings.json` is byte-identical after every `set_mode`/`set_model`,
  so these are session-scoped and do not rewrite the user's CLI config

New tests cover argv hygiene (no `-m` in argv, exactly `["exec","-o","acp"]`), the
mode/model calls landing over the wire in both `fullAuto` states, the wrapped error
when an agent predates the methods (a new `no-session-config` mode in the scripted
fake CLI), credential resolution across `FACTORY_HOME_OVERRIDE` / child `HOME` /
`FACTORY_API_KEY` / the keychain filename, and catalog resolution plus its fallback.

## Screenshots (UI changes)

None; no UI changes in this PR. Droid does not get an Onboarding row, matching Kimi
today. Happy to add rows for both if you want that step to cover every default-fleet
engine, or to make it render from `instances` instead, but that felt like a separate
concern from adding the driver.

## Checklist

- [x] Server behavior changes come with tests (see CONTRIBUTING.md, Tests)

## Notes

`FACTORY_API_KEY` is left in the child environment rather than stripped the way
`grok.ts` and `kimi.ts` strip their vendor keys, because droid advertises it as an
auth method and an API-key-only user has no other route. It is checked last in
`isAuthenticated`, so an ambient key can never be the thing that makes a
logged-out instance look ready. Say the word if you would rather it be stripped.

## Summary by CodeRabbit

- **New Features**
  - Added support for Factory Droid as a built-in agent.
  - Droid supports authentication through API keys or local credentials.
  - Added configurable model selection and autonomy settings.
  - Added cross-platform installation detection, including Windows executable paths.
  - Droid now applies model and autonomy settings during session setup.

- **Bug Fixes**
  - Improved fallback handling when model settings are unavailable or unreadable.
  - Added clearer errors for unsupported or incompatible session configuration.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -100,6 +100,7 @@ export function instanceConfigs(cfg: AppConfig): InstanceConfigMap {
       : {
           grok: { driver: "grokAgent" },
           kimi: { driver: "kimiAgent" },
+          droid: { driver: "droidAgent" },
           claude: { driver: "claudeAgent" },
           codex: { driver: "codex" },
           antigravity: { driver: "antigravityAgent" },
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -18,6 +18,7 @@ import { recordEvents, type EventRecorder } from "../../testing/events.ts";
 import { GrokAgentDriver } from "./grok.ts";
 import { GeminiAgentDriver } from "./gemini.ts";
 import { KimiAgentDriver } from "./kimi.ts";
+import { DroidAgentDriver } from "./droid.ts";
 
 const FAKE_CLI = join(dirname(fileURLToPath(import.meta.url)), "..", "..", "testing", "fake-acp-cli.ts");
 
@@ -36,6 +37,15 @@ describe("ACP decodeConfig", () => {
       win32: expect.stringContaining("install.ps1"),
     });
     expect(KimiAgentDriver.install?.signInCommand).toBe("kimi login");
+  });
+  it("droid defaults to the droid binary and declares cross-platform setup", () => {
+    expect(DroidAgentDriver.decodeConfig(undefined)).toEqual({ cli: "droid", fullAuto: false, workspace: undefined });
+    expect(DroidAgentDriver.install?.command).toMatchObject({
+      darwin: expect.stringContaining("factory.ai/cli"),
+      linux: expect.stringContaining("factory.ai/cli"),
+      win32: expect.stringContaining("factory.ai/cli"),
+    });
+    expect(DroidAgentDriver.install?.signInCommand).toBe("droid");
   });
   it("fullAuto only when explicitly true", () => {
     expect(GrokAgentDriver.decodeConfig({ fullAuto: "yes" }).fullAuto).toBe(false);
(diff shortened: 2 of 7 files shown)


kargnas:
Follow-up implemented in #143: Droid now discovers models and reasoning-effort metadata from the installed CLI plus Factory settings instead of a static catalog. The same PR validates and forwards the selected model and effort end to end.

PR #119, opened 2026-08-15
feat: add OpenCode Go engine

StefNoob:
## Summary

- Add OpenCode Go as an optional first-class engine through the maintained OpenCode CLI and ACP stdio runtime.
- Add dynamic `opencode-go/*` model discovery with cached/static fallback and exact ACP model selection.
- Add write-only credential handling, settings/onboarding integration, protocol tests, and documentation.

This is the implementation follow-up to draft #102.

## Security

`OPENCODE_API_KEY` is injected only into the OpenCode child process. It is never returned by the config API or sent to the renderer, logs, analytics, snapshots, error messages, or command arguments.

## Validation

- `pnpm test --run` — 297 passed, 6 skipped
- `pnpm typecheck`
- `pnpm build`
- `pnpm check:electron`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added optional OpenCode Go engine support with dynamic model discovery and fallback behavior.
  * Added secure API key setup through onboarding, Connections settings, and environment configuration.
  * Added OpenCode Go status reporting, provider branding, and automatic engine registration.

* **Documentation**
  * Added setup guidance and detailed integration design and implementation plans.

* **Tests**
  * Added coverage for credentials, model resolution, configuration, catalog handling, and ACP interactions.


StefNoob:
@coderabbitai review

StefNoob:
Addressed the valid CodeRabbit findings in commit f5833d1:

- restricted fake ACP environment dumps to an explicit test allowlist;
- validate `opencodeGo.apiKey` before config persistence;
- added stable provider error codes/classification hooks in the ACP contract/core;
- preserved dynamic model refresh behavior and ACP model confirmation checks;
- updated CLI/auth documentation and OpenCode Go setup links.

The suggestion to delete `OPENCODE_API_KEY` in `stripForeignProviderKeys` was not applied: that transform is also the final child-environment preparation for the OpenCode Go process, so deleting the key would prevent the configured OpenCode Go credential from reaching its authenticated CLI. Existing tests verify that foreign provider keys are removed while `OPENCODE_API_KEY` is preserved.

Validation: 301 tests passed, 6 skipped; typecheck, build, Electron syntax checks, and `git diff --check` passed.

milind-soni (approved):
Reviewed after refreshing onto current main. Verified OpenCode ACP model confirmation, catalog refresh/fallback, saved-login detection, credential isolation, config validation, and cross-platform CI/package smoke.

kargnas:
Follow-up implemented in #143: OpenCode Go now participates in the shared dynamic catalog contract, with model/default metadata discovered from the provider and selections validated before execution. The model picker also uses cached background and manual refresh.

PR #122, opened 2026-08-16
feat(acp): opt-in model selection through a session config option

NuCl34R:
## What

Three capabilities on the generic ACP runtime in `server/drivers/acp/core.ts`,
all of them opt-in, plus one bug fix. No new engine, nothing visible in the UI,
and no existing driver file is touched.

This is step 1 of the plan in #102. The measurements behind each item are in
[my comment there](https://github.com/milind-soni/OpenMausBot/pull/102#issuecomment-5304800457),
taken against `opencode` 1.18.18.

## Why

`opencode acp` cannot be driven by the current core: it takes no `-m`, and it
reports token usage somewhere the core does not read. Rather than fold those
quirks into a driver, they belong in the runtime the same way `pickAuthMethod`
and `authFailure` already do. Each capability below is inert unless a support
object asks for it, so this lands safely on its own, ahead of any engine.

## The three capabilities

**`selectModel?: { configId: string }`.** For harnesses whose ACP subcommand
takes no `-m`, the model is set through `session/set_config_option` after the
session exists. The hook fires on `session/load` as well as `session/new`, so a
resumed thread selects its model too.

It verifies rather than trusts. The agent has to *confirm* the new value in the
returned `configOptions`, and the turn aborts if it does not:

```ts
if (selectedModel !== turn.model) {
  throw new Error(`… did not switch to ${turn.model} (still ${selectedModel ?? "unknown"})`);
}
```

An agent that answers OK and keeps its old model is worse than one that errors,
because it silently burns a paid turn on something other than what the picker
shows.

**`transformEnv(env, config)`** now receives the instance config, so a support
can vary the child env with `fullAuto` instead of only stripping fixed keys.

**`isAuthenticated(env, config)`** may now return a promise. Some harnesses
cannot answer "can this actually run a turn" without asking the CLI.

## The bug fix

`session/prompt` usage was read from `result._meta` only. opencode reports it at
the result root, so the count was dropped:

```ts
const usage = result?.usage ?? result?._meta ?? {};
```

The order matters and is not arbitrary. opencode sends `_meta: {}` alongside the
real numbers, and an empty object is truthy, so reading `_meta` first would
return zero counts for the engine this is meant to fix. Grok and Gemini put
usage under `_meta` and have no root `usage`, so both shapes resolve correctly.

## What does not change

The two signature changes are widenings: an existing `transformEnv(env)` still
type-checks when called with a second argument, and a synchronous
`isAuthenticated` still satisfies `boolean | Promise<boolean>`. That is why no
driver file appears in this diff. Any engine that does not declare `selectModel`
runs exactly the code path it ran before.

## Verified

On this branch alone, against `upstream/main` at `13a1bb7`, under node 24:

| | main @ 13a1bb7 | this branch |
|---|---|---|
| `pnpm test` | 287 passed / 8 skipped, 35 files | **294 passed / 8 skipped, 35 files** |
| `pnpm typecheck` | 0 | 0 |
| `pnpm check:electron` | 0 | 0 |
| `vite build` | 0 | 0 |

The seven new tests, one per behaviour:

```
reads token usage from the root of the prompt result
selectModel confirms the requested model before prompting
a model the session does not advertise fails the turn instead of running another
a model switch acknowledged but not applied fails the turn
selects the model on a resumed session too, not just a new one
transformEnv sees the instance config
awaits an async isAuthenticated
```

The fourth one arrived from CodeRabbit's review of this PR and was worth having.
The unadvertised-model test rides the fake CLI's `-32602`, so it settles inside
`request()` and never reaches the confirmation guard; the guard's own case had
no coverage. It was checked both ways: with the guard neutered the new test does
not merely fail, it reports `ok: true` on a turn that ran `m-one` while `m-two`
was requested, which is precisely the silent wrong-model turn the guard exists
to stop.

The `_meta` shape keeps its existing coverage untouched, which is what proves
the `??` order did not regress the engines already using it. The fake ACP CLI in
`server/testing/fake-acp-cli.ts` grew the switches needed to produce the new
shapes.

One of those tests is a fix to a test I wrote earlier in this branch: it passed
`resumeCursor: "fake-acp-session"`, which is the same id `session/new` returns,
so a `session/load` that threw and fell back would have emitted a
byte-identical `sessionId` and the assertion would still have passed. Proved
rather than argued: making the fake's `session/load` return an error left the
old test green, and turns it red with a distinct cursor.

## Summary by CodeRabbit

* **New Features**
  * Added configurable model selection when starting or resuming sessions.
  * Sessions now validate requested models and report the selected model.
  * Added support for configuration-aware environment settings and authentication checks.
  * Improved compatibility with token usage information returned in multiple formats.
* **Bug Fixes**
  * Improved session loading and model switching behavior.
  * Authentication checks now complete reliably before session snapshots are created.
  * Expanded handling for invalid model selections and alternate usage responses.

### DIFF
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -40,6 +40,26 @@ const SELECT_MODEL_SUPPORT: AcpSupport = {
 };
 const SelectModelDriver = createAcpDriver(SELECT_MODEL_SUPPORT);
 
+/** Proves transformEnv can vary with the instance config, which is how the
+ *  opencode driver picks its permission policy from `fullAuto`. */
+const EnvPolicyDriver = createAcpDriver({
+  ...SELECT_MODEL_SUPPORT,
+  driverKind: "envPolicyTest",
+  selectModel: undefined,
+  transformEnv: (env, config) => {
+    env.TEST_POLICY = config.fullAuto ? "auto" : "ask";
+  },
+});
+
+/** Proves snapshot() awaits an async isAuthenticated, which is how the
+ *  opencode driver answers from a discovered catalog. */
+const AsyncAuthDriver = createAcpDriver({
+  ...SELECT_MODEL_SUPPORT,
+  driverKind: "asyncAuthTest",
+  selectModel: undefined,
+  isAuthenticated: async () => true,
+});
+
 describe("ACP decodeConfig", () => {
   it("grok defaults to the grok binary", () => {
     expect(GrokAgentDriver.decodeConfig({})).toEqual({ cli: "grok", fullAuto: false, workspace: undefined });
@@ -245,6 +265,24 @@ describe("ACP turns (fake CLI)", () => {
     const done = await recorder.until((e) => e.type === "turn.completed");
     expect(done).toMatchObject({ ok: true });
   });
+
+  it("transformEnv sees the instance config", async () => {
+    const dump = join(scratch, "policy.json");
+    process.env.FAKE_ACP_DUMP = dump;
+    instance = await EnvPolicyDriver.create({
+      instanceId: "policy-test",
+      displayName: undefined,
+      environment: {},
+      enabled: true,
+      config: { cli: FAKE_CLI, fullAuto: true },
+    });
+    recorder = recordEvents(instance.adapter);
+
+    await instance.adapter.sendTurn({ threadId: "t-policy", text: "go" });
+    await recorder.until((e) => e.type === "turn.completed");
+
+    expect(JSON.parse(readFileSync(dump, "utf8")).env.TEST_POLICY).toBe("auto");
+  });
 });
 
 describe("ACP snapshot", () => {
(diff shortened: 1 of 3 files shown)


NuCl34R:
Added in `cd0158a`. The finding was correct: the existing unadvertised-model
test rides the fake CLI's `-32602`, so it settles inside `request()` and never
reaches the confirmation guard at all. The guard's own case, an agent that
answers OK and keeps its old model, had no coverage.

Checked both directions rather than just adding a green test. With the guard
neutered, the new test does not merely fail, it reports success on a turn that
ran the wrong model:

```
- "ok": false
+ "ok": true
```

`m-one` ran while `m-two` was requested, and the turn reported itself fine. That
is the silent wrong-model turn the guard exists to stop, which is why an
acknowledged-but-unapplied switch is worth a test that an outright error is not.

`core.ts` is unchanged, this is coverage only. Suite goes from 293 to 294
passed, 8 skipped, 35 files, typecheck clean.

PR #128, opened 2026-08-16
Add peer comms approval gate and async delegate_bot handoff

### LINKED ISSUE #127: Bots can message each other without any user confirmation — peer comms need an approval gate
## Problem

Any bot with the agents integration enabled can call `ask_bot` to contact another bot at any time, with no user confirmation and no durable record of the contact. This is fine in an interactive session, but OpenMausBot also has **routines** — scheduled, unattended runs. A routine-triggered bot can currently:

- message any teammate bot (which may start a paid turn on that bot's provider),
- do so repeatedly, with nothing surfaced to the user until they happen to open the group thread.

There is no per-bot switch to require consent before a bot talks to a teammate, and no way to hand work to a teammate asynchronously (the asking bot blocks on the reply, which is awkward for pipeline-style workflows where a collector bot should hand off to a writer bot and finish).

## Proposal

1. **Per-bot approval gate** — a new `approvePeerComms` flag ("Ask me before contacting other bots" toggle in bot settings). When enabled, `ask_bot`/`delegate_bot` calls pause on a human approval card pushed into the *source* bot's thread, riding the existing options-card + `/api/bots/:id/respond` flow that provider permissions already use — no front-end changes needed to answer one. "Always allow" mirrors into the existing per-bot `alwaysAllow` list with a narrow key (`ask_bot:@Name`). Unanswered approvals time out after 15 min and in-flight approvals are cancelled on server restart, matching provider-permission semantics.
2. **`delegate_bot` — async handoff** — a companion tool that passes a task to a teammate without blocking: the target bot starts its own turn, and the delegation is tracked (pending → done/failed) so the source bot and the user can see the handoff state.

Both are covered by new harness tests (`comms.test.ts`, `delegations.test.ts`).

I have this implemented on a branch and will open a PR referencing this issue.

stephenlzc:
Closes #127

## What

Two related pieces, both harness-native:

**1. Peer comms approval gate** (`server/peer-approval.ts`)
- New per-bot `approvePeerComms` flag + "Ask me before contacting other bots" toggle in bot settings.
- When enabled, `ask_bot` / `delegate_bot` pause on a human approval card in the *source* bot's thread. The card rides the existing options-card flow: answered via `/api/bots/:id/respond`, intercepted by the harness (`resolvePeerComms`) before forwarding to the provider adapter — the front-end needs nothing new.
- "Always allow" mirrors into the existing `alwaysAllow` list via narrow keys (`ask_bot:@Name` / `delegate_bot:@Name`).
- 15-minute timeout on unanswered cards; pending approvals live in memory only and are cancelled on restart, matching provider-permission semantics.

**2. Async `delegate_bot` handoff** (`server/delegations.ts`, `server/comms-visibility.ts`)
- Companion tool for pipeline-style workflows: the source bot hands a task to a teammate and finishes its own turn instead of blocking on a reply.
- Delegations are tracked (pending → done/failed) and surfaced in the shared comms view, so unattended handoffs stay auditable.

## Notes

- `approvePeerComms` defaults to off (current behavior unchanged); users opt into the gate per bot.
- Settings UI: one toggle, follows the existing settings-row pattern.

## Test plan

- `pnpm typecheck` clean
- `pnpm test`: 304 passed / 8 skipped, incl. 351 new lines in `server/comms.test.ts` (approval card lifecycle, always-allow mirroring, timeout) and 310 new lines in `server/delegations.test.ts` (handoff state machine, visibility)
- Manually verified in the packaged macOS app: toggle renders and persists; unattended routine run with the gate disabled completes a full collector → editor handoff

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous task delegation and communication between bots.
  * Added shared visibility for bot conversations, including activity updates and responses.
  * Added optional approval controls for bot communications in bot settings.
  * Added the `delegate_bot` tool for assigning tasks to other bots.
* **Bug Fixes**
  * Prevented unauthorized, recursive, self-directed, missing-target, and overly deep peer interactions.
  * Added handling for denied, busy, or unavailable bot requests.
* **Tests**
  * Expanded coverage for delegation, approvals, communication mirroring, and asynchronous handoffs.

### DIFF
--- a/server/comms-visibility.ts
+++ b/server/comms-visibility.ts
@@ -1,0 +1,94 @@
+// Bot⇄bot comms visibility: channel creation, message mirroring, and
+// per-thread chips. Extracted from /api/internal/ask-bot so delegations
+// (delegate_bot) and any future peer flow reuse the same UX without a copy.
+
+import type { BotRecord, GroupRecord, Message, Store } from "./store.ts";
+
+/** What a peer-exchange helper needs from the outside world:
+ * the store (for persisted messages + groups) and the SSE broadcasters
+ * so chat clients see the change without waiting for a refresh. */
+export interface CommsBus {
+  store: Store;
+  /** SSE broadcast (kind: "message" envelope). */
+  broadcast: (payload: unknown) => void;
+  /** SSE broadcast (kind: "group" envelope) for a single group. */
+  broadcastGroup: (groupId: string) => void;
+}
+
+/** Find or create the bot⇄bot channel for the pair. The channel keeps
+ * the pair's full exchange, lives in the sidebar like any room, and the
+ * user can open it to chip in. */
+export function getOrCreateChannel(store: Store, from: BotRecord, target: BotRecord): GroupRecord {
+  return (
+    store.dmGroup(from.id, target.id) ??
+    store.createGroup(`${from.name} ⇄ ${target.name}`, [from.id, target.id], true)
+  );
+}
+
+/** Mirror `from`'s outgoing message into the channel, drop chips into
+ * both 1:1 threads linking to the channel, and bump the channel's unread
+ * count. The chips are what make bot-to-bot turns observable — those
+ * turns cost the user tokens, and a hidden exchange is exactly the kind
+ * of mistake peer coordination is supposed to avoid. */
+export function mirrorExchange(
+  bus: CommsBus,
+  from: BotRecord,
+  target: BotRecord,
+  message: string,
+  channel: GroupRecord | undefined,
+): void {
+  const note = (threadId: string, m: Omit<Message, "id" | "at">) => {
+    const message = bus.store.appendMessage(threadId, m);
+    bus.broadcast({ kind: "message", threadId, message });
+    return message;
+  };
+  if (channel) {
+    note(channel.threadId, {
+      role: "bot",
+      kind: "text",
+      text: message,
+      from: { botId: from.id, name: from.name, color: from.color },
+    });
+  }
+  note(from.threadId, {
+    role: "bot",
+    kind: "activity",
+    tool: { name: `Messaged @${target.name}` },
+    comm: channel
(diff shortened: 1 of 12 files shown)


milind-soni:
Looks great. Getting it in! Thanks for the contribution

milind-soni:
Merged via #129 — thank you, this is a good feature and the design held up under review: the depth cap still blocks A→B→C, a denied `ask_bot` doesn't hang the caller, and async handoff never regains agents tools.

I fixed three things on top rather than send it back, since you'd built the hard parts:

1. **The approval card was never settled.** `resolvePeerComms` took an `ApprovalBus` it never used (`_bus`), and a harness-native card emits no `request.resolved` — so `answered`/`dismissed` stayed unset, `PendingApproval` kept rendering it, and `Composer` stayed disabled behind it. Turning the toggle on and using it once left that bot's chat unusable, and the card is persisted so a restart didn't help. Cards now settle on answer and on timeout, stale ones are swept at boot, and deleting a bot cancels approvals naming it.
2. **The drained delegation was an uncaught `void startTurn(...)`** — it rejects on busy/deleted/unavailable, and an unhandled rejection is fatal on Node 24, which kills the harness child in the packaged app. Failures are chips now.
3. **Async removed backpressure with nothing in its place** — added a per-turn queue cap, dropped the queue when the turn fails or is interrupted (it used to survive and fire on an unrelated turn), and re-took the `busy` check after the approval await so a 15-minute-old snapshot can't start a turn on a bot that's now mid-turn.

Also made an unresolvable `fromBotId` a 403 — it was skipping both the gate and the mirroring while still running the peer turn — and moved channel/chip creation after the verdict so a denial leaves no trace.

New `server/peer-approval.test.ts` pins the lifecycle; I checked it fails without the settle fix. Suite 374 green.

One follow-up if you're interested: the delegated turn's *reply* isn't mirrored back into the A ⇄ B channel (`mirrorReply` only runs on the ask_bot path), so the channel is request-only for delegations.

PR #159, opened 2026-08-17
Add companion/: a sidecar that lets a paired phone reach the harness

mnthr7:
Part 1 of 3,  Review and merge in order — parts 2 and 3 are stacked on this one, so their diffs will shrink to just their own changes once this lands.

## What changed

A new top-level `companion/` directory: a standalone Node process that lets a paired phone reach the harness, without the harness changing at all.

```
  phone ──LAN/tailnet──▶ companion :8810 ──loopback──▶ harness :8799
                          ▲                             ▲
                          │ token, allowlist,           │ unmodified,
                          │ Origin refused              │ loopback-only
```

Three sockets, and the split between them is the security model:

| Socket | Bind | Who |
|---|---|---|
| `:8810` | `0.0.0.0` | devices — token required, allowlisted, scrubbed |
| `:8811` | `127.0.0.1` | you — pairing and revocation, never off-machine |
| `:8799` | `127.0.0.1` | the harness, spoken to as this machine |

- **Pairing** is a six-digit code shown on the computer, typed into the phone, redeemed once inside a window for a per-device token. Revocation is loopback-only: losing the phone must not mean losing the ability to lock it out.
- **The allowlist is default deny**, per method and path — it lists every request the app makes and nothing else. A route the harness gains later is closed to devices until someone adds it on purpose. Everything else gets `no route`, so a stolen token cannot enumerate the API.
- **A browser is refused before the token is read.** A native app sends no `Origin`; anything that does has found this port and has no business on it.
- **Responses are scrubbed** of harness bookkeeping, on JSON and on SSE alike. The SSE transform emits an event the moment it is complete and never touches the blank-line terminator or the `id:` line.
- **Ports stay clear of the harness**, which owns two: itself, and the webhook receiver one above it. Overlap is refused by name before anything binds, rather than raced for and lost by whichever process started second.
- **Discovery** is a zero-dependency mDNS responder. Failing is not an error anyone has to fix — pairing by typed address still works, and the pairing page says so rather than pretending the list will fill in.

Nothing is wired into the app here. `pnpm companion` runs it, and running it is the opt-in. The Settings toggle is part 2.

## Why

The loopback gate is right, and this does not weaken it. A process holding provider keys and an approval switch should refuse any request whose `Host` is not local — that guarantee is structural, and punching a hole in it for a phone punches it for everyone.

So the phone does not talk to the harness. It talks to a separate process that talks to the harness over loopback, exactly as the desktop window does. The gate is satisfied by construction, and **the harness needs no changes and does not know the sidecar exists** — `server/` is untouched by this PR.

No new runtime dependencies: the mDNS responder, the SSE transform and the device registry are plain Node, in keeping with "don't introduce a dependency where thirty lines will do".

## How it was verified

- `pnpm typecheck` and `pnpm test` — 64 files, 542 passed, 8 skipped. Run on Linux, Node 22.
- `companion/test/proxy.test.ts` boots the **real harness** as a child process and drives it through the **real proxy**, because every bug this design can have lives in the seam between them and none are visible to a unit test: SSE arriving but never terminating an event, a cursor dropped in transit, the loopback gate rejecting a proxied request. It covers pairing end to end — a code issued on the control server, redeemed through the device socket, the token then used and revoked.
- `companion/test/routes.test.ts` tests the allowlist directly, including that a route it has never heard of is denied. Those assertions fail against a permissive gate, which is how I know they are not vacuous.
- `companion/test/ports.test.ts` covers the refusal to bind a port the harness owns.
- By hand on macOS against a live harness: paired an iPhone over Tailscale and used it — chat, a streamed reply, a tool call end to end.

**Not verified, stated plainly:**

- **Bonjour discovery on a LAN.** It has unit tests over the wire format, but in my manual session the machine was already on a tailnet and the browser showed nothing, so I have not watched the discovery path succeed on a plain LAN.
- **Windows.** The sidecar is portable Node with no platform-gated code, but I have not run it there.
- **Device revocation by hand.** Covered by the end-to-end test; not separately clicked through.

## Screenshots (UI changes)

No UI in this PR — it is a headless process. The Settings panel is part 2.

## Checklist

- [x] Server behavior changes come with tests — `server/` is unchanged here; the new code brings its own
- [x] No `dist-server/` edits
- [x] macOS-only code is platform-gated; no `shell: true` / cmd.exe string-building — the one subprocess is `tailscale`, spawned through argv

## Summary by CodeRabbit

* **New Features**
  * Added a companion service for securely pairing devices and accessing approved mobile-app functionality.
  * Added local controls for pairing, viewing connected devices, and revoking access.
  * Added LAN, Tailscale, and Bonjour discovery for easier connection setup.
  * Added request authentication, route restrictions, credential protection, and streaming response sanitization.
  * Added clear handling for unavailable services and port conflicts.

* **Documentation**
  * Added setup, configuration, architecture, limitations, and troubleshooting guidance.

* **Tests**
  * Added comprehensive coverage for pairing, authentication, networking, discovery, proxying, routing, and data sanitization.


mnthr7:
@milind-soni  - this is ready for your review. This is the first one of changes for iOS, PR#160 is second and PR#161 is all the iOS code for the companion app.

PR #160, opened 2026-08-17
Add Settings → Companion, a toggle that starts the sidecar

mnthr7:
## What changed

Settings gains a Companion panel: turn it on, see the address to type into the phone, open a pairing window, list paired devices, revoke them. Turning it off stops the process.

- `electron/companion.mjs` owns the lifecycle via `utilityProcess.fork`, and **waits for the control port to answer** before reporting success rather than assuming the fork worked. A missing `dist-companion/index.js` reports what to run instead of failing as a timeout.
- **The renderer never talks to the control port.** Everything goes through `ipcMain.handle`, which keeps the UI on one origin, avoids CORS, and puts the narrow list of things the renderer may ask for in one file rather than implying it from whatever the control server happens to serve.
- Packaging stages the compiled sidecar beside the harness, and `package:prepare` builds it — so a packaged app has it, and a dev checkout is told to run `pnpm build:companion`.
- `check:electron` now covers `companion.mjs` along with the other plain-JS entrypoints.

There is still no flag left behind claiming a listener that is not there: off means the process is stopped.

## Why

Part 1 is a separate process, and the first version of that meant a terminal command and a browser tab to pair in. That is a bad trade for something the rest of the app does in a panel — and it did not have to be one, because the app already forks the harness as a child process, so forking one more is a thing it already knows how to do.

The panel also **reports rather than guesses**. When there is no MagicDNS name it lists which Tailscale CLI paths were tried and what each said, because telling someone to turn on MagicDNS when they already have it on is worse than saying nothing. (That is what an earlier version did, and it sent me looking in the wrong place.)

## How it was verified

- `pnpm typecheck`, `pnpm test` (64 files, 542 passed, 8 skipped) and `pnpm check:electron` all pass.
- By hand on macOS in the Electron shell: turned the panel on, confirmed the sidecar came up, opened a pairing window, paired a physical iPhone from it, and used the phone against the running harness. Turned it off and confirmed the process stopped.

**Not verified, stated plainly:** the packaged-app path. `package:prepare` builds and stages the sidecar, but I have run this from a dev checkout, not from a built `.dmg`. Worth a look before merging if that matters to you.

## Screenshots (UI changes)

SCR-20260816-lrvt

## Checklist

- [x] `pnpm check:electron` passes
- [x] Server behavior changes come with tests — `server/` is unchanged
- [x] No `dist-server/` edits
- [x] macOS-only code is platform-gated; nothing breaks the packaged app
- [x] UI changes include before/after screenshots — **add before opening**

## New Features
- Added desktop companion support for connecting smartphones to the app.
- Added controls to start or stop the companion service, pair devices, cancel pairing, and revoke access.
- Added pairing codes, device status, network addresses, discovery, and connected-device details.
- Added secure authentication, route protection, credential scrubbing, and local-network connectivity.
- Added automatic startup and packaging integration.

## Documentation
- Added setup, configuration, troubleshooting, and security guidance for the companion service.


mnthr7:
@coderabbitai review

mnthr7:
All review comments are addressed as of 33a56ad. The open threads predate
four commits and describe code that has since changed:

- proxy.ts `headersSent` guard, upstream timeout, SSE backpressure — da93215
- electron lifecycle: serialized transitions, PID-matched adoption, awaited stop — 88c2ee7, a962bd3
- control.ts cross-origin refusal on state-changing methods — 14fec64
- devices.ts case-insensitive Bearer, pairing rollback on failed persist — f0c13bc
- listener.ts enable/disable serialization + late-error handling — a4986f6
- wire.ts bounded SSE buffer — da93215
- test hygiene: awaited dgram sends, probed ports, fake-clock expiry test — 1d595e5, 33a56ad
- waitForExit now takes the signal to send — 33a56ad

Also fixed two CI failures that were pre-existing latent races in test teardown
(EACCES, then ENOTEMPTY), and swept the repo so the pattern is gone from all
five files that had it — ee72fa1.

Not verified: the packaged-app fork path on macOS.

mnthr7:
@milind-soni  - this PR is ready too its the second after PR#159 for the iOS companion app.

PR #166, opened 2026-08-17
feat: per-engine CLI path override with pre-save probe

kargnas:
image
image

## What

Settings → **Engines**: point each engine (claude, codex, grok, …) at a specific CLI binary — a versioned build, a wrapper script (`/opt/homebrew/bin/ag claude agp`), or an absolute path. Drivers already read `config.cli`; this adds the missing API + UI surface.

- **`PATCH /api/instances/:id`** — set / clear (`""` reverts to driver default) an override, persists to `~/.openmausbot/config.json`, hot-reloads the fleet
- **`GET /api/cli-candidates?name=`** — every binary found on the augmented PATH, PATH order
- **`POST /api/cli-test`** — pre-save probe (`<cli> --version`, same PATH a real turn uses). Probe failure blocks the save behind an explicit "Register this path anyway?" — the classic miss is a path the terminal sees but the GUI app can't
- **`Set CLI…` picker UI** — detected-binary dropdown + manual path input, Reset for override removal

Wrapper commands parse in `resolveCliSpawn` (single choke point every driver already crosses), so no driver code changed: fixed args lead the invocation (`ag claude agp --help`), quotes group spaced paths, and an *existing* file at a spaced path wins over the tokenizer (what our own candidates list emits).

## Security fixes found by multi-agent review (all verified against a live server)

| Severity | Issue | Fix |
|---|---|---|
| Critical | `PATCH /api/instances/__proto__` poisoned `Object.prototype` → every engine would spawn an attacker binary | `Object.hasOwn` lookup in `withInstanceCli` → 404 |
| High | `/api/cli-test` with caller argv exfiltrated env vars (`printenv XAI_API_KEY`) | probe runs the head token only |
| Medium | New mutating routes lacked the content-type JSON gate (simple-request bypass) | same gate as local-VM routes → 415 |
| Medium | SIGTERM-ignoring child hung the probe socket forever | `killSignal: SIGKILL` + maxBuffer |

Also fixed: secret env injected by `instanceConfigs()` was being persisted into the `instances` section on save (now stripped — credentials stay in the top-level keys only), unquoted spaced paths splitting into ENOENT, quoted-path re-split, shadow instances missing `cliCandidates`, `resetPathCache()` ordering vs `describe()`, and a set of UI state bugs (save-while-editing, successful save shown as failure, row disappearing after Reset on unknown-driver shadows, select ghost values, `role="alert"`).

## Testing

- `pnpm typecheck` clean; `pnpm vitest run` **456 passed, 8 skipped** (new: config round-trip + secret-strip regression, registry raw-config override detection, `splitCliString`/`resolveCli` wrapper + quoted + spaced-path, HTTP round-trip of all three routes, probe error branches)
- Live-server verification: all four security exploits blocked, real wrapper (`ag claude agp` → `2.1.233 (Claude Code)`) and quoted spaced-path binary run end-to-end, TERM-trapping child times out at exactly 10s
- Not tested on a real Windows box (PATHEXT paths covered by unit tests only)

## Summary by CodeRabbit

* **New Features**
  * Added an Engines settings section for configuring custom CLI binaries per engine.
  * Supports detected or manually entered paths, validation, version checks, saving, and resetting overrides.
  * Displays available CLI candidates and installation guidance when binaries cannot be verified.
  * Supports wrapper commands, quoted paths, versioned builds, and custom CLI arguments.

* **Bug Fixes**
  * Improved persistence and validation of engine CLI overrides.
  * Preserves custom environment variables while removing temporary credential settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ memory of its thread, model, computer, and apps) and rebuilds it open, local-fir
 already have:
 
 - **Bring your own agents.** Bots run on the `claude`, `codex`, and `grok` CLIs installed on your own machine
-  — your existing logins and subscriptions, no new accounts, no proxy in the middle.
+  — your existing logins and subscriptions, no new accounts, no proxy in the middle. Point any engine at a
+  custom CLI binary (a versioned build or wrapper) in **Settings → Engines**.
 - **Local first.** One small harness server on `127.0.0.1` owns every agent process. Transcripts, keys, and
   events live in `~/.openmausbot`, not a cloud.
 - **Agents with hands.** Each bot can get a real computer — a cloud Linux desktop it drives while you watch
--- a/server/config.test.ts
+++ b/server/config.test.ts
@@ -1,6 +1,58 @@
 import { describe, expect, it } from "vitest";
 
-import { instanceConfigs, type AppConfig } from "./config.ts";
+import { instanceConfigs, withInstanceCli, type AppConfig } from "./config.ts";
+
+describe("Instance CLI override", () => {
+  it("sets, replaces, and clears config.cli on a default-fleet instance", () => {
+    const cfg: AppConfig = {};
+    const set = withInstanceCli(cfg, "claude", "/opt/claude-2.1/bin/claude");
+    expect(set.ok).toBe(true);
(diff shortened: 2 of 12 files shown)


milind-soni:
Really good one, Thanks Appreciate it

Issue #173, opened 2026-08-17
Surface Grok CLI local / OpenAI-compatible models in the picker


maxkongerskov:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

Related: #54 asks for OpenAI-compatible endpoints but doesn't specify how they should show up in the picker. This is the concrete version of that.

### Area

Provider drivers (new provider or driver behavior)

### Problem

The Grok driver (`grokAgent`) already does the right thing at runtime: it spawns `grok -m <slug> agent stdio`. Anyone who has local models in `~/.grok/config.toml` can already run them — Ollama, oMLX, EXO, LM Studio, Unsloth, or any `/v1/chat/completions` server.

The picker cannot. `server/drivers/acp/grok.ts` hardcodes:

- `grok-4.6`
- `grok-4.5`

So a signed-in Grok CLI with twenty local slugs still only offers the two cloud models. Switching a bot onto `ollama-ornith-35b-bf16` (or anything else `grok models` lists) requires PATCHing `/api/bots/:id` by hand. The rest of the product — tools, approvals, computer MCP — already works once the slug is set.

That's the gap. OpenMausBot is supposed to be bring-your-own-agent on the models you already have. The catalog is the part that doesn't know about those models.

### Proposal

Don't add a second chat client. Teach the existing Grok instance to publish the models Grok already knows.

1. **Catalog from the CLI / config, not a static list.**  
   `grokAgent` should `resolveModels()` the same way OpenCode Go already does: parse `grok models` or `~/.grok/config.toml` `[model.*]` blocks, keep `grok-4.6` / `grok-4.5` as the cloud defaults, and expose every other slug in `/api/instances` → the model picker. New local models then appear after a refresh, no app release.

2. **Live host chips are optional but cheap.**  
   For known OpenAI-compatible bases (`127.0.0.1:11434`, `:8080`, `:52415` EXO, `:1234`, `:8888`, plus whatever `base_url` is in the config), probe `GET /v1/models` and mark up / down / currently loaded. Don't block picking a down host — just don't pretend it's live.

3. **Add-an-endpoint in App Settings.**  
[...]

Issue #54, opened 2026-08-13
Support openai-compatible endpoints


amirrezasalimi:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Provider drivers (new provider or driver behavior)

### Problem

Please support openai-compatible endpoints

### Proposal

-

milind-soni:
Triage update: #56 is the active provider-manager implementation for OpenAI-compatible endpoints. Leaving this issue open until that PR is reviewed against current main.

kundannett-creator:
This is already supported: `server/drivers/openai-compat.ts` is a generic driver for any OpenAI-compatible chat-completions endpoint. It takes a configurable `url` (base URL) and `apiKeyEnv`, defaulting to OpenRouter's endpoint (`https://openrouter.ai/api/v1`) out of the box, and pulls a live model catalog from the endpoint's `/models` route when a key is set. Any endpoint that speaks the OpenAI chat-completions shape (self-hosted llama.cpp, vLLM, Together, Groq, etc.) should work by pointing `url` at it.

bulyaki:
“Already supported” is a bit misleading here.

Yes, `server/drivers/openai-compat.ts` exists and the backend can clearly talk to arbitrary OpenAI-compatible endpoints. The problem is that the frontend does not expose that capability properly.

If I manually add an `openai-compat` instance in `config.json`, the app does recognize it, but then the model picker classifies it as a **local-model provider** because `access: "custom"` is treated as “Local” in the UI. The result is that a perfectly valid cloud endpoint such as ClinePass appears under:

“Local models”
“No local models found”
“Start oMLX, Ollama, Unsloth, LM Studio, or EXO”

So the backend feature may technically exist, but from the user’s point of view it is not really implemented.

This feels like one of those half-finished, vibe-coded integration paths where the backend abstraction was added, the frontend made a simplifying assumption, and nobody tested the whole flow end-to-end with an actual remote OpenAI-compatible provider.

I would call this “backend support exists, frontend support is incomplete” rather than “already supported.”

vadim-a-yegorov:
@amirrezasalimi This is a bug, not a feature request. The openai-compatible backend exists, as well as the UI to add it, but UI malfunctions. Please rewrite the issue accordingly and remove the enhancement tag.

PR #174, opened 2026-08-17
List Grok CLI local models in the picker

maxkongerskov:
## Why

`grokAgent` already spawns `grok -m <slug> agent stdio`. Local models already run — tools, approvals, and computer MCP included. The picker could not see them, and Custom was empty on every other agent because those CLIs do not share Grok's config.

Fixes the concrete half of #54. Proposed in #173.

## What

- Official cloud rows stay on the main picker. **Custom** is pinned at the bottom of every provider.
- Custom probes the same local hosts the sidecar did (oMLX, Ollama, Unsloth, LM Studio, EXO) and lists whatever they serve.
- Picking a Custom model **injects it into the selected agent** for that bot:
  - Grok: write a `[model.slug]` block and pass `-m`
  - Codex: `model` + `modelProvider`
  - Claude: `ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` / `--model`
- Rail icons stay clickable without a cloud login. Official cloud rows can still show sign-in help.

## How to try

1. Create a bot
2. Open the picker and select an agent
3. Open **Custom** and pick a live local model
4. That model is injected into that agent for that bot

### DIFF
--- a/server/contracts.ts
+++ b/server/contracts.ts
@@ -220,7 +220,7 @@ export interface EngineInstall {
 // a rejection to an unavailable shadow snapshot.
 export interface ModelCatalog {
   default: string;
-  options: Array<{ id: string; label: string }>;
+  options: Array<{ id: string; label: string; custom?: boolean }>;
 }
 
 export interface DriverCreateInput<Config> {
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -615,8 +615,8 @@ describe("ACP snapshot", () => {
     try {
       // favourites first in the user's own order, then the built-in slice
       expect(instance.models.options.slice(0, 2)).toEqual([
-        { id: "custom:Azure-Opus-0", label: "Azure Opus" },
-        { id: "custom:LMStudio-Qwen-0", label: "Qwen (local)" },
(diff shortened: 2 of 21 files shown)


maxkongerskov:
Pushed a second commit so the picker can scroll. With 20+ local slugs the old dropdown clipped (`overflow-hidden`, no max-height), which made the catalog unusable.

I've been running this as a packaged macOS build for a bit — `grok -m <slug>` already did the work; this just lets you pick the slug in-app. Still happy to change anything on the catalog parse.

maxkongerskov:
@milind-soni CodeRabbit’s findings are addressed (only honor `[models].default`, temp-dir cleanup). The review is green.

This is ready for a human look when you have a minute — happy to change the catalog parse if you want a different shape.

maxkongerskov:
Pushed a follow-up so the original picker look stays intact: official models stay on the first screen, and hosted slugs the CLI already knows live under a **Custom** row that opens a scrollable list. Same pattern for Grok, Droid customModels, and extra Kimi entries — no Custom row if an engine has none.

milind-soni:
This is looking great. Merging it in shortly. Thanks a lot! Appreciate it

PR #186, opened 2026-08-17
Search messages across every bot and room

aivsomkar:
## In plain language

**Before:** the sidebar search only filtered bots by name. A phrase said three weeks ago in some bot's task was unfindable.

**What changes in the app:** type two or more characters in the sidebar search and a **Messages** section appears with hits from every bot's tasks and every room — bot, task title, highlighted snippet, time. Tool runs are searchable too ("which bot ran that migration"). Click a hit and you land **on that message**: right bot, right task, right version of the conversation (it switches the branch only if the hit is on an old version), scrolled to and briefly highlighted.

---

## In plain terms

A roster of bots, threads that live for weeks, and no way to find anything — the sidebar search only filtered bots by name. Now it also finds **messages** across every bot's tasks and every room, and clicking a hit lands you on that message: right bot, right task, right branch.

## Changes

- **`server/search.ts`** + `GET /api/search?q=&bot=&kind=&limit=` — linear scan over every bot task and room thread (threads are already in memory once touched; an index would be premature). Matches text and, for activity chips, the **tool name** — so "which bot ran that migration" works. Each hit: owner, task title, role/kind, windowed snippet with match offset, `onActivePath`.
- **Sidebar**: from 2 characters, a **Messages** section under the bot list — avatar, task title, highlighted snippet, time; hits off the visible branch get an *"other version"* tag. Debounced 250 ms.
- **Landing on a hit**: select bot/room → switch task if it's another one → `POST /active-branch` **only when** the hit is off the visible branch (that rewinds the leaf and marks the provider session stale, so it's never done idly) → scroll to and flash the message. Search box clears.
- `focusMessage` state + `useFocusMessage` hook; message rows carry `data-mid` (they're `display: contents` wrappers, so the hook scrolls/highlights the child). Works in ChatView and GroupView.

Not in scope: fuzzy/regex, an index, searching inspector logs, in-thread find (⌘F).

Item 1.7 of `docs/plans/agent-harness-upgrades-v2.md`.

## Test plan

- [x] `server/search.test.ts` (6): case-insensitive across bots and rooms newest-first; active-path flag on a branched thread; tool-name + text search for activity with kind filter; bot filter + limit; snippet windowing/ellipsis and match offset; empty query
- [x] `pnpm typecheck` clean; `pnpm vitest run` green (65 files, 533 passed)
- [x] Manual on real data: `yin` finds hits in two of one bot's tasks; clicking switches task and scrolls/flashes; activity hits show the wrench; hits off the visible branch switch versions


milind-soni:
Superseded by #213, which preserves this PR’s useful transcript search and exact-message landing while integrating it with the newer Cmd+K palette and bounded transcript windows. #213 also addresses the review findings around stale focus replay, deleted conversations, fallback highlighting, and duplicate landing logic, and has now merged green.

PR #192, opened 2026-08-17
Turn-engine hardening, per-bot Composio gate, and per-bot workspaces with file memory

milind-soni:
Implements the P0/P1 items from the Aug 17 codebase assessment plus the file-based memory design, on top of latest main.

## What's in here

**Reliability — the turn engine can no longer wedge or double-run a bot**
- `server/turn-watchdog.ts`: activity-based stall detection. A dispatched turn whose thread emits *nothing* for `OMB_TURN_STALL_MS` (default 20 min) is interrupted and settled with an error chip. Turns parked on a human approval card are exempt, and any provider event resets the clock — long turns that keep streaming are never touched. (ask_bot and rooms already had ceilings; the main 1:1 path had none.)
- Rooms now respect **and set** `bot.busy`: a bot can't run a 1:1 turn and a room turn concurrently anymore (the second gets a visible "busy — skipped this round" chip), and `POST /api/bots/:id/interrupt` now also reaches a turn running on a room's thread.

**Security — remaining items from the audit**
- **Per-bot Composio gate** (`bot.composio`, PATCH-able boolean): the workspace key no longer reaches every bot on every turn. Existing/user-created bots keep access (unset = allowed); **imported team members start with it off**.
- Teams import no longer seeds a greeting (`createBot({seedMessages:false})`).
- Persona fields are bounded at the write boundary: PATCH `/api/bots` rejects name >100, title >200, description >4000 (matching the team-manifest caps), and the chief-of-staff roster clips name/role/about (80/120/200) and caps the roster at 40 bots with an explicit overflow line.
- `timingSafeEqual` for the internal comms bearer token.

**Product — per-bot workspaces + simple-files memory**
- Every CLI-engine turn now runs with `cwd` = `~/.openmausbot/workspaces/<botId>/` instead of `$HOME` — a bot with file tools and acceptEdits gets a desk, not the whole house. (API `grok` and `boxAgent` are excluded; they have no local-filesystem story.)
- The workspace is the bot's memory: `MEMORY.md` is injected into the system prompt every turn under a hard budget (first 200 lines / 24 KB), `memory/` holds topic files the bot reads on demand with its ordinary file tools, and the prompt teaches curation plus a taint rule (never record claims arriving from other bots/webhooks/imported files). Plain markdown — the user can open, edit, or delete anything the bot believes. Deleting a bot deletes its workspace.
- Rooms inject the speaking bot's memory too, so a bot is the same colleague in a room as in its 1:1.

**Observability**
[...]

PR #193, opened 2026-08-17
Turn liveness (2.1) and portable context (2.2)

aivsomkar:
## In plain language

Two things, both about what happens when a bot is running or when its conversation moves to a different engine.

**1. If a bot's engine goes silent, the app says so (2.1).**
Sometimes an engine just stops responding — a stalled request, a stuck prompt. Before, the bot sat on "Working…" with no hint anything was wrong.
- *What changes:* after ~2 minutes of silence you'll see an amber **"Quiet for 2m"** pill next to Stop, and the sidebar preview says the same. It's a heads-up, not a verdict — a long silent tool run looks identical from outside — so it sits next to Stop and lets you decide. A bot waiting on *you* (an open approval card) is not counted as quiet.
- **Stopping** a truly wedged turn is done by the stall watchdog that landed on `main` in #192 (no activity for 20 minutes → the turn is stopped with a note). This PR was originally going to stop unattended turns itself; after #192 there is one clock for stopping and this PR only adds the early "quiet" signal on top of it.

**2. Handing a conversation to a different engine now carries the whole story (2.2).**
When you switch a bot's model, edit an old message, or a room member takes a turn, the engine has no memory of the thread and the app has to replay it. Before, it got the last 40 text lines — no record of tools that ran — regardless of whether the target model could hold 8k or 200k tokens.
- *What changes:* the engine now gets a **summary** of the older part plus the **recent messages verbatim** (including what tools ran), sized to that model. You'll see a thin **"Context compacted · earlier messages are still here"** line in the chat; click it to read the summary. **Nothing is deleted** — scroll up and every message is still there. Continuing on the same engine is unaffected (the CLI keeps its own memory).

---

## Summary

Two Round-2 harness upgrades on one branch, as agreed. This branch also carries #180's `buildTurnContext` seam (merged in, since 2.2 builds on it) — once #180 lands on `main`, this diff shrinks to just 2.1 + 2.2.

---

## 2.1 — Notice when a running turn goes quiet

Rescoped twice. v1 called this a "liveness reaper" for driver processes that die silently — every child-process driver already settles on `close`, so the gap was **hung** turns. Then `main` (#192) landed a `TurnWatchdog` that stops any turn with no activity for 20 minutes (waiting-on-human exempt). So this branch keeps only the part that isn't on `main`:

- **`server/liveness.ts`** — the early, soft signal: quiet ≥ **2 min** → **flag** (`quietSince` on the bot broadcast, transient like `busy`); events resume → **clear**; a card reaching a human pauses the clock (same exemption the watchdog applies). It never stops anything.
- **UI** — amber *"Quiet for 3m"* pill next to Stop (re-renders every 30s), sidebar preview says the same and keeps counting. Detached routine/webhook tasks surface on the bot too (`wireBot` looks across every task thread). A forced provider reload settles the clock so nothing is flagged after the fact.
- Env override `OMB_TURN_QUIET_AFTER_MS` for tests/dev.

## 2.2 — Rebuild a thread's context for the model that will read it

Every path that hands a thread to a model without a native session — a fresh engine after a switch, a rewind, a transcript-replay driver, a room member's turn — got the last 40 **text** messages or nothing. Tool activity was dropped: a rebuilt context showed people talking with no evidence of work performed. And 40 is not a size; the same thread may be handed to an 8k local model and a 200k Claude.

**Governing rule:** two transcripts, only one shrinks. The display path is never touched, nothing is deleted. Only the model-facing rebuild is bounded — and bounded **for the target**.

- **`context-rebuild.ts`** — activity renders compactly (`[tool: Bash ✓]`); budget = 40% of the target window (floor 4k; unknown → 128k, scaled rather than a flat reserve so a small window isn't permanently over the line); fold planning always keeps the newest 12 entries verbatim.
- **`context-window.ts`** — `contextWindow` on catalog entries when a driver declares it (new optional field), else a pattern table over the model id.
- **`compactor.ts`** — when the rebuild won't fit, the older part is summarized through `generateText` (the bot's own engine, else any that has one; the previous summary is fed into the next so facts carry forward) and a **compaction record** — a message in the tree with a `parentId` like any other — marks `firstKeptId`. No summarizer, or one that fails → bound by dropping the oldest, write nothing; never fails the turn.
- **`store.appendCompaction` / `modelContext`** — the model-facing view is summary + messages from `firstKeptId`; a rewind past the record no longer sees it.
- **Dispatch** rebuilds only when it will actually replay (fresh / rewound / transcript driver) — a resumed turn pays nothing. grok's `transcript` leads with the summary. **Rooms** use the same rebuild sized for the responder's model, attributed — and now read the active path (they read `messagesFor` before, so a rewound room replayed the wrong branch).
- **UI** — a quiet *"Context compacted · earlier messages are still here"* divider, expandable to the summary, in chats and rooms. `OMB_CONTEXT_WINDOW` forces a window for dev/test.

Measurement is chars/4 for now; anchoring on provider-reported usage waits for #185's per-turn usage to be on `main`.

Items 2.1 and 2.2 of `docs/plans/agent-harness-upgrades-v2.md`.

## Test plan

- [x] 2.1: `liveness.test.ts` (6, fake clock incl. the waiting-on-human pause) + `liveness-e2e.test.ts` (3, real server + fake ACP `hang` mode: interactive turn flagged and still running, Stop clears; a scheduled routine in a detached task and a webhook run are flagged, then stopped by main's watchdog with its chip)
- [x] 2.2: `context-rebuild.test.ts` (9), `context-window.test.ts` (2), `compactor.test.ts` (6: fits → nothing written; small window → summary + record + display path intact; previous summary fed forward; no summarizer → drop-oldest, no record; failing summarizer → same; excludeId), `store.test.ts` (+3), `turn-context.test.ts` (+2)
- [x] `pnpm typecheck` clean; `pnpm vitest run` green (70 files, 568 passed)
- [x] Manual: 137-message thread on Claude → switched to Antigravity with `OMB_CONTEXT_WINDOW=12000` → compaction divider appears, reply reflects the whole conversation and the tools used; back to Claude → same; scrolling up reaches message one. Quiet pill appears during a silent tool run and clears on reply.

## Summary by CodeRabbit

- **New Features**
  - Added quiet-state indicators showing how long active bot tasks have been inactive.
  - Added notifications when tasks become quiet or are automatically stopped after stalling.
  - Added collapsible summaries for compacted conversation history.
  - Improved conversation continuity when switching models, resuming tasks, or rewinding conversations.
  - Added model-aware context handling to preserve relevant recent history within limits.
- **Bug Fixes**
  - Prevented stale activity status from persisting after provider reloads or completed tasks.
  - Improved group conversation context with attributed, compacted history.


milind-soni:
Thank you for this — turn liveness and portable context were both real gaps when this was written, and the harness-2.1 write-up shaped how they were eventually solved. Re-checked today: the branch is 1,102 commits behind `main`, and `main` has since covered the same ground by other routes — `server/turn-watchdog.ts` for liveness, #180 and #205 for the pieces around it, and the context work is being split from #759's rewrite into small PRs on current `main` (#1038 is the first). There is nothing left here that would merge as-is, so closing it with thanks rather than leaving it to age. If any specific idea from it is still missing on `main`, a small PR against current `main` would be welcome.

PR #197, opened 2026-08-17
Add a floating cursor above Automations to clear the team

maxkongerskov:
## Why

Clearing every bot should not need a trash icon or a “delete bots” label in the sidebar.

## What

- A still brand cursor hangs above Automations.
- Click it: a short press, then it becomes **delete all bots?**
- Click again to confirm. Escape or a click outside fades the question back to the cursor without squashing the text.

## How to try

1. Have at least one bot.
2. Click the cursor above Automations.
3. Click away — the label should fade out in place.
4. Open it again and click the question to delete every bot.

### DIFF
--- a/src/components/ClearTeamSprite.tsx
+++ b/src/components/ClearTeamSprite.tsx
@@ -6,14 +6,16 @@ import { cn } from "@/lib/cn";
 import { useStore } from "@/state/store";
 import { MausAvatar } from "./Avatar";
 
-const ARM_MS = 320;
+const ARM_MS = 280;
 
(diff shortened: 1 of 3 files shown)


milind-soni:
Closing after review. A bulk-clear action could be useful, but this version hides a destructive operation behind a mascot and dispatches one independent delete per bot. Any mid-flight failure can leave a partially deleted team, and there is no undo or recoverable archive path. If we bring this back, it should live in explicit team management with a clear typed confirmation and an atomic or recoverable server operation.

PR #212, opened 2026-08-18
feat: add MiniMax driver (M3, OpenAI-compatible API)

collectivewinca:
## Why

MiniMax M3 is a useful coding/agentic model, and several of us already have a key in `~/.minimax/user-settings.json`. OpenMausBot had no driver for it, so a `driver: "minimax"` instance stayed shadowed as unknown.

## What

- Adds `server/drivers/minimax.ts`: OpenAI-compatible `/chat/completions` with SSE streaming, same shape as the existing API-key `grok` driver.
- Registers `MinimaxDriver` in `server/drivers/builtIn.ts` **without removing** Qwen, Hermes, Kimi, Droid, or Antigravity.

Key resolution order: instance `environment`, `MINIMAX_API_KEY`, then `~/.minimax/user-settings.json`.

## Note on a prior PR

There is an earlier attempt on our fork only: https://github.com/collectivewinca/OpenMausBot/pull/1 — that branch is 314 commits behind `main` and its `builtIn.ts` would drop drivers added since mid-August. This PR is rebased onto current `main`. Please ignore / do not merge the fork-only PR.

## Config

```json
{
  "instances": {
    "forge": {
      "driver": "minimax",
      "displayName": "Forge — Dev & Automation",
      "environment": { "MINIMAX_API_KEY": "…" }
    }
  }
}
```

## Summary by CodeRabbit

* **New Features**
  * Added MiniMax as a supported AI provider.
  * Supports global and China endpoints, multiple MiniMax models, and API credentials via `MINIMAX_API_KEY`.
  * Added synchronous and streaming responses with usage reporting.
  * Added request cancellation, connection status updates, response logging, and session handling.
  * Supports credential discovery through environment settings or local MiniMax configuration.


milind-soni:
Thanks! will get it in

milind-soni (commented):
Nice addition — follows the grok.ts transcript-replay/SSE shape closely and reads clean. A few things to address:

1. **`snapshot()` makes a live network call on every poll.** The registry calls `snapshot()` on every describe/refresh (`server/harness/registry.ts`), so with this driver every sidebar refresh hits `/models` with a 5s timeout — offline or rate-limited users will see laggy/unavailable-flapping state for an instance that's actually fine. Suggest caching the connectivity result (e.g. re-probe on demand or on turn start) instead of probing per snapshot.

2. **No key path through the app's own config.** Every other driver can be configured via `~/.openmausbot/config.json` (`appConfigSchema` + `injectedEnvironment` in `server/config.ts`). MiniMax only resolves from instance env / process env / `~/.minimax/user-settings.json`, so desktop users who configure other providers in-app have no equivalent here. Consider adding a `minimax: { key }` block like #348's PR did for openaiCompat.

3. **`respondToRequest` throws** where sibling text-only drivers return `"unavailable" as const` (see `server/drivers/grok.ts:207`). It's caught in `answerRequest`, but matching the convention keeps error paths uniform.

4. **No tests.** Even a small decodeConfig/defaultConfig test file like `openai-compat.test.ts` would help; CI also hasn't run on this PR yet.

Non-blocking: `generateText` hardcodes `"minimax-01"` while turns default to `MiniMax-M3` — intentional? And the model list is static; if MiniMax exposes `/models` you could populate it lazily like the OpenRouter driver.

milind-soni:
Updated this PR against current main and addressed the review findings. The driver now follows the official `mmx-cli` config (`~/.mmx/config.json`), current MiniMax-M3/M2.7 models, combined cancellation + timeout, safe SSE cleanup, non-probing snapshots, prompt-safe native logs, uniform request handling, and six focused tests.

Verified with 1,671 passing tests (12 skipped), typecheck, targeted lint, broker/updater/viewer checks, and packaged-server smoke. The Vercel status is an external-fork authorization failure; this PR does not change docs.

milind-soni:
@coderabbitai review

PR #229, opened 2026-08-18
fix(claude): stop zombie permission cards after turn teardown

### LINKED ISSUE #211: Turn teardown discards the permission broker without ensuring child CLI/MCP exit — late approval requests become dead "zombie" cards
## Summary

When a bot turn ends (the harness receives the CLI's `result` frame), the app tears down the permission broker and removes the turn from `active`, but does **not** ensure the spawned `claude -p --resume` child (and its MCP grandchildren) have actually exited. A still-running child can then emit a late permission request, which surfaces in the UI as an unanswerable card ("Auto mode couldn't answer this one." — clicking Allow yields `no active turn with a permission broker`). The leaked processes also linger indefinitely (observed: 50+ minutes, idle CPU) until manually SIGTERMed.

## Environment

- OpenMausBot 0.1.23, macOS (Apple Silicon, Darwin 25.x)
- Bots backed by the `claude` CLI (`claude -p --resume …`)
- Reproduction is promoted by the child doing backgrounded work (`run_in_background`, scheduled wake-ups), but the root cause is teardown-side.

## What we observed (timeline)

1. Bot turn starts → harness spawns `claude -p --resume` with `--permission-prompt-tool mcp__ogb__approve`; a `perm-<thread8>.sock` appears under `~/.openmausbot/`.
2. CLI emits its `result` frame → harness logs `turn.completed` in `~/.openmausbot/events/<tid>.ndjson`, destroys the broker, deletes the active entry.
3. The child process tree is still alive (verified via `ps`: same PID/start-time, plus MCP children). No SIGTERM/SIGKILL is ever sent.
4. Minutes later the orphan issues a permission request → UI shows a card that can never be answered. Left overnight, one dead turn re-issued its permission request 5 times on the ~15-minute auto-deny timeout cycle before finally going away.

We run a 60s dry-run observer that fingerprints these orphans by command-line marker + descendant `perm-*.sock` + process group id; sample JSONL observations (PID/PGID/PPID, start time, thread prefix, `busy` flag, socket presence, `turn.completed` age) can be shared on request.

## Why this looks app-level (not only an upstream CLI issue)

The `claude` CLI does have known lifecycle issues of its own (e.g. anthropics/claude-code#25180, anthropics/claude-code#21287, anthropics/claude-agent-sdk-python#610), but the harness is the spawner and owns the lifecycle. Two independent defects compound here:

1. **Process teardown** — `result` received → broker destroyed, but child exit is neither awaited nor forced.
2. **Late-approval acceptance** — permission requests arriving after `turn.completed` are still surfaced as actionable cards instead of being dropped.

Either fix alone removes the user-visible zombie card; both together remove the leak.

## Proposed fix (3 points)

1. **Await/force child exit on terminal state.** On `result`: keep the child handle, `SIGTERM` the child's process group, grace (e.g. 5s), then `SIGKILL`. Spawning with `detached: true` makes the child its own group leader, so the group signal (`process.kill(-child.pid, sig)`) reaps MCP grandchildren too — macOS has no `PR_SET_PDEATHSIG`, so the app has to do this itself.
2. **Drop late permission requests.** Any permission request whose turn/session id has no live broker should be logged and dropped (or auto-denied), never rendered as an actionable card.
3. **Startup sweep.** On harness start, scan for processes carrying the app's marker (`--permission-prompt-tool mcp__ogb__approve`) whose thread already has a final `turn.completed`, and reap them — recovers from crashes where teardown never ran.

Related: #127 — point 2 is effectively a staleness gate on the same approval path.

Happy to share reproduction logs (PID/PGID timelines, events NDJSON excerpts); we can reproduce reliably.

SomSamantray:
## What changed

`server/drivers/claude.ts`'s `settle()` now kills the spawned `claude -p --resume` process tree on every terminal path (result, error, or exit-before-result), and `createPermissionBroker()` now replies with an explicit system-source deny/answer to any permission or question ask that arrives after the broker has closed, instead of either registering it as a new, permanently unanswerable card or leaving the caller's MCP tool call hanging.

## Why

`settle()` closed the permission broker and forgot the turn (`active.delete(threadId)`), but never killed the child — a one-shot process is expected to exit right after printing `result`, but a backgrounded MCP grandchild can keep it alive. `net.Server.close()` only stops accepting *new* connections; it doesn't touch one that's already open. So a still-alive child's MCP proxy could raise a fresh permission ask on that still-open connection, which the connection's handler happily turned into a new `request.opened` card — for a turn the driver had already forgotten. Clicking Allow on that card then hit `no active turn with a permission broker`, because `respondToRequest` looks the broker up by `threadId` and finds nothing. That's the "Auto mode couldn't answer this one" zombie card in #211, with the process itself lingering (observed 50+ minutes) until manually killed.

Fixes #211.

### Known trade-offs (not blocking, flagged for visibility)

- `killCliTree(child)` now runs unconditionally on every successful turn too, not just on interrupt/error. This is deliberate — a plain `SIGTERM` to the process group (POSIX) / `taskkill /T /F` (Windows) is a no-op when the process already exited, and it's the only way to guarantee a backgrounded grandchild doesn't outlive its turn. The trade-off: a well-behaved CLI doing brief async cleanup right after printing `result` is now cut off a beat earlier than before, on every turn rather than only on error/interrupt.
- Adversarial review during this fix surfaced a **separate, pre-existing** bug: two asks sharing the same `id` on one connection can silently overwrite each other's `pending` entry, producing an unanswerable card through a different path than the one this PR fixes. Out of scope here — filed separately, not touched in this PR.

## How it was verified

`pnpm typecheck` and the full `pnpm vitest run` suite pass locally (973/973 tests, 3 new). Both fixes were verified with genuine red-before-green evidence: each was manually reverted in isolation, the corresponding new test was confirmed to fail (the child stayed alive; the late-ask test timed out waiting for a reply that never came), then the fix was reapplied and the suite re-run green.

New/extended coverage in `server/drivers/claude.test.ts`:
- A new `result-then-hang` fake-CLI mode (prints `result` but never exits) plus a `pid` field added to the existing `FAKE_CLAUDE_DUMP` test-harness payload, used to assert the underlying OS process is actually gone after `settle()` — not just that `turn.completed` fired.
- A late-ask test: open a connection, interrupt the turn, await `turn.completed`, then send a second ask on the *same still-open connection* — asserts no new `request.opened` fires and the connection gets a real deny reply instead of hanging.
- The same scenario for a `question`-kind ask, confirming the `answer` (not `deny`) branch.

## Screenshots (UI changes)

N/A — server-only change, no UI surface touched.

## Checklist

- [x] No `dist-server/` edits
- [x] macOS-only code is platform-gated; no `shell: true` / cmd.exe string-building — unchanged; this PR doesn't touch spawn/shell handling
- [x] No secrets in logs, responses, events, or argv — unchanged; no new logging or event fields carry credentials

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Claude sessions that could leave child processes running after producing a result.
  * Late permission requests now receive an explicit denial instead of creating new pending requests.
  * Late question requests now receive a clear system response.
  * Prevented additional responses from being accepted after a request broker has closed.

* **Tests**
  * Added regression coverage for process cleanup and late request handling.

* **Documentation**
  * Documented the planned approach for resolving Claude session teardown issues.

### DIFF
--- a/docs/plans/2026-08-18-001-fix-claude-turn-teardown-zombie-cards-plan.md
+++ b/docs/plans/2026-08-18-001-fix-claude-turn-teardown-zombie-cards-plan.md
@@ -1,0 +1,135 @@
+---
+artifact_contract: ce-unified-plan/v1
+artifact_readiness: implementation-ready
+execution: code
+product_contract_source: ce-plan-bootstrap
+---
+
+# fix: Kill the child process and drop late asks on Claude turn teardown
+
+**Origin issue:** [milind-soni/OpenMausBot#211](https://github.com/milind-soni/OpenMausBot/issues/211) — "Turn teardown discards the permission broker without ensuring child CLI/MCP exit — late approval requests become dead 'zombie' cards"
+
+---
+
+## Summary
+
+When a Claude-driven turn ends, `server/drivers/claude.ts`'s `settle()` tears down the permission broker and forgets the turn — but never kills the spawned `claude -p --resume` child. If the child (or an MCP grandchild doing backgrounded work) doesn't exit on its own, it can later emit a new permission ask on its still-open broker connection. Because `net.Server.close()` doesn't touch already-open sockets, that ask is still processed and surfaces as a `request.opened` card — but the turn's `active` entry (and the broker reference the UI needs to answer it) is already gone, so the card can never be resolved. The fix makes `settle()` unconditionally kill the child's process tree and makes the broker's `close()` actually stop honoring asks on any connection, closed or not.
+
+## Problem Frame
+
+`sendTurn()` in `server/drivers/claude.ts` spawns the `claude` CLI per turn and wires a `createPermissionBroker()` instance to a per-turn unix socket (named pipe on Windows). The CLI's own spawned MCP `ogb` process (`server/permission-proxy.ts`) forwards `approve`/`ask_user` tool calls over that socket to the broker, which the harness renders as `request.opened` cards.
+
+Two independent gaps compound into the reported symptom:
+
+1. **The child is never killed on teardown.** `settle()` (invoked from the `result` stream-json frame, a spawn `error`, or an exit-before-`result` `close`) deletes the turn from `active` and closes the broker, but nothing calls `killCliTree(child)` — that utility is only wired to `interruptTurn` (an explicit user stop). A `-p` one-shot CLI process is expected to exit right after printing `result`, but if it (or a grandchild doing `run_in_background` work) doesn't, it keeps running with no teardown-side enforcement.
+
+2. **A closed broker still answers still-open connections.** `createPermissionBroker.close()` calls `server.close()`, which per Node's `net` docs "stops the server from accepting new connections" but does **not** touch sockets that already connected. The `conn.on("data", ...)` handler registered in `createNetServer((conn) => {...})` stays fully wired to any live connection — a lingering child's MCP proxy can still send a `{t:"ask",...}` message, which the handler happily adds to the (still-referenced-via-closure) `pending` Map and forwards via `opts.onAsk(ask)`, emitting a brand-new `request.opened` event. But `active.delete(threadId)` already ran inside `settle()`, so `respondToRequest` finds no broker for that thread and returns `"unavailable"` — the exact unanswerable "Auto mode couldn't answer this one" card the issue describes.
+
+**Prior art in this codebase:** `server/drivers/antigravity.ts` already tracks live children independently of `active` (a `children: Set<ChildProcess>`) and reaps them via `reapChildren(escalate)` — `killCliTree` first, then an optional SIGKILL after a 2s grace on POSIX when `escalate` is set. That mechanism is only invoked from `stopAll()`/`dispose()` (whole-driver shutdown), not per-turn, so it does not by itself close this gap — but its escalation shape is the right reference for this fix.
+
+## Requirements
+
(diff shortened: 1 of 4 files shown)


milind-soni (commented):
Excellent work here — the plan doc alone (KTD2's catch that `permission-proxy.ts` has no per-ask timeout, so a silent drop hangs the MCP tool call forever) is genuinely good analysis, and I verified the implementation locally on your branch: all 30 tests pass (`pnpm exec vitest run server/drivers/claude.test.ts`) and `pnpm typecheck` is clean.

However, **main has moved under this PR in a way that supersedes both halves of the fix**:

1. **Fix B (late asks → zombie cards): already landed on main, in stronger form.** Current `server/drivers/claude.ts` has the `closed` flag + `systemEndedReply()` you designed (claude.ts:254-267, 362-367), *plus* an `isActive()` gate (claude.ts:295-302) that rejects late/background asks not just after close but **between turns** — which matters because main no longer tears the broker down per turn at all.

2. **Fix A (kill child on settle): would now be wrong.** Main's driver retains sessions/processes across turns (`armIdle()`, `session.broker?.pause()` at claude.ts:726). Unconditionally killing the child in settle would break process retention entirely.

All four files this PR touches conflict with current main. Given both halves are either already shipped or invalidated by the redesign, I'd recommend closing this rather than rebasing — but the analysis in the plan doc is worth preserving if issue #211's "startup orphan sweep" follow-up ever gets picked up.

milind-soni:
Closing as superseded by the retained-session Claude redesign on main. The stronger late/background-request guard is already present, while killing the CLI tree on every settle would now break intentional process reuse. The plan analysis remains useful for the separate startup-orphan-sweep follow-up.

PR #230, opened 2026-08-18
fix: reject colliding ask ids in the permission broker

SomSamantray:
## What changed

`createPermissionBroker()`'s socket handler now rejects an incoming permission/question ask whose `id` already has a live entry in its `pending` map, instead of silently overwriting it.

## Why

`pending` is scoped to the whole broker, not to one connection. Two asks arriving with the same id — a buggy or adversarial client, since the CLI's own permission-proxy always mints a fresh UUID per ask — would let the second `pending.set()` clobber the first. Once the *original* ask resolved, its `finish()` deleted the shared key by id and took the second, still-visible request card down with it, leaving it permanently unanswerable. The guard now denies the colliding ask immediately, on the wire, before it ever reaches `onAsk` — so it can no longer orphan a legitimate request or open a second UI card for the same id. Confirmed via `server/permission-proxy.ts` that no legitimate retry path reuses an ask id, so this is a defense-in-depth fix rather than a response to an observed collision.

## How it was verified

- `npx vitest run server/drivers/claude.test.ts` — 31/31 pass, including 3 new regression tests: same-connection collision, cross-connection collision (the case that motivated this fix, since `pending` is broker-wide), and reuse of an id after the original already resolved (confirmed as *not* a collision).
- `npx tsc --noEmit` — clean.

## Screenshots (UI changes)

N/A — backend-only change to the permission broker's socket handling.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate active permission requests from being processed simultaneously.
  * Duplicate requests are denied immediately with a clear message.
  * Original pending requests remain intact, and IDs can be reused after resolution.
  * Improved duplicate handling across the same and separate connections.
  * Applied protection consistently to permission and question requests.

* **Documentation**
  * Added implementation and verification guidance for duplicate request protection.

### DIFF
--- a/docs/plans/2026-08-18-001-fix-permission-broker-ask-id-collision-plan.md
+++ b/docs/plans/2026-08-18-001-fix-permission-broker-ask-id-collision-plan.md
@@ -1,0 +1,124 @@
+---
+title: "fix: Reject colliding ask ids in the permission broker"
+plan_type: fix
+artifact_contract: ce-unified-plan/v1
+artifact_readiness: implementation-ready
+execution: code
+product_contract_source: ce-plan-bootstrap
+created: 2026-08-18
+---
+
+# fix: Reject colliding ask ids in the permission broker
+
+## Summary
+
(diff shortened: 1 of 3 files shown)


milind-soni:
Appreciate it! thanks

milind-soni (approved):
Reviewed the collision guard and resolved follow-up coverage. Full suite (984 app tests, broker tests, updater tests, and packaged-server smoke test) plus typecheck pass locally.

PR #249, opened 2026-08-19
chore: measure the palette against WCAG AA

koeseo:
`pnpm check:contrast` parses `src/styles.css` and measures 21 text/surface pairs. It reads the stylesheet rather than keeping a second copy of the values, so it cannot pass against a palette that is no longer the shipped one.

Two things it does that reading the hex values does not:

- **It composites alpha.** `--color-ink-secondary` is `#fcfcfc99`. Measured as opaque it looks like 15:1 on `--color-app`; composited, it is 2.86:1 — which is the number the user actually sees.
- **It measures white on filled surfaces**, because that is what the components render (`bg-accent … text-white`), not the token against the page ground.

### What it finds today

| pair | ratio | needs | where |
|---|---|---|---|
| white on `--color-accent` | 3.65:1 | 4.5:1 | every primary button, 12–13px — 28 sites |
| white on `--color-danger` | 3.10:1 | 4.5:1 | the hang-up buttons (`CallView.tsx:532`, `GroupCallView.tsx:492`), 14px |
| `--color-accent` on `--color-card` | 4.15:1 | 4.5:1 | accent links inside a `Card` — `ApiKeys.tsx:121`, `EnginesSettings.tsx:241` |

All three are listed in `KNOWN`, so **this lands without changing a single colour** and exits 0 as shipped. It fails only on something new.

### Why I did not just fix the colours

The shape of the problem rules out the obvious fix. `--color-accent` is fine as text on the page ground (5.33:1 on `--color-app`) and only falls short on the lighter card — while white falls short *on* the accent. Darkening the token repairs the buttons and hurts the links. A real fix wants a separate fill colour, and picking it is a design call on a palette you sampled deliberately. So this only measures; deleting a line from `KNOWN` is how a fix gets locked in.

### Verification

- as shipped: exit 0, `✓ 21 pairs measured, no new failures`
- drop `--color-ink-secondary` to `#fcfcfc55`: exit 1, naming all five surfaces
- `pnpm typecheck` clean, `oxlint` clean on the new file

## Summary by CodeRabbit

* **Accessibility**
  * Added automated WCAG 2.1 AA contrast checks for text, surfaces, buttons, links, borders, and focus indicators.
  * Contrast validation supports transparent colors and reports existing, undefined, unmeasurable, or newly failing color combinations.

* **Chores**
  * Added a command for running contrast validation during project checks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "scripts": {
     "clean": "node scripts/clean.mjs",
     "lint": "oxlint .",
+    "check:contrast": "node scripts/check-contrast.mjs",
     "dev": "vite",
     "companion": "node --experimental-strip-types companion/src/index.ts",
     "dev:server": "node --experimental-strip-types server/index.ts",
--- a/scripts/check-contrast.mjs
+++ b/scripts/check-contrast.mjs
@@ -1,0 +1,164 @@
+#!/usr/bin/env node
+// Measures the palette in src/styles.css against WCAG 2.1 AA.
+//
+//   node scripts/check-contrast.mjs        (or: pnpm check:contrast)
+//
+// It parses the stylesheet instead of keeping a second copy of the values, so
+// the check can never pass against a palette that is no longer the shipped
+// one. Two things it does that a quick eyeball does not:
+//
+//   - composites alpha. --color-ink-secondary is #fcfcfc99, and measuring it
+//     as opaque #fcfcfc overstates every secondary-text pair in the app.
+//   - measures white on filled surfaces, which is what the components
+//     actually render (`bg-accent … text-white`), not the token against the
+//     page ground.
+//
+// The two pairs already below AA are listed in KNOWN below, so adopting this
+// check does not force a palette change in the same commit. Anything new
+// fails the run.
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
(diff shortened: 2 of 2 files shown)


koeseo:
All three addressed — the second and third were real holes in the check itself.

**Comments were being parsed as declarations.** A token removed from the palette but left in a comment reads exactly like a live one, so it would be measured at its stale value instead of reported as undefined. `parseTokens` strips comments first.

**`KNOWN` was a licence, not a floor.** It matched pair names only, so white-on-accent could have slid from 3.65:1 to 2:1 and the run stayed green — the exact hole this check exists to close. It is now a `Map` of the ratio each pair measures today; a carried pair that gets worse fails like anything else. 0.01 of headroom, since the floors are quoted to two decimals.

I also made the inverse visible: a known pair that climbs past AA is announced, so the line gets deleted rather than sitting there shielding a pair that no longer needs it.

And the header said two pairs; there are three.

Verified all three ways:

- `--color-accent` → `#0a5db3`: `✗ … 2.33:1 … — WORSE than the recorded 4.15:1`, exit 1
- `--color-accent` → `#0d5aa8`: `✓ #ffffff on --color-accent now measures 6.90:1 — remove it from KNOWN`
- `--color-warning` commented out: `✗ undefined token in pair` instead of quietly measuring the dead value

Unchanged palette still exits 0.

PR #250, opened 2026-08-19
feat: let people turn usage analytics off

koeseo:
Analytics stay **on by default** — the counts keep working — but there is now a switch in Settings → General, and the choice survives restarts.

### Where the guard sits

Before `init()`, not just around `capture()`. An install that opted out never calls `posthog.init()`, so nothing leaves the machine, not even the library's own bootstrap request. Flipping it while the app is running takes effect immediately through `opt_out_capturing()`, which also drops what is already queued.

`identifyEmail()` carries a second, explicit check, because it is the one call that would send a personal identifier. The address is still stored locally in the profile either way — opting out stops it being *reported*, not being used.

### Why bother, when autocapture is already off

The restraint in `analytics.ts` is real and deliberate — `autocapture: false` with the reason written down, `capture_pageview: false`, `identified_only`, a short named-event list. The problem is that the people who worry about this are exactly the people who will not read the source to find that out. Naming what is sent, next to a switch, is what makes existing restraint legible.

`docs/ios-privacy.md` already promises this posture for the companion app ("no advertising, analytics, tracking, or third-party SDKs"). This lines the desktop side up with it.

### Notes

- `optAction()` carries the decision as a plain function, so all four cases are checked without standing up a client to observe — no module mocking, per the anti-slop rule.
- Unusable storage is treated as a fresh install rather than a failure, so a locked-down renderer does not take the toggle down with it.
- 9 tests, including the one that matters: opting out *before* init must not reach PostHog to tell it so.

`pnpm typecheck` clean, `oxlint` clean on the changed files, full suite green.

## Summary by CodeRabbit

* **New Features**
  * Added an Analytics setting in General settings.
  * Users can opt in or out of anonymous analytics collection.
  * Analytics preferences are saved and applied when the app starts.
  * Analytics tracking is disabled when opted out, including event and email activity.
  * Renamed the Keys section to Connections.
  * Added support for managed connected apps, Box, VPS, OpenCode Go, and self-hosted connected-apps keys.


koeseo:
Both right, and the first one was the feature failing at its own job.

**An opt-out was only as durable as the write that persisted it.** With storage rejecting the write, the setter swallowed the error, the next `analyticsEnabled()` found nothing and answered "enabled", and a later `initAnalytics()` would start the client the user had just switched off. The choice now lives in module state, set *before* the write is attempted — storage is how it survives a restart, not where it lives.

**The cold-start test proved nothing.** An earlier test had already set the module-scoped `ready` flag, so `initAnalytics()` returned on that rather than on the opt-out. It now loads the module fresh. `resetModules` is not module mocking — nothing is replaced, the real module is simply loaded again, which is what makes the guard the only thing that can stop it.

Verified by deleting the `!analyticsEnabled()` guard from `initAnalytics`: the test fails. It did not before.

Added one more while I was in there: an opt-out that cannot be persisted still holds for the session.

10 tests, full suite green.

PR #251, opened 2026-08-19
feat: add a Multica driver — a workspace's agents as contacts

koeseo:
[Multica](https://github.com/multica-ai/multica) assigns issues to coding agents and runs them on its own server. This driver puts that workspace behind the same chat window as every other engine: the model picker lists the workspace's **agents** instead of models, sending a message opens a ticket assigned to the selected one, and the run streams back into the bubble as `content.delta` until it settles.

### The one that spawns nothing

Every other driver here starts a CLI on this machine. This one starts nothing — the work happens on the Multica server. Two consequences that are the actual point:

- **A turn survives closing the laptop.** Reopen the app and the run is still going, because it never depended on this process.
- **It is the same ticket everywhere.** Opened here, it is visible in the Multica web UI; opened there, a reply lands in this chat. Follow-up turns comment on the running ticket rather than opening a new one.

### Setup is a CLI that is already signed in

Credentials are read from the config the `multica` CLI wrote, following its own layout — no profile → `~/.multica/config.json`, a named profile → `~/.multica/profiles/<name>/config.json`. That file carries the workspace id too, so a single-workspace user configures nothing at all. `MULTICA_SERVER_URL` + `MULTICA_TOKEN` override it for a server the CLI has never seen. Nothing is stored on our side.

Unavailability reads as an instruction rather than a missing file: an engine list that says *"Multica CLI is not signed in — run `multica login`"* is the whole setup guide.

### Shape

| file | job |
|---|---|
| `server/integrations/multica-client.ts` | the only place that speaks HTTP; every response shape parsed once |
| `server/drivers/multica.ts` | mapping only — text → ticket fields, run status → outcome, run message → canonical event |

`interruptTurn` also cancels the run on the server, not just the local subscription.

### Verification

21 tests. The credential lookup is covered hardest because its contract lives in another repo — default config, named profile, half-written config, absent config, env override, half-set env. `pnpm typecheck` clean, full suite green (994 passing).

Happy to adjust the naming or drop it out of the default fleet if you would rather it be opt-in via an `instances` entry.

## Summary by CodeRabbit

* **New Features**
  * Added Multica as an available agent provider.
  * Workspace agents can now be discovered and selected as models.
  * Added support for creating and commenting on issues, monitoring task progress, receiving runtime updates, and cancelling tasks.
  * Added credential discovery through environment settings or Multica CLI configuration.
* **Tests**
  * Added coverage for configuration handling, task status mapping, message conversion, cursor handling, and issue field generation.

### DIFF
--- a/server/drivers/builtIn.ts
+++ b/server/drivers/builtIn.ts
@@ -17,6 +17,7 @@ import { HermesAgentDriver } from "./acp/hermes.ts";
 import { OpenAICompatDriver } from "./openai-compat.ts";
 import { PiDriver } from "./pi.ts";
 import { MinimaxDriver } from "./minimax.ts";
+import { MulticaAgentDriver } from "./multica.ts";
 
 export const BUILT_IN_DRIVERS: readonly AnyProviderDriver[] = [
   GrokDriver,
@@ -35,4 +36,5 @@ export const BUILT_IN_DRIVERS: readonly AnyProviderDriver[] = [
   AntigravityDriver,
   BoxAgentDriver,
   MinimaxDriver,
+  MulticaAgentDriver,
 ];
--- a/server/drivers/multica.test.ts
+++ b/server/drivers/multica.test.ts
@@ -33,6 +33,20 @@ describe("issueTitleFromText / issueDescriptionFromText", () => {
   it("falls back to a default title for empty input", () => {
     expect(issueTitleFromText("   ")).toBe("OpenMausBot task");
   });
+
+  it("keeps the whole request when one long line has to be truncated", () => {
+    // Regression: the title caps at 200 and the brief is "the lines after the
+    // first", so a single long line used to lose everything past character
+    // 200 — the agent received a cut-off sentence and nothing else.
+    const long = `Baue die Startseite um: ${"x".repeat(250)} und dann noch der Rest`;
+    expect(issueTitleFromText(long)).toHaveLength(200);
+    expect(issueDescriptionFromText(long)).toBe(long);
+    expect(issueDescriptionFromText(long)).toContain("und dann noch der Rest");
+  });
+
+  it("still puts only the tail in the brief when the title fits", () => {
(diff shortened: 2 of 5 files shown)


koeseo:
All three were right, and the middle one was losing user text.

**A one-line prompt over 200 characters lost its tail.** The title truncates at the cap and the brief is "the lines after the first", which is empty for a single line — so the agent received a cut-off sentence and nothing else. When the title has to be cut, the whole text now becomes the brief. Regression test included; it fails without the fix.

**`fetch` had no ceiling.** This is the one I am glad you caught: the poll loop only reaches its own 30-minute guard, and only observes an interrupt, *between* requests. A socket that never answered would strand the turn forever, unkillable from the UI. Every request is now bounded at 30s via `AbortSignal.timeout` — `engines` already requires Node ≥24, so it is available.

**`pollMs` accepted 0 and negatives**, which would spin the loop with no pause at two requests per iteration, against someone else's server. Clamped to 500ms with a finite check.

23 tests. Both behavioural fixes verified by removing them — each takes its own test down.

milind-soni:
Super interesting, Are you also planning a kanban view with this

koeseo:
Yes — and the split is the point.

The kanban board is not a second place to work; it is the factory floor seen from the office. A card is a Multica ticket, so the board is a view onto work that is already running on a server, not a local todo list that pretends to be one.

**The front half is for deciding and delegating.** A room is a team, a ticket gets assigned in chat, someone asks a question, you answer it, the run reports back into the bubble. That is the conversation layer this app is already good at.

**The back half is for the work itself.** It runs on the Multica server, so it survives closing the laptop, and it is the same ticket whether it was opened here, in the Multica web UI, or from a phone.

Concretely, the board would let you see every ticket in the workspace by state, open one into the chat it belongs to, and trigger or interrupt a run from the card — without becoming a second source of truth. Multica stays the system of record; this stays the place where people talk about it.

The driver in this PR is the piece underneath that. Happy to sketch the board separately once this lands, so the two can be reviewed on their own merits.

PR #252, opened 2026-08-19
feat: four switchable skins, and the token that makes them possible

koeseo:
Midnight (the current palette, **unchanged**) plus three alternatives, switchable in Settings → General and remembered per machine. They fill a grid:

|  | cool | warm |
|---|---|---|
| **light** | Lagoon — ceramic, teal | Atelier — paper, daylight |
| **dark** | **Midnight** — the original | Foundry — night shift, brass |

### Why this is mostly a CSS change

Because of a property this codebase already has: **916 token classes against 8 hardcoded hex values and no Tailwind greys.** A skin is therefore one block of custom properties — Lagoon cost 51 lines. That token discipline is what made this worth trying at all.

The selector is `[data-skin=…]` rather than `:root[data-skin=…]`, so any element can open a skin context for its subtree. That is how the picker renders a live miniature of each skin instead of a painted mockup.

### One token does real work beyond theming

`--color-accent-ink` — the ink carried by a filled accent or danger surface. Components hardcode `text-white` on those fills at ~20 sites, which holds only while every accent is dark enough to carry white.

That constraint is measurable, and the numbers are worth having on their own: white on the current accent is **3.65:1** and on danger **3.10:1**, both short of the 4.5:1 that AA asks of the 12–14px labels they carry. (Measured in more detail in #249, which stands alone.)

Midnight sets the token to `#ffffff`, so **its rendering is byte-identical**. But Foundry can invert the relationship — brass fill, near-black lettering, 7.4:1 — which is what lets it read as lit metal rather than a dark paint chip. With hardcoded white, every accent would have to stay dark and the skin would lose its point.

Same story for the disabled state: `disabled:opacity-40` works on a dark ground and inverts on a light one, where Atelier's accent at 40% over white lands on `#d9bfa8` and white lettering measures 1.75:1. Disabled buttons get their own colours instead of transparency.

### Contrast is measured, not asserted

`scripts/check-skin-contrast.mjs` parses the stylesheet, reads `@theme` and `:root` as the **inherited base** — so a token a skin does *not* set is still measured as it actually ships — and fails on an undefined token rather than skipping it. 27 pairs per skin, all three new ones clear. Midnight is advisory, since its job is to be the original, gaps included.

That inheritance detail is not theoretical: while rebasing this onto v0.1.24 the new `--color-focus` landed, no skin overrode it, and all three would have silently worn Midnight's blue — 2.27:1 on Lagoon. The check now catches that class of regression.

Fully understand if a four-way skin picker is not where you want this app to go — the `--color-accent-ink` and disabled-state parts stand on their own if you would rather take only those.

### DIFF
--- a/scripts/check-skin-contrast.mjs
+++ b/scripts/check-skin-contrast.mjs
@@ -1,0 +1,152 @@
+#!/usr/bin/env node
+// Reads the skin blocks out of src/styles.css and measures every text/surface
+// pair the components actually produce. Run it after touching a palette:
+//
+//   node scripts/check-skin-contrast.mjs
+//
+// It parses the CSS rather than taking a second copy of the values, so the
+// check can never pass against a palette that is no longer the shipped one.
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+const root = join(dirname(fileURLToPath(import.meta.url)), "..");
+const css = readFileSync(join(root, "src/styles.css"), "utf8");
+
+function declarations(body) {
+  const tokens = {};
+  for (const [, name, value] of body.matchAll(/(--[\w-]+)\s*:\s*([^;]+);/g)) {
+    tokens[name] = value.trim();
+  }
+  return tokens;
+}
+
+/** The tokens every skin starts from: the `@theme` defaults plus the bare
+ * `:root` block. A skin that does not redefine one of these still SHIPS it,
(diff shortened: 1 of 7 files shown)


milind-soni (approved):
Reviewed against current main, including the newly merged Ubuntu/Wayland work. The CSS-token architecture, pre-paint application, local persistence, live previews, and accessible selection semantics are sound. Verified all four skins visually, reload persistence, production build, contrast checks, typecheck, full test suite, and packaged-server smoke test. No blocking issues.

PR #269, opened 2026-08-19
Add one pinned message per thread, shown as a banner above the transcript

### LINKED ISSUE #268: Pin a message to the top of a thread (pinned banner)
## Summary

There is no way to pin a message in a thread today (grep for "pin" finds only bot-list pinning). But threads accumulate exactly the content you want to keep in reach: the constraint the bot must respect, the decision that was made, the file path that matters, the answer you keep scrolling back to.

This proposes **pin one message per thread**, shown as a compact banner above the transcript. One pin, not a list — matching iMessage/Messenger mental models where a single pinned item is the common case, and keeping the data model minimal.

## Proposed behavior

- Hover a message → overflow actions → "Pin" (or unpin). One pinned message per thread; pinning another replaces the first.
- The banner sits above the transcript: sender name, one-line truncated text, click to jump to the message, an X to unpin.
- `pinnedMessageId?: string` on the bot/group records via the existing PATCH allowlist; the banner resolves it against the transcript from local state. A pinned message that was edited to another branch or deleted simply makes the banner disappear — the id resolves to nothing.
- Works in 1:1 threads and rooms (rooms pin any member's message, attribution shown).
- No new endpoints beyond the PATCH field; no new dependencies.

## Prior art

My native macOS multi-agent client ships `PinnedHeaderBannerView` — sender-attributed, one-line, click-to-jump, X-to-unpin — in daily use. This ports the pattern to the web idiom here.

## Scope

- `server/store.ts` — `pinnedMessageId?: string` on BotRecord + GroupRecord
- `server/index.ts` — PATCH allowlist entries (bot + group)
- `src/` — pin action in the message hover menu, `PinnedBanner` component, store field + action
- Tests: PATCH round-trip + banner render resolution

Happy to trim (e.g. 1:1 threads only first) if rooms make it bigger than wanted.

willsigmon:
Closes #268.

## What

A pin button joins the message hover actions (next to copy/edit/reactions). The one pinned message renders as a compact banner between the header and the transcript: sender name, one-line truncated text, **click to jump** to the message, **X to unpin**. Works in 1:1 threads and rooms; rooms attribute the sender (`You` or the member bot's name).

One pin per thread — pinning another message replaces the pin, matching iMessage/Messenger mental models and keeping the data model to a single field.

## Design

- `pinnedMessageId?: string` on `BotRecord` and `GroupRecord`, through the existing PATCH allowlists (validated as a message-id shape; `null`/`""` clears so the key drops off the record)
- The id is stored verbatim and **resolved against the transcript at render** — a pin whose message was edited to another branch or deleted simply renders nothing. No cross-record integrity to maintain, no migration edge cases
- Clearing sends `""` from the client (the debounced PATCH's `JSON.stringify` drops `undefined`), which the server normalizes to unset — same convention the codebase uses for `cwd`
- Banner jump reuses the existing `focusMessage` action (opens a transcript window around the target and scrolls to it) — no new scroll machinery

## Screenshot

Banner live on a real thread (sender + one line, between header and transcript):

pinned banner

## Validation

- `pnpm typecheck` ✓
- `pnpm vitest run src` — 94/94 ✓ · `server/store.test.ts` — 47/47 ✓
- New API assertions: pin / round-trip / replace / reject bad shape (`"not an id!"`, `42`) / clear via null
- Manual in the real app: pin via hover button → banner appears with sender + text; the bubble's pin flips to Unpin; X and Unpin both clear it; jump scrolls to the message
- Same honest note as #267: the spawn-dependent slice of `pnpm test` currently times out on this machine **identically on clean `main`** (verified by stash-and-rerun); unit + client suites are green and CI is authoritative

Prior art: the same pattern (sender-attributed one-line banner, click-to-jump, X-to-unpin) ships in a native macOS multi-agent client I maintain.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to pin one message in group rooms and bot conversations.
  * Added banners showing pinned messages, including sender and message text.
  * Select a pinned-message banner to jump to the message, or remove the pin.
  * Pins can be cleared when no longer needed.

* **Bug Fixes**
  * Added validation to reject invalid pinned-message identifiers with a clear error response.

### DIFF
--- a/server/index.test.ts
+++ b/server/index.test.ts
@@ -281,6 +281,19 @@ describe("harness HTTP API", () => {
     expect((await api("GET", `/api/threads/${bot.threadId}/export?format=pdf`)).status).toBe(400);
     expect((await api("GET", "/api/threads/nope/export")).status).toBe(404);
 
+    // one pinned message per thread: pin, round-trip, replace, clear; the
+    // id is stored verbatim — resolution is the UI's job
+    const pin = await api("PATCH", `/api/bots/${bot.id}`, { pinnedMessageId: "msg-abc_123" });
+    expect(pin.status).toBe(200);
+    expect(pin.body.bot).toMatchObject({ pinnedMessageId: "msg-abc_123" });
+    const repin = await api("PATCH", `/api/bots/${bot.id}`, { pinnedMessageId: "msg-second" });
+    expect(repin.body.bot).toMatchObject({ pinnedMessageId: "msg-second" });
+    expect((await api("PATCH", `/api/bots/${bot.id}`, { pinnedMessageId: "not an id!" })).status).toBe(400);
+    expect((await api("PATCH", `/api/bots/${bot.id}`, { pinnedMessageId: 42 })).status).toBe(400);
+    const unpinned = await api("PATCH", `/api/bots/${bot.id}`, { pinnedMessageId: null });
+    expect(unpinned.status).toBe(200);
+    expect(unpinned.body.bot).not.toHaveProperty("pinnedMessageId");
+
     // deleted conversations drop out of search rather than 404ing it
     await api("DELETE", `/api/bots/${bot.id}`);
     const after = await api("GET", "/api/search?q=nice%20to%20meet");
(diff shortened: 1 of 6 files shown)


willsigmon:
Addressed the room-pin coverage nit: `PATCH /api/groups/:id` now has the same pin / replace / invalid / empty-string-clear assertions as bots (`517bd7d`).

PR #271, opened 2026-08-19
Pin Grok ACP stdio to the picked local model

maxkongerskov:
## What changed

Grok ACP stdio now actually runs the model picked in the UI.

- Place `-m` and `--reasoning-effort` after `agent` and before `stdio`. Grok 1.0.6 only applies agent flags in that window; `grok -m slug agent stdio` is accepted as a TUI option and then ignored.
- After `session/new`, pin the slug with `session/set_model` (same path Hermes/Droid use). `-m` on argv is not enough: `session/new` still starts on `[models].default`.
- Resolve inject ids (`omlx::…`) through `resolveTurnModel` against the instance HOME, then pass that slug on argv. `spawnArgs` must not call `ensureGrokInjectSlug` itself — that helper defaults to `process.env` and would miss the instance override.
- After a successful pin, `session.started` reports the slug we set, not `initialize`'s `currentModelId` (`grok-4.6`).

## Why

Picking an oMLX / local model on a Grok bot still billed and ran `grok-4.6`. The picker wrote a `[model.slug]` block, but ACP never bound it.

## How it was verified

- `pnpm exec vitest run server/drivers/local-inject.test.ts server/drivers/acp/acp.test.ts` — argv order, inject slug, and `session/set_model`.
- `pnpm typecheck` and `pnpm test` pass locally.
- Live Grok ACP turn with an oMLX slug loaded that model instead of `grok-4.6`.

## Screenshots (UI changes)

None. Server-only; the existing model chip now shows the slug that actually ran.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Grok model selection so requested models are correctly resolved and applied.
  * Corrected command-line argument ordering for model, reasoning, permission, and connection settings.
  * Session startup now reports the requested model when no model is provided by the session.
  * Added clearer error reporting when Grok rejects a requested model.
* **Tests**
  * Added coverage for model selection, argument ordering, session configuration, and configuration persistence.

### DIFF
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -506,6 +506,25 @@ describe("ACP turns (fake CLI)", () => {
 
     expect(JSON.parse(readFileSync(without, "utf8")).argv).not.toContain("--reasoning-effort");
   });
+
+  it("puts Grok -m after agent so ACP stdio binds the local slug", async () => {
+    const dump = join(scratch, "grok-argv-order.json");
+    await create(GrokAgentDriver);
+    process.env.FAKE_ACP_DUMP = dump;
+    await instance.adapter.sendTurn({ threadId: "t-argv", text: "hi", model: "grok-4.5", effort: "high" });
+    await recorder.until((e) => e.type === "turn.completed");
+
+    const argv = JSON.parse(readFileSync(dump, "utf8")).argv as string[];
+    const agent = argv.indexOf("agent");
+    const modelFlag = argv.indexOf("-m");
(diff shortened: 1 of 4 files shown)


milind-soni (approved):
Reviewed against current main. The Grok ACP argv ordering, resolved local-model slug, and session/set_model pinning address the silent fallback to the default model. Cross-platform CI is green; local typecheck, the full test suite, and packaged-server smoke test pass. The cancelled packaging job was an unrelated Ubuntu apt-mirror timeout before project build code ran.

PR #274, opened 2026-08-20
Keep a room's approval answerable after its turn ends

josephleee:
## What happens

A room approval card that outlives its turn locks the room permanently. Every button on the card fails, and because a pending approval takes over the composer, there is no way to type past it either. Restarting the app does not help — the card is durable, so it comes back.

I hit this in a room with a real pending `Bash` approval: after quitting and reopening the app, Allow / Always allow / Deny / Cancel turn all did nothing, and the room could not be used again.

```
POST /api/threads/<room-thread>/respond {"requestId":"…","behavior":"allow"}
→ 404 {"error":"nothing is waiting on an answer in this conversation"}
```

## Why

A room resolves the bot behind an approval through `group.busyBotId`. That field is deliberately in-memory only — `saveGroups` strips it on write and load sets it to `null`:

```ts
writeFileAtomic(GROUPS_FILE, JSON.stringify(this.groups.map(({ busyBotId, ...g }) => g), null, 2));
```

The card, though, is persisted. So any approval that outlives its turn — a finished turn, or a restart — leaves a durable card whose speaker is gone, and `/api/threads/:id/respond` 404s on the `!owner` gate.

A 1:1 does not have this problem: it resolves through `store.botByThread()`, a structural lookup that always succeeds. That asymmetry is the bug.

Two things compound it:

1. The `!owner` gate sits *above* the peer-approval intercept, so peer cards — which belong to the bus and need no speaker at all — are unreachable in a room too.
2. `answerRequest` already knows how to close an approval it cannot deliver (marks the card `unavailable` + dismissed and says so in the transcript), but it finds the card only through `askMessageByRequest`, which is memory-only. After a restart it appends the notice **without dismissing the card**, so the composer stays blocked even once the route is reachable.

## The change

- Resolve the peer-approval intercept before the owner lookup.
- When `busyBotId` is gone, fall back to the member named by the card's `from`.
- When no owner resolves at all, still call `answerRequest` so the card is closed out rather than stranding the room. A thread with nothing pending still gets the 404.
- `answerRequest` falls back to finding the card by the request it carries when the in-flight map has been lost.

Server-only; no UI changes, no new dependencies.

## Testing

New API test seeds a room whose transcript holds a pending approval with no `busyBotId` — the restart case — and asserts the answer lands, the card is dismissed, and a room with nothing pending still 404s. It fails on `main` with `expected 404 to be 200`.

- `pnpm typecheck` clean
- `pnpm test`: 1073 passed. `server/drivers/codex-catalog.test.ts` fails identically on `main` on this machine (it reads the locally installed `codex` model catalog) — unrelated to this change.
- `oxlint server/index.ts` reports the same rule counts before and after.

## Follow-up (not in this PR)

`POST /api/groups/:id/interrupt` has the same root cause in its own way — with `busyBotId` null, `instance` is `undefined`, so `instance?.adapter.interruptTurn(...)` is a no-op while the route still returns `200 {ok:true}`. That is why "Cancel turn" fails *silently* where the others at least surface an error. Kept out to keep this PR to one concern; happy to send it separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Approval requests can now be recovered after a restart or temporary connection loss.
  * Stale or unreachable approvals are dismissed and shown as unavailable instead of blocking message composition.
  * Improved handling when responding to rooms without an active pending request.

### DIFF
--- a/server/index.test.ts
+++ b/server/index.test.ts
@@ -75,6 +75,16 @@ beforeAll(async () => {
         dm: true,
       },
       {
+        id: "test-stranded-room",
+        threadId: "test-stranded-room-thread",
+        name: "Stranded room",
+        memberIds: ["test-bot-a"],
+        defaultResponder: { kind: "member", botId: "test-bot-a" },
+        bulletin: "",
+        unread: false,
+        createdAt: 3,
+      },
+      {
         id: "test-pinned-room",
         threadId: "test-pinned-room-thread",
         name: "Pinned room",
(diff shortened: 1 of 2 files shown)


milind-soni:
Thanks for reporting, Looks good

Issue #289, opened 2026-08-20
Room turns are capped at a flat 5 minutes, even while streaming and even while parked on an approval


josephleee:
## What I see

Room turns keep dying at the five-minute mark while the bot is doing real work:

```
QA봇's room turn exceeded 5 minutes and was stopped
error: claude exited 143 before result
```

The second line is the first line's consequence, not a separate fault: the cap calls `interruptTurn` → `killCliTree`, the CLI takes SIGTERM, and 143 is 128+15.

## Where it comes from

`server/index.ts:1704` — a flat wall-clock ceiling on every room turn, with no env var and no mention in the docs:

```ts
}, 5 * 60_000);
```

## Why this looks like an inconsistency rather than a policy

`server/turn-watchdog.ts` states the project's stance at the top of the file:

> This watchdog watches ACTIVITY, not duration: a turn may legitimately run for an hour while events keep flowing, but a turn whose thread has emitted nothing at all for `stallMs` is wedged. Turns parked on a human approval are exempt — waiting on a person is not a stall.

The room ceiling bypasses all three of those properties:

| | stall watchdog | room ceiling |
|---|---|---|
| basis | only when the thread emits **nothing** | wall clock, regardless of activity |
| default | 20 min (`OMB_TURN_STALL_MS`, floor 60s) | 5 min, hard-coded |
| parked on a human approval | exempt (`waitingOnHuman`) | **not exempt** |
| applies to | 1:1 and rooms | rooms only |

So the same piece of work is fine for an hour in a 1:1 and killed at five minutes in a room, and `OMB_TURN_STALL_MS` cannot raise the room ceiling because it governs the other mechanism.

## The approval case is the sharper half

The room timer is an unconditional `setTimeout` — nothing checks whether the turn is parked waiting on a person. Take longer than five minutes to answer an approval in a room and the turn is killed underneath you. `turn.completed` then clears `busyBotId`, but the card itself is durable and stays on the thread, which is exactly the stranded-approval state fixed in #274. #274 repairs the recovery path; this ceiling is one way the state gets created in the first place.

(The room does eventually self-heal — the stall watchdog's grace release clears `busyBotId` — but that is up to 20 minutes later, and it does not bring the killed work back.)

## Direction I would suggest

[...]

PR #290, opened 2026-08-20
Close the approvals a cancelled turn can no longer answer

josephleee:
## What happens

Press **Cancel turn** on an approval card and the card stays. The composer stays blocked behind it, and the button that was supposed to be the way out has quietly made things worse.

## Why

`Cancel turn` lives on the approval card itself, and `Composer.tsx` is explicit about what a pending approval does:

```
{/* An approval takes over the composer: you answer it before you
    can type again, so a waiting bot is impossible to miss. */}
```

The interrupt routes stop the turn but never touch the card the turn had raised. `pendingApprovals` decides what is open from the card alone:

```ts
.filter((m) => m.kind === "options" && m.card?.requestId && m.card.tool && !m.card.answered && !m.card.dismissed)
```

Nothing there knows whether a turn is running, so the card survives the interrupt. The process that asked the question is dead, so it can never be answered — and since the interrupt also clears `busyBotId`, the room ends up in exactly the stranded state where the card's own buttons have no speaker to resolve against. Pressing the escape hatch is what springs the trap.

The only place that closes an approval today is `answerRequest`'s `unavailable` branch. Interrupts had no equivalent.

## The change

A small `closeOpenApprovals(threadId)` that marks every still-open approval on a thread `answered: "unavailable"` + `dismissed`, the same way `answerRequest` marks one it cannot deliver, and drops its `askMessageByRequest` entry. Called from both interrupt routes — including the room thread a bot is busy on when it is stopped from its own 1:1 chat, which the route already reaches for the interrupt itself.

Server-only; no UI changes, no new dependencies.

## Testing

New API test seeds a room holding an unanswered approval, calls `POST /api/groups/:id/interrupt`, and asserts the card is closed. It fails on `main` with `expected undefined to be true`.

- `pnpm typecheck` clean
- `pnpm test`: 1073 passed. `server/drivers/codex-catalog.test.ts` fails identically on `main` on this machine (it reads the locally installed `codex` model catalog) — unrelated.
- `pnpm test:packaged-server` passes; `oxlint server/index.ts` reports the same rule counts before and after.

## Related

- #274 fixes the other half: answering a card whose speaker is already gone. The two are independent — this one stops the state being created by a button press, that one lets you out of it once you are in. Neither depends on the other.
- #289 covers a third way in, the flat 5-minute room ceiling, which is a behaviour question rather than a fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Interrupting a room or direct bot conversation now automatically closes unresolved approval requests.
  * Open approvals are marked as unavailable and dismissed when an active turn is interrupted.

### DIFF
--- a/server/index.test.ts
+++ b/server/index.test.ts
@@ -75,6 +75,16 @@ beforeAll(async () => {
         dm: true,
       },
       {
+        id: "test-cancel-room",
+        threadId: "test-cancel-room-thread",
+        name: "Cancel room",
+        memberIds: ["test-bot-a"],
+        defaultResponder: { kind: "member", botId: "test-bot-a" },
+        bulletin: "",
+        unread: false,
+        createdAt: 4,
+      },
+      {
         id: "test-pinned-room",
         threadId: "test-pinned-room-thread",
         name: "Pinned room",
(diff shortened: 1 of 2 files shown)


milind-soni:
Good catch, thanks. Really appreciate it!

milind-soni:
Superseded by #292, which includes the corrected implementation and has been merged into main.

PR #311, opened 2026-08-20
feat: add internationalization support

qiyuechen0929:
## What changed

Added an open-source localization tool for OpenMausBot with support for multiple languages, including:

- Simplified Chinese
- Traditional Chinese
- Japanese
- Korean
- Spanish
- French
- German
- Russian

The tool provides a way to localize the OpenMausBot UI while keeping the original application logic unchanged.

## Why

OpenMausBot currently does not have built-in multi-language UI support. I created this as a possible approach for adding localization support.

This is my first contribution to OpenMausBot, so I would really appreciate any feedback or suggestions.

## How it was verified

I tested the localization tool on my Windows environment and verified that the UI can be localized and restored.

There may still be bugs or compatibility issues, especially with different OpenMausBot versions. I would be happy to make any necessary changes based on your feedback.

## Summary by CodeRabbit

* **New Features**
  * Added multilingual interface localization for Chinese, Traditional Chinese, German, Spanish, French, Japanese, Korean, and Russian.
  * Added a Windows launcher for applying, listing, and restoring language translations.
  * Added safe translation updates with backups, restoration, dry-run support, and optional syntax validation.

* **Documentation**
  * Replaced the README with bilingual setup, usage, troubleshooting, and localization guidance.

* **Validation**
  * Added tools to extract interface text, validate language files, inspect installations, and verify applied translations.


milind-soni:
Thank you for the contribution. I’m closing this version because it does not add runtime localization to the application sources: it patches installed compiled JavaScript, overwrites README content, and includes generated Python bytecode. A focused replacement would localize the actual desktop source and the native iOS target (for example with source catalogs/app resources), with no post-install code rewriting or generated artifacts committed.

PR #328, opened 2026-08-21
feat(profile): add persistent desktop profiles and avatar roster

willsigmon:
> **Merged after #327.** The final branch contains the maintainer's StrictMode-safe profile patch queue integration and preserves the local-computer Auto acknowledgement boundary.

## What changed

- make the selected agent's header avatar and name open the persistent right-hand profile rail
- put identity controls first in that rail: custom/generated avatar, crop shape, name, title, and description
- add persistent Comfortable (320 px), Compact (272 px), and Avatars only (80 px) roster densities, plus a direct collapse/expand action
- render uploaded/generated agent avatars and overlapping room stacks throughout the roster while retaining the mascot fallback
- keep icon-only rows keyboard accessible with labels, tooltips, focus states, and effective targets
- move agent voice selection, preview, and shared ElevenLabs-key configuration into that agent's profile; App Settings no longer owns a competing voice-selection flow
- serialize optimistic profile writes so rapid edits cannot apply stale responses out of order
- preserve the merged local-computer Auto warning boundary while ensuring request-only acknowledgement data never leaks into optimistic bot state

## Why

Agent settings should be a first-class, persistent workspace rather than a context-menu-only destination. A compact avatar rail also makes custom identities useful when users want more conversation space, matching the supplied Grok-style reference without replacing OpenMausBot's existing sidebar architecture.

The already-merged compact iOS header from #248 is preserved and supplies the native name/avatar entry point; this PR does not reimplement that broader responsive-header work.

## How it was verified

- `pnpm typecheck`
- focused Vitest coverage for the bot patch queue, title activation, rename behavior, sidebar preferences, and store folding
- `pnpm build`
- cumulative native parity is supplied by #327 and its **115-test** Swift suite
- unsigned generic iOS Simulator build — **BUILD SUCCEEDED**

### Persistent profile rail

Desktop agent profile rail

### Avatar-only roster

Desktop avatar-only roster

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — typecheck and this slice's tests pass; the full floor hits the existing upstream team-import timeout documented in #327
- [x] Server behavior changes come with tests (see CONTRIBUTING.md → Tests)
- [x] No `dist-server/` edits (it's build output)
- [x] macOS-only code is platform-gated; no `shell: true` / cmd.exe string-building
- [x] No secrets in logs, responses, events, or argv

PR #327, opened 2026-08-21
feat(profile): add paired-safe agent profiles and avatars

willsigmon:
## What changed

- add one validated agent-profile boundary for name, title, description, notifications, avatar, crop, ElevenLabs voice, and spoken replies
- persist custom PNG/JPEG/GIF/WebP avatars only through app-owned attachment URLs; external URLs and SVGs stay rejected
- add image upload plus optional GPT Image 2 generation with a bounded product prompt and a write-only, OS-protected OpenAI key
- add native iOS profile parity: open an agent from the compact chat header, edit paired-safe fields, upload/generate an avatar, choose/preview a per-agent voice, and render custom avatars throughout the native roster/chat UI
- keep provider keys off paired clients and strip provider session cursors from the wire contract
- prevent stale avatar-generation responses from overwriting a newer upload/removal, validate paired avatar paths strictly, and tolerate future crop values on older iOS clients
- document the intentionally non-destructive avatar-retention behavior and the reference-aware cleanup follow-up

## Collaboration/rebase note

Merged after maintainer review. The final branch includes the maintainer's current-`main` integration plus paired-write coverage, existence checks, and the accepted no-key guidance that sends users to Settings on the computer without exposing credentials to iOS.

## Why

Agent identity and voice were scattered across broad settings surfaces and were not available from the native companion. This establishes a single shared contract first, so the desktop profile rail can be layered on without introducing a web-only data path or widening the paired-device capability boundary.

## How it was verified

- `pnpm typecheck`
- focused Vitest coverage for companion routes, avatar/profile validation and generation, server API/wire behavior, TTS, workspace credential storage, and packaging
- `pnpm check:electron`
- `pnpm test:packaged-server`
- `xcrun swift test` — **115 tests, 0 failures**
- unsigned generic iOS Simulator build — **BUILD SUCCEEDED**
- the two existing team-import cases that exceed the local 20-second floor pass together with `--testTimeout=120000`; that default-timeout cascade is unchanged from upstream

### Native agent profile

Native iOS agent profile

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — typecheck passes; the full default floor hits the existing upstream team-import timeout described above
- [x] Server behavior changes come with tests (see CONTRIBUTING.md → Tests)
- [x] No `dist-server/` edits (it's build output)
- [x] macOS-only code is platform-gated; no `shell: true` / cmd.exe string-building
- [x] No secrets in logs, responses, events, or argv

PR #329, opened 2026-08-21
feat(routines): clarify scheduled tasks and exact notifications

willsigmon:
> **Merged after #328 and #327.** The final maintainer integration routes room notifications to rooms and pins completion deduplication with a real partial-stream regression test.

## What changed

- rename and reorganize the existing automation surface as **Tasks & routines**, with plain-language definitions for Task, Routine, and Webhook instead of presenting cron syntax as the user model
- keep one-time and repeating schedules on ordinary date/time/day controls and make desktop-vs-cloud execution explicit
- use a fresh task for every routine or webhook run while preserving visible receipts for queued, running, needs-you, completed, missed, failed, and cancelled outcomes
- display webhook deliveries in the same execution history without pretending an event trigger is a calendar schedule
- implement the previously inert per-agent notification switch with a small server-owned policy: approval, question, takeover, non-empty completion, and routine failure
- include both agent ID and exact task thread ID in every notification, then switch to that detached task when the notification is clicked/tapped
- add native iOS Tasks & Routines management, run receipts, notification permission state, and exact notification deep-link navigation
- keep unknown future schedule kinds visible but read-only on older iOS clients instead of writing them back incorrectly
- document what proactivity means: Auto mode, a configured Routine, a configured Webhook, or explicit peer coordination—not an invisible heartbeat that invents work

## Why

“Cron jobs” obscure the product model and make it hard to tell a conversation from a reusable schedule. Users should be able to see what starts work, which agent/context it inherits, where it runs, and what happened. Notification QA also needs a deterministic target; opening whichever task is currently active is incorrect for detached routine work.

The webhook admin/secret surface remains computer-only. iOS can inspect routine-backed receipts and manage normal schedules without receiving provider or webhook credentials. Closed-app iOS delivery still requires a future APNs relay and is documented rather than implied.

## How it was verified

- `pnpm typecheck`
- focused Vitest coverage for companion route policy, notification wiring, routine execution, notification targets, and store folding
- `pnpm build:companion`
- `xcrun swift test` — **120 tests, 0 failures**
- unsigned generic iOS Simulator build — **BUILD SUCCEEDED**
- manual QA matrix is recorded in `docs/notification-and-proactivity-qa.md`

### Tasks & routines calendar and receipts

Tasks and routines

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — typecheck and all focused tests pass; the full floor hits the existing upstream team-import timeout documented in #327


willsigmon:
Addressed the routines/notifications review pass in the rewritten routines commit:

- fixed-format `HH:mm` values now use `en_US_POSIX`, Gregorian calendar, and the current local time zone
- startup recovery persists interrupted runs as failed before invoking the same failure-notification callback
- unsupported companion routine actions now receive operation-specific wording instead of claiming all routine management is desktop-only
- the notification test helper safely handles an empty 204 response

Focused routine/notification/companion tests, 117 Swift package tests, typecheck, and unsigned simulator build pass locally. Fresh CI is running on `56e58b3`.

PR #333, opened 2026-08-21
Fix Codex web search activity completion

08820048:
## What changed

- Normalize Codex webSearch item/completed notifications as completed tool events.
- Extend the fake Codex app-server with a web search lifecycle.
- Assert that web search activity transitions from started to successful completion.

## Why

Codex web searches emitted an activity-start event, but their matching completion notification was ignored. The persisted activity therefore kept an undefined outcome, so the UI spinner never stopped after the turn finished.

## How it was verified

- pnpm exec vitest run server/drivers/codex.test.ts
- pnpm typecheck
- pnpm test
- pnpm check:electron

The full suite passed with 1381 tests passing and 12 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Web searches now correctly report when they start and finish.
  * Search activity is consistently recognized alongside other tool actions during responses.

### DIFF
--- a/server/drivers/codex.test.ts
+++ b/server/drivers/codex.test.ts
@@ -86,7 +86,9 @@ describe("CodexDriver turns (fake app-server)", () => {
       "turn.started",
       "session.started",
       "item.started", // commandExecution ls -la
+      "item.started", // webSearch OpenMausBot
       "item.completed", // commandExecution done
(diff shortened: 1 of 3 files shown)


milind-soni:
Looks good, getting it in

PR #341, opened 2026-08-21
fix(rooms): explain when archived members cannot respond

08820048:
## Summary

- show a persistent room activity error when a message explicitly targets an archived bot
- explain when a room has no active members available to respond
- preserve intentional silence for unmentioned messages in mentions-only rooms
- add API regression coverage for all three cases

## Testing

- pnpm typecheck
- pnpm exec vitest run server/index.test.ts (73 tests passed)

## Summary by CodeRabbit

- **New Features**
  - Added clear activity messages when a requested responder is archived or unavailable.
  - Notifies users when all room members are archived and no response can be generated.
  - Explains when the configured default responder is archived and cannot respond.

- **Bug Fixes**
  - Prevents archived members from being selected as direct-message fallback responders.
  - Prevents response attempts when no active responders are available.
  - Preserves normal message behavior when no archived responder is mentioned.
  - Supports mixed archived and active mentions without interrupting available responses.

### DIFF
--- a/server/index.test.ts
+++ b/server/index.test.ts
@@ -317,13 +317,42 @@ describe("harness HTTP API", () => {
     expect(body.bots[0].messages.length).toBeGreaterThanOrEqual(2);
   });
 
-  it("keeps direct-message channels folderless at the API boundary", async () => {
-    const attempted = await api("PATCH", "/api/groups/test-dm", { cwd: home });
-    expect(attempted.status).toBe(400);
-    expect(attempted.body.error).toMatch(/direct-message.*working folder/i);
-    const state = await api("GET", "/api/bots");
-    expect(state.body.groups.find((group: { id: string }) => group.id === "test-dm")).not.toHaveProperty("cwd");
-    expect((await api("DELETE", "/api/groups/test-dm")).status).toBe(200);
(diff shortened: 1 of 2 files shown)


milind-soni (approved):
Nice UX gap to close — silent sends into rooms where every member is archived are genuinely confusing. Reviewed the routing path end to end:

- `roomResponders` already filters hidden members before mention resolution (`server/store.ts:393`), so an archived bot can never be dispatched even when mentioned — the new `mentionedArchived` feedback message is purely additive.
- The DM fallback correctly falls back through `availableMembers` instead of raw `members`, and the all-archived case now gets the 'no active room members' activity card instead of silence.
- Mixed-mention case (archived + active) emits the archived notice *and* still dispatches to the active member — matches the test, and is the right behavior.
- Tests clean up after themselves in `finally` and use `expect.poll` for async delivery. Green CI across all three platforms.

One tiny nit, non-blocking: the archived-mention notice fires even when the same message also successfully dispatches active responders — that's intended per the test, but worth remembering if anyone later adds a setting to suppress activity cards.

milind-soni:
Superseded by #385, which carries this fix onto current main, resolves the room-management conflicts, and includes the missing archived-default and mixed-mention coverage. Thank you @xuyi-lab for the original implementation.

PR #344, opened 2026-08-21
Show clickable options when a bot asks you a question

AdityaUmale:
### What was wrong

The Claude CLI has its own tool for asking you to pick from a list. It came to
us through the permission tool, so we treated it as "may I run this?" and drew
an Allow/Deny box over a chunk of JSON. The buttons the bot wanted you to press
were never shown. In auto mode the question could be answered for you, so you
never saw it at all.

### What this does

Reads it as what it is: a question. You get a normal option card with real
buttons and a short line under each one saying what it means.

The answer goes back in the field the tool asks for. Just saying "allow" is not
enough — the CLI then runs the tool with no way to show you anything, and tells
the bot you did not answer. Checked end to end against claude 2.1.238.

### Also fixed, same area

- Auto mode can never answer a question for you. That was the point of asking.
- Questions asked inside a room now show up. They used to draw nothing while
  the bot sat waiting for 15 minutes.
- Closing a question now works. It used to do nothing at all.
- A card decides allow/deny from the card, not from the button's text. A
  question with an option called "Allow" used to be sent as an approval.
- A broken question is refused with a reason instead of falling back to the
  Allow/Deny box this PR removes.
- A "nobody answered in time" note is no longer saved as your answer.
- Permission cards never take their buttons from text the model wrote. On the
  phone those become the decision buttons, where they all mean deny while
  "always allow" quietly writes a real grant.

### Known gaps

- Multi-select does not highlight what you picked. The server overwrites the
  saved answer with the word "answer". This is older than this PR and affects
  single-select too.
- On a multi-select card, pressing Enter in the text box drops your ticks.
- A call with several questions is asked one card at a time, each with its own
  15 minute wait. One card with one wait may be better — worth a view.
- Room cards rely on speaker info that can be cleared, and a room turn is cut
  off at 5 minutes even while a card is open.
- The iPhone app does not know about multi-select or the per-option lines yet.

### Testing

typecheck, full suite (1535 pass), electron check and UI build all pass.
New tests cover the proxy, the driver and the answer rules. Also ran it by hand
in the desktop app.

## Summary by CodeRabbit

- **New Features**
  - Added interactive question cards with explanatory option hints and multi-select support.
  - Added support for answering multiple questions sequentially, including partial or unanswered responses.
  - Question cards now work in room threads and group conversations.
  - Added clearer answer, dismissal, and selection handling for question cards.

- **Bug Fixes**
  - Question requests are no longer automatically approved, including remembered permissions and alternate tool names.
  - Improved handling of invalid, unavailable, timed-out, or disconnected question requests.

### DIFF
--- a/server/auto-approve.test.ts
+++ b/server/auto-approve.test.ts
@@ -4,7 +4,7 @@
 // question is never answered by the machine.
 import { describe, expect, it } from "vitest";
 
-import { approvalKey, autoDecision, looksDestructive, looksSensitive } from "./auto-approve.ts";
+import { approvalKey, autoDecision, autoVerdict, looksDestructive, looksSensitive } from "./auto-approve.ts";
 
 describe("looksDestructive", () => {
   const dangerous = [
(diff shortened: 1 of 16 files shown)


alisson-acioli:
Any ETA for this fix to be released? I ran into this today, @milind-soni

aivsomkar:
@AdityaUmale — thank you for this, and sorry for the duplicate: I built #953 for the same bug without finding this PR first, and only spotted it while filling in the platform table.

Your fix is the one that ships. In particular you had the mechanism right and I did not: I was delivering the answer on the permission contract's **deny** channel, which reaches the model but arrives as `is_error: true`. You filled in the tool's own `answers` field. I tested it against claude 2.1.260 before switching over:

```
proxy → {"behavior":"allow","updatedInput":{…,"answers":{"Do you prefer tea or coffee?":"Tea"}}}
model ← The user answered: "Do you prefer tea or coffee?"="Tea". Read the answers carefully …
```

#953 now **merges this branch** rather than reimplementing it, so your commit and authorship are in the history. It carries your mechanism, the `auto-approve.ts` backstop, the room-question branch in `GroupView`, `card-answer.ts` and the dismiss fix, `source` on the broker wire so a timeout's note is never filed as a person's answer, and your `permission-proxy.test.ts` and driver tests.

Two things were reconciled, both because #953 adds a card that draws every question at once with a tab each:

- **One ask for the whole set** instead of one per question, so a person sees what they are committing to before answering any of it. `questionAnswersByQuestion` maps the single reply back per question, and falls back to the whole message when only one question was asked — which is how the phones' flat option buttons still answer.
- **`choiceHints`/`multiSelect` on the card are dropped**, and the flat multi-select picker with them. `questions[]` carries the same information in a richer form, and one representation beats two. Your tests for them were rewritten against the new shape rather than deleted.
- One behaviour change to yours: a question with **no options** is now kept rather than refused, because the new card always offers a free-text row, so a person can still answer it. "No answerable question" now means no question *text*.

Closing in favour of #953. The credit is yours — please do review it, especially the proxy.

Issue #313, opened 2026-08-20
iOS: switching tasks leaves the old transcript on screen, and sends to the new thread


KesleyDavid:
## Summary

Switching tasks while a chat is open leaves `ChatView` rendering the **previous** task's transcript. Because the composer targets the bot's *live* thread while the transcript is read from a *frozen* one, a message typed on that screen is delivered to a conversation the user is not looking at, and never appears.

The screen looks entirely normal throughout — the old transcript is still plausible, so there is nothing to suggest the two halves disagree.

## Steps to reproduce

1. Open a bot's chat.
2. Tap the name → **Tasks**, and pick a different task. (Or `+` → **New task**, which is a single tap.)
3. The sheet closes and the transcript does not change — it is still the previous task's.
4. Type a message and send it. It is delivered to the newly-selected task's thread, and does not appear on screen.

It also happens with no interaction on the phone at all: switch the task on the desktop, and an open `ChatView` stays on the old thread as the `.bot` frame lands.

## Cause

`ChatView` is handed a `Chat` value by navigation (`ios/App/ChatListView.swift:114`), which is a snapshot of the bot at push time. The file already accounts for that snapshot going stale — `current` re-reads the live record from the store:

```swift
// ios/App/ChatView.swift:44-51
/// The live chat record, so busy/unread stay current as frames land.
private var current: Chat { … session.state.bot(bot.id).map(Chat.bot) ?? chat … }
```

…but the transcript is read from the frozen one:

```swift
// ios/App/ChatView.swift:40-42
private var messages: [Message] {
    session.state.visibleTranscript(forThread: chat.threadId)
}
```

Switching a task changes `bot.threadId` (`server/store.ts:992`, via `ios/App/TaskManagerView.swift:23` → `ios/App/Session.swift:597-601`), so from that moment `chat.threadId` and `current.threadId` refer to different conversations.

Nothing rebuilds the view: it has no `.id(…)` keyed on the thread, none of the three `.task {}` blocks are keyed, and `Chat`'s `==`/`hash` use only `id` (`ios/App/Session.swift:703-720`) — so SwiftUI considers the old and new `Chat` the same navigation value.

The store keeps the old transcript rather than dropping it (`ios/Sources/CompanionCore/Store.swift:196-215` writes `messages[bot.threadId]` without clearing the previous thread's entry), which is why the stale screen looks healthy instead of empty.

## Why it is more than a display bug

[...]

Issue #312, opened 2026-08-20
iOS: "Always allow" on a peer-approval card sends deny


KesleyDavid:
## Summary

On iOS, a permission card's answer is derived from the **label of the button the user tapped**, and any label that isn't literally `"allow"` is sent to the harness as `behavior: "deny"`.

The peer-approval card offers three options — `["Allow", "Deny", "Always allow"]` — so tapping **"Always allow"** denies the request. The user asks for a broader permission and gets a refusal instead.

## Where

`ios/App/Session.swift:449-463`:

```swift
func answer(threadId: String, requestId: String, choice: String, isPermission: Bool) async {
    await perform {
        if isPermission {
            try await $0.respond(
                threadId: threadId,
                requestId: requestId,
                behavior: choice.lowercased() == "allow" ? "allow" : "deny"
            )
        } else {
            try await $0.respond(threadId: threadId, requestId: requestId, behavior: "answer", message: choice)
        }
    }
}
```

The card's option buttons pass the raw label into that function (`ios/App/ChatView.swift:841-847`), and the harness builds the peer-approval card with three options in `server/peer-approval.ts:83-93`:

```ts
options: ["Allow", "Deny", "Always allow"],
```

`"always allow".lowercased() != "allow"` → `behavior: "deny"` → `server/peer-approval.ts:154` reads `behavior === "allow"` as false and settles the card as a denial.

## Steps to reproduce

1. Give a bot `approvePeerComms = true`.
2. Have it call `ask_bot` (or `delegate_bot`) against another bot, so the harness pushes the peer-approval card into the source bot's thread.
3. On iPhone, open that thread and tap **"Always allow"** on the card.
4. The contact is **denied**. No standing grant is recorded either, since the card's own option button does not go through `alwaysAllow`.

## Expected

Tapping "Always allow" should allow the request and record the grant against the card's `allowKey`, which is what the desktop does.

## Why the existing safeguard doesn't catch it

`ChatView.allowChoice` (`ios/App/ChatView.swift:799-803`) is careful to answer with a label the card actually offered, and its comment explains why:

> Answering with a string the card never offered writes the grant and then hands the harness a choice it can reject, so the bot stays stopped with nothing on screen to explain it.

[...]

Issue #241, opened 2026-08-18
An Android companion app — proposing before any PR


KesleyDavid:
CONTRIBUTING says big changes should start as an issue. This is that, before
any Android code.

The iOS companion is already in-tree (`ios/` + `companion/`, via #146 / #160 /
#161 / #204). I'd like to add the same OpenMausMobile for Android: a native
Kotlin + Compose client in `android/`, next to `ios/`. Same product, same
sidecar — so people can use the companion on either phone.

I have not started a PR. I want agreement on putting it in this repo, and on
the split, first.

## What this is (and is not)

The phone is a thin remote client of the existing sidecar, not the desktop UI
on a smaller screen and not a hosted copy of the web app.

```
  phone ──LAN / Tailscale──▶ companion :8810 ──loopback──▶ harness :8799
                              token, allowlist, Origin refused
```

- The computer still owns agents, keys, transcripts, and computers.
- Pairing stays Settings → Companion (QR / 6-digit / Bonjour / manual /
  MagicDNS). The Android app only redeems `openmausbot://pair`.
- Same HTTP + SSE allowlist in `companion/src/routes.ts`.
- Same decode/fold contract as `ios/Sources/CompanionCore`, pinned by
  `ios/Tests/CompanionCoreTests/Fixtures`.

`server/` stays untouched. `companion/` stays untouched unless a route the
merged iOS app already uses is missing — I don't think any is.

## What I am actually asking

**Do you want `android/` in this repo?**
Native Kotlin + Compose, sitting next to `ios/`. A fine answer is no — it can
live in a fork.

A Play Store listing can wait. Sideload / Play Internal is enough for a first
merge. Same review note as iOS: a reviewer with no OpenMausBot install still
needs a path that opens (demo/preview).

## Proposed v1 (parity with merged `ios/` on main)

In: pair, roster, chat, approvals, tasks/search/versions/reactions/share,
computer preview + cloud join, local notifications while connected.

Out of v1: voice/call, keys, connectors, VM, killed-app push, and anything
not yet on `main` in `ios/`. Follow the merged iOS app; don't outrun it.

## How I would split it

1. `android/core` — models, SSE, fold, client, pairing, failover. Tests
   against the existing iOS fixtures. No Activity.
2. Session + pair + roster + chat + approvals.
3. Tasks / search / versions / reactions / share.
4. Computer preview + cloud join + settings + local notifications.
[...]

PR #349, opened 2026-08-21
android/core: the companion's shared core, ported to Kotlin

KesleyDavid:
Refs #241.

The first slice of the Android companion: `android/core`, a plain Kotlin JVM module with no Android dependency and no UI. It is the port of `CompanionCore` — wire models, the SSE parser, the state fold, the HTTP client, pairing-invite and connection parsing, failover, the markdown block splitter, the Session and its storage.

Nothing outside `android/` is touched, and the module adds no dependency on anything that already exists here.

## Why this is worth reviewing on its own

It is the layer both future PRs and the existing iOS app agree on. **Its tests read the iOS fixtures directly** — `ios/Tests/CompanionCoreTests/Fixtures/` is wired in as a resource directory rather than copied — so if a payload shape changes, the Swift tests and these fail together. That is the property that keeps the two ports from drifting on the wire, and it only works if the fixtures stay the single source of truth.

**195 tests, all green.** `cd android && ./gradlew :core:test`. It builds without the Android SDK.

## What the port mirrors deliberately

Behaviour follows the iOS shared core exactly, including a few things I would have written differently:

- `Connection.urlHost` keeps a scope zone on link-local IPv6 and drops an interface zone from anything else, as `8faabd7c` made it do.
- A permission card's answer is derived the way `OptionCard.responseBehavior` derives it — the one refusal is `Deny`, trimmed and case-insensitive, and every other offered choice means allow — so the fix in #334 is carried over rather than re-invented.
- A group name that trims to nothing is omitted so the harness names the room after its first member; a name of only newlines is sent literally, because `CharacterSet.whitespaces` does not contain them. That is iOS's behaviour, so it is this module's behaviour.
- `validAvatarPath` accepts `.jpg` and refuses `.jpeg` and uppercase, matching `Client.swift` and `shared/bot-avatar.ts`. Worth noting the companion allowlist's `jpe?g` regex is a step looser than both — a request can pass the proxy and then be refused by `readAttachment`. I have not touched that here; happy to open a separate issue if it is not already known.

Where I thought iOS had a bug I reported it rather than diverging — #312 and #313 came out of this work.

## How it was built

Written by an agent orchestra with a hard rule that the author never reviews their own code: one model implements, a different one reviews strictly against the Swift, and nothing is committed until the reviewer approves. That is why the commits are shaped the way they are, and why several of them exist only to close a review finding.

## Still to come

The `:app` module — platform layer, Compose UI, the screens — is built and passing (533 further tests) but is not in this PR. Per the plan in #241 it follows as separate PRs so each stays reviewable. **We are still working on it**: it has not yet run on a physical device, so device verification, screenshots and a CI job are outstanding. I would rather land the contract layer first and take feedback on it before the rest arrives.

Happy to reshape, squash, or split this differently — tell me what is easiest to review.

## Summary by CodeRabbit

* **New Features**
  * Added the Android companion foundation for pairing, secure connections, reconnection, and host failover.
  * Added support for chats, rooms, messages, tasks, search, notifications, profiles, avatars, voices, routines, exports, and cloud desktop sessions.
  * Added real-time event streaming and resilient state updates.
  * Added Markdown parsing for rich message display.
* **Tests**
  * Added comprehensive coverage for networking, pairing, streaming, state management, chats, profiles, routines, and Markdown handling.


KesleyDavid:
Still working on this, as mentioned — posting what has actually been exercised so far, and what has not, so the picture is not left to guesswork.

## Automated

**925 tests green** against `main` at `3557e748`: 258 in `:core` (this PR's module) and 667 in the app module that sits on top of it. `:core` is a plain Kotlin JVM module, so its suite runs anywhere without an Android device.

The `:core` tests are derived from the Swift fixtures in `ios/Tests/CompanionCoreTests` wherever an equivalent existed, so wire decoding, the SSE parser, the state fold, pairing/connection parsing, failover and the markdown splitter are pinned against the same cases the iOS core is.

## Exercised on a real device

Two emulators (API 34 and API 26 — the `minSdk` floor), both paired to a live companion on a real desktop, not a mock:

- **Install and launch, no crash**, on both API levels; no `FATAL` in logcat.
- **Pairing end to end**, both routes: mDNS discovery, and manual address + code. Confirmation screen shows name *and* address.
- **Permissions**, correct per version: asked one at a time on 33+, and correctly not asked at all on API 26.
- **Live SSE**: created a bot from the phone, sent turns, watched streaming deltas, the working indicator, activity receipts, tailed bubbles, date separators and per-bot mascots.
- **Rotation** in both directions with an unsent draft in the composer — draft survives, landscape reflows.
- **Process death** (`am kill`) and return: the stack restores into the conversation, not the roster, and the unsent draft survives.
- **Chrome comparison** API 26 vs 34: elevation, shadow, layout and typography.
- **Font at 200%**, and a **320dp narrow screen** — measured against the accessibility node bounds rather than eyeballed; zero nodes outside the viewport in either.
- **RTL** under an Arabic app locale: chrome, avatars, chevrons, bubble sides, localized dates.
- **TalkBack**: touch targets ≥48dp, semantics merging, disabled-state handling.

Three defects were found this way and fixed. None would have shown up in a diff review:

1. The SQL/data table card was **transposed** relative to `SQLResultTableView` — built column-major where iOS builds row-major, so a screen reader read `LANGUAGE, Python, Java, Rust, YEAR, 1991, …` and the row association was gone.
2. On an RTL device, English text took an RTL paragraph direction (Compose resolves an unspecified direction from the layout direction, unlike the platform `TextView`), so trailing punctuation jumped to the front and a counting reply reversed. Worth flagging for iOS: SwiftUI's default is `contentBased`, so this looks like a gap on our side rather than a shared one.
3. The app disconnected the stream immediately on backgrounding, so a turn finishing after the user pressed Home produced no notification at all. iOS holds the stream for 25 seconds via `linger()`; the port now has an equivalent window. Verified with a logging proxy in front of the companion: returning inside the window produces no new `GET /api/events`, returning after it produces exactly one, carrying the confirmed cursor.

## Not yet exercised

Being explicit rather than leaving it implied:

- Tapping a notification through to the exact thread; two notifications for one thread; re-pairing after `Unauthorized`; deep links; routine-editor draft isolation.
- Dark theme on any screen; the system photo picker versus the API 26 fallback; IME behaviour in modal sheets; font at 1.3×.
- Live approvals including the three-option card, group creation, task deletion while active, profile avatar upload/generate, routines end to end, transcript sharing.
- Size and cancellation limits on image attachments; audio interruption.
- On physical hardware only: NSD discovery on API 26–30 (the emulator's virtual network cannot settle it — its old `mdnsd` backend never sees a response through the AVD NAT, which is an environment limit rather than an app defect), real dictation, and the notification shade under RTL.

## On the delta since this PR opened

`main` has moved 183 commits since this branch's base, and the typed-endpoint work (`hosted`/`tailnet`/`lan`/`bonjour`, the trust ordering, `pairRequestId`, `GET /api/companion/endpoints`) landed after it. The port is being brought up to that contract now — it is the main thing still in flight, and it lives in exactly the module this PR contains. Happy to fold it in here or keep it as a follow-up, whichever you prefer when you get to this.

KesleyDavid:
Superseded by #513, and I want to be specific about why rather than just closing it.

This branch is `android/core` as it stood on 21 August. Since then two upstream parity deltas landed, and three of the things they fixed were security defects in exactly this code:

- **No typed endpoint.** `Connection` here strips `https://` and always rebuilds `http://`, and ignores the `endpoints` parameter from the QR. That is a real credential downgrade: a QR that offered hosted HTTPS could end up carrying the pairing credential over cleartext. `Endpoint.kt` does not exist on this branch.
- **No pairing preflight.** `POST /api/pair` goes out with no health check and no identity — `pairRequestId` appears zero times here, five times in #513. The QR credential was also burned before any I/O, so a timeout wasted the invite and a lost response left an orphaned device.
- **No trust class in failover.** Hosts rotate as a raw list, so the long-lived bearer could be offered to any LAN or Bonjour address that answered.

Merging this first would put that on `main` until #513 followed. That did not seem like a reasonable thing to ask of you, so the whole port is in one PR instead.

Nothing is lost. #513 contains this core plus the five parity deltas, and the history is still the five-way split proposed in #241 — 41 commits, each one reviewed before it landed, so `git log` reads the way the separate PRs would have.

The CodeRabbit comments on this PR were addressed; the ones that were real are fixed in #513, and the ones that were not, I said why at the time.

For the record, the count in my 25 August comment here is stale: it was 925 tests then, it is 1112 now, and the port has since been exercised on a physical phone and an API 26 emulator.

PR #354, opened 2026-08-22
fix(drivers): flush assistant text before tools

### LINKED ISSUE #352: ACP/Pi/Box flush assistant text only at turn settle, so message → tool order is inverted
### Before filing

### Area

Provider drivers (claude / codex / grok)

### What happened

On ACP, Pi, and Box, a turn whose real event order is **message → tool → message → tool** is shown and persisted as **tools first, then one concatenated assistant bubble at the end**.

This is not an array-sort bug in the chat view. The live tail (`MessagesList`, then `StreamingBubble`) is intentional — iOS does the same — and it is already correct for Claude and Codex. Those drivers emit `item.completed` (`assistant_text`) *before* `item.started` (`tool`). The harness (`server/index.ts`) then appends a bot text message, the client `clearStream`s on that `message` frame, and the next tool chip lands after it.

ACP/Pi/Box never emit that mid-turn `assistant_text`. They `+=` every `content.delta` into one string and flush it only in `settle()` / prompt completion. So while tools are running, the stream bubble stays at the bottom of the list (below the chips), and the persisted transcript is one late blob instead of separate messages.

Claude already has the canonical sequence (contract test locks it in):

```
content.delta → item.completed (assistant_text) → item.started (tool) → …
```

ACP's happy-path test currently *expects the inverted order* and comments it as "assistant_text (summed) on settle":

```
content.delta → item.started (tool) → item.completed (tool) → item.completed (assistant_text) → turn.completed
```

The fake ACP CLI already emits `agent_message_chunk` then `tool_call` (`server/testing/fake-acp-cli.ts` `playTurn`). The driver is what delays the text until settle (`server/drivers/acp/core.ts` `state.text += delta`, flush only in `settle()`). Pi is the same (`assistantText` flushed in `settle()`). Box is the same (`lastText` flushed when the prompt finishes).

The client live buffer in `src/state/store.tsx` is not independently wrong: it already `clearStream`s when a settled bot text `message` arrives, not only on `turn.completed`. Without a mid-turn `item.completed`, that path never fires.

Permission cards have the same visual failure: `request.opened` becomes an options message while the preamble is still in the stream, so the bubble sits under the card.

## Suggested direction

Add a `flushAssistantText()` on ACP and Pi: if the buffer has non-whitespace text, emit `item.completed` (`assistant_text`) and reset the buffer. Call it before `item.started` (tool) and before `request.opened`, and keep the existing settle flush for the tail.

Box is slightly different: `lastText` is a full-text prefix snapshot, not a delta accumulator. Resetting it to `""` would re-emit the already-flushed prefix on the next full-text event. Track a flushed offset and emit only the unflushed suffix.

Leave Antigravity / plain Grok chat alone — they only receive the text at the end, so there is nothing to split.

Update `server/drivers/acp/acp.test.ts` to expect Claude's order, and add a `message → tool → message → tool` fixture on the ACP and Pi fakes. The existing Pi `tooluse` fake starts with a tool, so it would not catch this.

No client or iOS change should be required if the drivers emit in the right order. The harness already maps `item.completed assistant_text` → a text message and concatenates multiple of them in `askBotAndWait`.

Happy to send the PR once this direction is ok — CONTRIBUTING asks to agree on behavioural changes first.

### Version and install

from source · `main` @ `89d25dd` · v0.1.27 · macOS

### Logs

Not a crash. Repro is the existing ACP contract test (`server/drivers/acp/acp.test.ts` "normalizes a full turn into the canonical event sequence"): fake CLI sends text then a tool, driver emits the text only at settle.

donggyun112:
## What changed

ACP, Pi, and Box now emit `item.completed` (`assistant_text`) at tool and permission boundaries instead of concatenating every delta until `turn.completed`.

Box keeps a `pendingText` buffer of actually streamed deltas (reset on flush) so a later non-prefix response is not sliced away. Fake CLIs gained an `interleave` mode; Box got its first contract tests against a scripted HTTP fake.

## Why

A real turn that is **message → tool → message → tool** was shown and persisted as tools first, then one late assistant bubble. Claude and Codex already emit text before tools; ACP's happy-path test had locked in the inverted order. See #352.

## How it was verified

- `pnpm typecheck`
- `pnpm test` — 147 files, 1514 passed, 12 skipped
- Driver contract tests: ACP/Pi `interleave` plus Box prefix-growth, non-prefix-after-flush, and non-prefix prompt-result cases
- No UI code change; live order follows once drivers emit `item.completed` before `item.started`

## Screenshots (UI changes)

n/a — no renderer changes

## Checklist

Closes #352

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response handling so assistant messages appear in the correct order around tool activity.
  * Ensured text is displayed before permission requests, tool executions, and completion states.
  * Preserved all assistant text blocks, including trailing messages and final results.
  * Added fallback messaging for successful runs with no assistant output.

* **Tests**
  * Expanded coverage for interleaved assistant messages, tool events, completion states, and failure handling across supported drivers.

### DIFF
--- a/server/drivers/acp/acp.test.ts
+++ b/server/drivers/acp/acp.test.ts
@@ -225,10 +225,10 @@ describe("ACP turns (fake CLI)", () => {
       "turn.started",
       "session.started",
       "content.delta",
+      "item.completed", // assistant_text before the tool, not summed on settle
       "item.started", // tool tc-1
       "item.completed", // tool tc-1 done
       "thread.token-usage.updated",
-      "item.completed", // assistant_text (summed) on settle
       "turn.completed",
     ]);
     expect(recorder.events.every((e) => e.turnId === turnId && e.provider === "grokAgent")).toBe(true);
@@ -239,6 +239,34 @@ describe("ACP turns (fake CLI)", () => {
     const done = recorder.events.at(-1)!;
     expect(done).toMatchObject({ type: "turn.completed", ok: true });
     expect(instance.adapter.hasSession("t-happy")).toBe(false);
+  });
+
+  it("emits each assistant text block before the tool that follows it", async () => {
+    await create(GrokAgentDriver, "interleave");
+    await instance.adapter.sendTurn({ threadId: "t-interleave", text: "go", model: "grok-4.5" });
+    await recorder.until((e) => e.type === "turn.completed");
+
+    const types = recorder.events.map((e) => e.type);
+    expect(types).toEqual([
+      "turn.started",
+      "session.started",
+      "content.delta",
+      "item.completed", // before one
+      "item.started", // tc-1
+      "item.completed", // tc-1
+      "content.delta",
+      "item.completed", // before two
+      "item.started", // tc-2
+      "item.completed", // tc-2
+      "content.delta",
+      "thread.token-usage.updated",
+      "item.completed", // after — no following tool, so settle flushes
+      "turn.completed",
+    ]);
+    const texts = recorder.events
+      .filter((e) => e.type === "item.completed" && (e as { itemType?: string }).itemType === "assistant_text")
+      .map((e) => (e as { text: string }).text);
+    expect(texts).toEqual(["before one", "before two", "after"]);
   });
 
   it("reads token usage from the root of the prompt result", async () => {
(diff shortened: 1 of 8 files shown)


milind-soni:
Carried forward on current main with terminal ordering, interruption/failure flushing, whitespace-buffer cleanup, and focused driver coverage, then merged in #382. Closing this original in favor of the merged replacement; thank you @donggyun112.

PR #356, opened 2026-08-22
fix(steer): keep queued 1:1 sends off the transcript until drain

### LINKED ISSUE #355: Queued 1:1 sends become the active leaf mid-turn, so the current turn's remaining output hangs off the steering line
### Before filing

### Area

### What happened

A 1:1 message sent while the bot is busy is not a real pending input. It is appended as an ordinary user `Message` (`steered: true` or `queued: true`) and immediately becomes `activeLeafId`. Later `item.started` / `item.completed` events from the *current* turn parent to that line.

Rooms already hold the in-flight send client-side until the room settles (`src/components/Composer.tsx`). 1:1 does the opposite: Composer posts immediately, and the harness lands it in the tree.

Live steer and queue fallback are different:

| | Live steer (`steered`) | Queue fallback (`queued`) |
|---|---|---|
| Did this turn's model see it? | yes | no |
| Correct order | flushed prior assistant text → user line → later tools/text of *this* turn | current turn finishes first, *then* the user line starts the next turn |
| Today | stream still sits at the bottom while the user line is already in the list | current-turn tools/text appear *after* a line the model has not seen |

Queue fallback can persist:

```
user
queued steering          ← current turn never saw this
remaining assistant/tool of the current turn
next turn answering the queued text
```

It should be:

```
user
assistant/tool of the current turn
steering
next turn
```

`appendMessage` always moves the leaf (`server/store.ts`). Busy POST always appends (`server/index.ts` live-steer path, `server/steer-queue.ts` `queueSteeredMessage`). The unit test *requires* immediate transcript insert (`server/steer-queue.test.ts` "appends a queued user message to the thread immediately"), and drain is written to reuse that already-inserted message so `startTurn` does not append a duplicate (`server/index.ts` `drainQueuedSends`).

Related to #352 (streaming text held outside the transcript until too late) but this is a leaf/parent bug, not a driver flush. #354 should not grow to cover it.

## Suggested direction

- **Queue fallback:** keep pending inputs *out* of `messages[]` until drain. On `turn.completed`, append them (leaf is then the last item of the finished turn) and start the follow-up. The room Composer already does this client-side.
- **Live steer:** flush the in-flight assistant stream first (the #352/#354 rule), *then* commit the `steered` user message. Later tools of that same turn hanging off that line is correct.
- UI: render queue fallback as a pending row outside the transcript, same idea as the room "Queued — sends when … finishes" chip.

A monotonic sequence on every event is not required if those two rules hold — array order is then the order.

Happy to send a PR once this direction is ok — CONTRIBUTING asks to agree on behavioural changes first.

### Version and install

from source · `main` @ `89d25dd` · v0.1.27 · macOS

### Logs

Not a crash. Repro: start a long 1:1 turn on an engine without live steer (or one whose `steer()` fails), send a second message while busy, then let the first turn emit a tool or assistant_text. The second user line sits above that remaining output. `steer-queue.test.ts` currently asserts the immediate-append shape.

donggyun112:
## What changed

Queue-fallback mid-turn sends (engines without live `steer`) no longer `appendMessage` immediately. They wait in the in-memory steer-queue; on `turn.completed` drain appends them (leaf is then the finished turn's last item) and starts one follow-up turn with the joined prompt.

The 1:1 composer shows the existing room pending chip while those lines are held, so they are visible without sitting on the transcript path.

Live Claude `steer` is unchanged: the model saw the line, so it still commits as `steered: true` when sent.

## Why

`appendMessage` always moves `activeLeafId`. A queued user line in the middle of a live turn made remaining tool/assistant events parent to a message the current turn never saw.

```
user
queued steering          ← current turn never saw this
remaining assistant/tool of the current turn
next turn
```

should be:

```
user
assistant/tool of the current turn
steering
next turn
```

See #355. Separate from #352/#354 (driver text flush).

## How it was verified

- `pnpm typecheck`
- `pnpm test` — 146 files, 1508 passed, 12 skipped
- `server/steer-queue.test.ts` and `server/steer-e2e.test.ts` (queue stays off the transcript while busy; drain appends after the first turn; Claude live-steer still mid-turn)

## Screenshots (UI changes)

n/a — pending chip already existed for rooms; 1:1 now uses it instead of a transcript row

## Checklist

Closes #355

## Summary by CodeRabbit

* **New Features**
  * Busy one-to-one messages remain hidden until processing begins.
  * Queued messages appear as regular user messages and are combined into one follow-up turn.
  * Group conversations continue to display queued messages with a discard option.
  * Pending one-to-one messages remain visible when switching conversations until processed.

* **Bug Fixes**
  * Improved queued-message handling during active bot responses.
  * Prevented stale or duplicated queued messages from appearing in conversation history.
  * Improved queued-message status tracking across conversations.
  * Ensured all queued messages are excluded from transcript replay after processing.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -1152,12 +1152,13 @@ bus.subscribe((event: RuntimeEvent) => {
 });
 
 function drainQueuedSends() {
-  drainSteeredMessages(store, (botId, threadId, prompt, userMessage) =>
+  drainSteeredMessages(store, (botId, threadId, prompt, userMessage, excludeIds) =>
     // A plain attended turn — no automationSource, no unattended, no comms
     // depth: exactly what typing the same words into an idle bot would run.
     // Drain just appended the held lines; userMessage keeps startTurn
-    // from duplicating the last one.
-    startTurn(botId, prompt, { threadId, userMessage }).catch((err) => {
+    // from duplicating the last one, and excludeIds drops every drained
+    // line from the transcript-replay so they are not also in `prompt`.
+    startTurn(botId, prompt, { threadId, userMessage, excludeMessageIds: excludeIds }).catch((err) => {
       store.appendMessage(threadId, {
         role: "bot",
         kind: "activity",
@@ -1282,6 +1283,8 @@ async function startTurn(
   opts?: {
     commsDepth?: number;
     userMessage?: Message;
+    /** Extra transcript ids to omit (every drained queued line, not just the last). */
+    excludeMessageIds?: string[];
     /** Routines run in detached tasks; pin the destination for the whole turn. */
     threadId?: string;
     /** Cloud routines run the whole agent inside the bot's Box VM instead
@@ -1350,9 +1353,10 @@ async function startTurn(
 
   // transcript for API-backed drivers: settled text turns on the ACTIVE
   // branch only — abandoned forks never reach the model
+  const skipTranscript = new Set<string>([userMessage.id, ...(opts?.excludeMessageIds ?? [])]);
   const transcript = store
     .activePath(threadId)
-    .filter((m) => m.kind === "text" && m.text && m.id !== userMessage.id)
+    .filter((m) => m.kind === "text" && m.text && !skipTranscript.has(m.id))
     .slice(-40)
     .map((m) => ({ role: m.role === "user" ? ("user" as const) : ("assistant" as const), text: m.text! }));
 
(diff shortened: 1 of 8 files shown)


milind-soni:
Superseded by #387, which carries this work onto current main and fixes the remaining unbounded cross-client queue-tombstone issue with regression coverage. Thank you @donggyun112 for the original implementation.

PR #361, opened 2026-08-22
fix(cursor): send the session's ACP model id, not the argv slug

### LINKED ISSUE #360: Cursor: session/set_model is sent argv slugs, so every model fails with -32602
# Cursor driver: `session/set_model` sends argv slugs into the ACP model namespace, so every model fails with -32602

**Repo:** milind-soni/OpenMausBot
**Area:** `server/drivers/acp/cursor.ts`
**Severity:** Cursor is unusable — every model selection fails, not just unknown ones.
**Found on:** `cursor-agent 2026.08.11-e8db854`, OpenMausBot @ `8c95d11`, Linux.

## Symptom

Any turn on a Cursor instance with a model selected fails:

```
Cursor rejected model "auto" via session/set_model: Invalid params.
Check that cursor-agent is current and that this account can use that model.
```

The message points at the CLI version and the account's entitlements. **Both are
red herrings** — the CLI is current and the account can use the model. That
wording sent me to check a subscription over what is purely an id-format bug.

## Root cause

Cursor exposes **two model namespaces that do not match**:

| Surface | Ids |
|---|---|
| `cursor-agent models`, `--model` flag | `auto`, `gpt-5.3-codex` |
| ACP `session/new` → `models.availableModels` | `default[]`, `gpt-5.3-codex[reasoning=medium,fast=false]` |

`configureSession` passes `turn.model` — an argv slug — straight into
`session/set_model`, which accepts only the parameterised ACP ids. Note `auto`
is not merely spelled differently; the ACP entry is called `default[]`.

## Reproduction (no OpenMausBot needed)

Drive `cursor-agent acp` over stdio: `initialize`, `session/new`, then
`session/set_model`.

```
set_model("auto")                  -> -32602 Invalid params
set_model("gpt-5.3-codex")         -> -32602 Invalid params
set_model("default[]")             -> OK
```

`session/new` advertises:

```json
{"currentModelId": "gpt-5.3-codex[reasoning=medium,fast=false]",
 "availableModels": [{"modelId": "default[]", "name": "Auto"},
                     {"modelId": "grok-4.6[effort=high,fast=true]", "name": "grok-4.6"},
                     {"modelId": "claude-opus-5[thinking=true,context=300k,effort=high,fast=false]", ...}]}
```

**A concrete model fails too**, so this is not "that account cannot use auto".

## Why it is fatal rather than degraded

`configureSession` tolerates `-32601` (method absent) but rethrows everything
else, and a throw there fails the turn. `-32602` therefore kills a request that
would otherwise have run correctly — `spawnArgs` already pinned `--model`, which
the file's own comment calls "the reliable pin".

## Fix

Two parts:

1. **Resolve the id against what the session advertised.** `configureSession`
   could not do this before: the ctx has `request`, `sessionId`, `config` and
   `turn`, but not `session/new`'s result. Patch adds `sessionModels` to the ctx
   in `acp/core.ts` (it is already in scope as `sessionResult`), and
   `resolveCursorAcpModelId()` in `cursor.ts` maps exact id → base-before-`[` →
   display name → the `auto`/`default` special case.

2. **Treat `-32602` like `-32601`.** When the id cannot be mapped, argv has
   already pinned the model, so refusing the turn is strictly worse than
   proceeding.

Returning `null` from the resolver falls back to the raw slug, so a CLI that
advertises no models behaves exactly as before.

## Tests

- Unit: `resolveCursorAcpModelId` — `auto`→`default[]`, bare→parameterised,
  already-parameterised passthrough, empty list → null, unknown → null.
- Wiring: fake ACP CLI extended with `FAKE_ACP_SESSION_MODELS` (off by default,
  so existing modes stay byte-identical) to advertise Cursor-shaped ids; asserts
  `set_model` receives `default[]` while argv keeps `--model auto`.
- Regression: new `FAKE_ACP_MODE=set-model-invalid-params` returns `-32602`;
  asserts the turn still completes.

Existing cursor/acp tests pass unchanged.

gmuniz21:
Fixes #360.

## The bug

Cursor keeps two model namespaces that do not match:

| Surface | Ids |
|---|---|
| `cursor-agent models`, `--model` | `auto`, `gpt-5.3-codex` |
| ACP `session/new` → `models.availableModels` | `default[]`, `gpt-5.3-codex[reasoning=medium,fast=false]` |

`configureSession` passed `turn.model` — an argv slug — straight into `session/set_model`, which accepts only the parameterised ids. So **every** model returned `-32602 Invalid params`, not merely unknown ones. Since `auto` is `STATIC_CURSOR_MODELS.default`, a stock install could not run a single Cursor turn.

The thrown message blamed the CLI version and the account's entitlements, which sends people to check a subscription over an id-format mismatch.

## Verified against `cursor-agent 2026.08.11-e8db854`

```
set_model("auto")                                    -> -32602 Invalid params
set_model("gpt-5.3-codex")                           -> -32602 Invalid params
set_model("default[]")                               -> OK
set_model("gpt-5.3-codex[reasoning=medium,fast=false]") -> OK
```

## The fix

`configureSession` could not have solved this alone — its ctx had `request`, `sessionId`, `config` and `turn`, but not `session/new`'s result. So:

1. **`acp/core.ts`** passes `sessionModels` into the ctx (already in scope as `sessionResult`). Additive and optional; no other driver changes behaviour.
2. **`acp/cursor.ts`** adds `resolveCursorAcpModelId()`: exact id → base before `[` → display name → the `auto`/`default` special case. **No match returns `null` and the raw slug is sent exactly as before**, so a CLI advertising no models is byte-identical to today.
3. **`-32602` is now tolerated like `-32601`.** `spawnArgs` already pinned `--model`, so the turn runs the right model regardless; throwing refused a request that would have succeeded.

## Tests

- 5 unit cases for the resolver (incl. empty list → `null`, unknown → `null`)
- Wiring case: `set_model` receives `default[]` while argv keeps `--model auto`
- Regression: a `-32602` still completes the turn
- Fake ACP CLI gains `FAKE_ACP_SESSION_MODELS` and a `set-model-invalid-params` mode, **both off by default** so existing modes stay byte-identical

Full suite green locally: 146 files, 1519 passed, 8 skipped.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved model selection for ACP-based Cursor sessions by matching requested models with available model IDs and display names.
  - Added support for automatic/default model selection and parameterized model identifiers.
  - Prevented unsupported model-setting responses from interrupting a turn when the command-line selection is already valid.
  - Improved error reporting when model selection is rejected for other reasons.

### DIFF
--- a/server/drivers/acp/core.ts
+++ b/server/drivers/acp/core.ts
@@ -127,6 +127,12 @@ export interface AcpSupport {
     sessionId: string;
     config: AcpConfig;
     turn: SendTurnInput;
+    /** `session/new` (or `session/load`) advertised model list, verbatim. Some
+     * CLIs namespace their ACP model ids differently from their argv `--model`
+     * slugs (Cursor answers `default[]` where the CLI calls it `auto`), so a
+     * driver that only knows the argv slug cannot form a valid set_model
+     * without this. Empty when the agent advertised none. */
+    sessionModels: Array<{ modelId?: string; name?: string }>;
   }): Promise<void>;
 }
 
@@ -605,6 +611,9 @@ export function createAcpDriver(support: AcpSupport): ProviderDriver<AcpConfig> 
                   sessionId,
                   config,
                   turn: cliTurn,
+                  sessionModels: Array.isArray(sessionResult?.models?.availableModels)
+                    ? sessionResult.models.availableModels
+                    : [],
                 });
                 // initialize's currentModelId is the CLI default (grok-4.6),
                 // not the model this turn asked for. After a successful pin,
--- a/server/drivers/acp/cursor.test.ts
+++ b/server/drivers/acp/cursor.test.ts
@@ -16,6 +16,7 @@ import {
   decodeCursorModelCatalog,
   decodeCursorModelText,
   STATIC_CURSOR_MODELS,
+  resolveCursorAcpModelId,
 } from "./cursor.ts";
 
 const FAKE_CLI = join(dirname(fileURLToPath(import.meta.url)), "..", "..", "testing", "fake-acp-cli.ts");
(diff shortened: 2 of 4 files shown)


milind-soni:
Carried forward on current main with the requested resolver coverage, tested across the ACP/Cursor suites, and merged in #379. Closing this original in favor of the merged replacement; thank you @gmuniz21.

PR #363, opened 2026-08-22
fix(hermes): offer the model Hermes' own config names, not only local hosts

gmuniz21:
## The problem

Hermes' model picker offers **only** local inference hosts (Ollama, LM Studio, EXO, oMLX, Unsloth). But Hermes is a BYOK harness whose own `hermes setup` writes a hosted provider key into `~/.hermes/.env`.

So with a hosted provider configured, the picker shows:

```
Local models — No local models found
Start oMLX, Ollama, Unsloth, LM Studio, or EXO, then reopen this picker.
```

…and the agent is greyed out and unselectable — while `hermes` is installed, authenticated (`state: available`, `authenticated: true`) and perfectly able to answer from its own config.

## The fix

`resolveModels` now also offers the model Hermes' config names, **when a hosted key is actually configured**.

The option is deliberately **not** a local-inject id. That matters: `hermesAcpModelId` returns `null` for it, so `configureSession` sends no `session/set_model`, and `spawnArgs` passes no `-m` (ACP ignores it anyway). Hermes therefore falls through to its own configured provider — which is exactly the intent.

**Read-only.** `ensureHermesInjectProvider` writes `config.yaml`; doing that from a catalog probe would rewrite the user's real Hermes config as a side effect of opening a menu. This only reads.

**Returns `null` when no hosted key is set**, so local-only installs keep precisely the catalog they have today.

## Verified

Against Hermes Agent v0.20.5 with an OpenRouter key configured:

- picker offers `anthropic/claude-opus-4.6 (Hermes config)`
- a real turn completes: `turn.completed ok: true`, reply received, no runtime errors
- spend registers on the provider side afterwards

## Tests

- configured key → offered
- **commented-out key → not offered** (the shipped `.env` carries `# OPENROUTER_API_KEY=`; treating that as configured would offer a model that cannot authenticate)
- no `.env` → `null`, local-only setups unchanged
- unreadable `config.yaml` → still usable, generic label
- the id maps to no ACP model id

Full suite green locally: 1524 passed, 8 skipped.

## Summary by CodeRabbit

* **New Features**
  * Hermes now supports OpenRouter-hosted model discovery when an API key is configured.
  * Configured and discovered models retain their provider-specific IDs and appear as custom options.
  * Configured, remote, and local models are combined without duplicates, with the configured model selected by default.
  * Model discovery no longer changes Hermes configuration files.

* **Tests**
  * Added coverage for hosted keys, commented-out keys, missing or unreadable configuration, custom flags, local mappings, and model discovery scenarios.

### DIFF
--- a/server/drivers/acp/hermes.test.ts
+++ b/server/drivers/acp/hermes.test.ts
@@ -1,0 +1,58 @@
+import { mkdtempSync, writeFileSync, mkdirSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { removeTempDir } from "../../testing/cleanup.ts";
+import { HERMES_CONFIG_MODEL_ID, hermesAcpModelId, hermesConfiguredModel } from "./hermes.ts";
+
+describe("hermesConfiguredModel", () => {
+  const dirs: string[] = [];
+  afterEach(async () => {
+    for (const d of dirs.splice(0)) await removeTempDir(d);
+  });
+
+  const home = (env: string, cfg?: string) => {
+    const root = mkdtempSync(join(tmpdir(), "omb-hermes-"));
+    dirs.push(root);
+    const h = join(root, ".hermes");
+    mkdirSync(h, { recursive: true });
+    writeFileSync(join(h, ".env"), env);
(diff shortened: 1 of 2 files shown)


milind-soni:
Carried forward on current main with blank-key parsing, bounded ACP discovery, process cleanup, and the outstanding review fixes, then merged in #380. Closing this original in favor of the merged replacement; thank you @gmuniz21.

PR #370, opened 2026-08-22
docs: plan reusable team channels

milind-soni:
## Summary
- propose Teams and Channels so one saved bot roster can work across several isolated project conversations
- build on the existing Room/Group thread model instead of creating a parallel conversation engine
- define the runtime changes required for the same bot to work safely in multiple channels at once
- cover working-folder/computer leases, memory isolation, migration, API contracts, mobile behavior, tests, and small implementation PRs

## Scope
Planning document only. This PR contains no schema, UI, API, or runtime implementation.

## Related work
- #342 can provide the team persistence and sidebar-switcher foundation if it lands
- #339 remains an optional folder-to-team onboarding flow
- neither PR currently provides multi-channel hierarchy or safe simultaneous turns for the same bot

## Validation
- reviewed against the current `GroupRecord`, room turn queue, Team Library import boundary, and global `bot.busy` behavior
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Added a planning specification for reusable Teams and project-scoped Channels.
  * Defined channel-specific conversations, settings, transcripts, unread state, memory, and resource ownership.
  * Documented team management, bot roster inheritance, security boundaries, mobile behavior, migration, and deletion workflows.
  * Outlined APIs, events, phased delivery, testing requirements, completion criteria, and open implementation decisions.

### DIFF
--- a/docs/plans/2026-08-22-channels-plan.md
+++ b/docs/plans/2026-08-22-channels-plan.md
@@ -1,0 +1,382 @@
+# Channels: reusable teams across projects
+
+- Status: proposed
+- Date: 2026-08-22
+- Scope of this pull request: planning only
+
+## Summary
+
+Add **Teams** and **Channels** so one saved roster of bots can work across several project conversations without duplicating the bots or mixing project context.
+
+A channel should be a project-scoped conversation with its own transcript, working folder, shared instructions, responder policy, unread state, and runtime sessions. A team should be a reusable roster. The same designer, product manager, and engineer can therefore appear in `Website launch`, `Q4 kickoff`, and `Acme renewal`, while each channel remains isolated.
+
+This should extend the existing Room implementation rather than introduce a second conversation engine. OpenMausBot already gives every room a thread, members, bulletin, working folder, responder policy, and transcript. The work is primarily a product hierarchy plus a runtime-isolation change.
+
+## Why this is not only a sidebar feature
+
+The current data model already lets the same bot id belong to several rooms. However, `runGroupMemberTurn` rejects a room turn whenever `bot.busy` is true. This deliberately prevents one bot from running in two conversations at once.
+
+That guard is correct for today's runtime: activity, interruption, some computer backends, and several cleanup paths still assume one active turn per bot. Removing it only in the UI would create crossed streams, interrupt the wrong task, or let two channels control one computer.
+
+The feature is complete only when the same bot can safely run in multiple channels at once on supported engines and resources. A visual grouping that still says “busy in another conversation” does not meet the product promise.
+
+## Product model
+
+Use four distinct concepts:
+
+- **Bot**: a reusable agent identity—name, role, avatar, engine defaults, permissions, connected-app policy, and durable personal memory.
+- **Team**: a saved roster of existing bots. A team does not clone or own the bots.
+- **Channel**: one project conversation for a team, with isolated context and project settings.
+- **Task**: a focused unit of work inside a bot or channel conversation, preserving the existing task model.
+
+“Channel” becomes the user-facing name for a normal multi-bot room. The backend may keep `GroupRecord`, `/api/groups`, and `group` events during migration so this feature does not require a risky all-at-once rename. Auto-created bot-to-bot DM channels remain an internal/special case and do not appear under Teams.
+
+## User experience
+
+### Sidebar
+
+Add a **Teams** area to the existing sidebar. Each team expands to its channels:
+
+```text
+Teams
+  Launch team
+    # q4-kickoff
+    # website-launch
+    # acme-renewal
+  Research team
+    # weekly-brief
+```
+
+Each channel row shows its latest speaker/preview, unread state, and active member count. Busy state belongs to the channel. A team may optionally show an aggregate such as “2 active” but should not look blocked because one member is working elsewhere.
+
+Standalone existing rooms appear as standalone channels in a compatibility section. Users can attach one to a team later without losing its transcript.
+
+### Creating a team
+
+The user can:
+
+1. create a team from existing bots;
+2. save the roster produced by a Team Library import; or
+3. convert an existing room roster into a team.
+
+Creating a team never creates duplicate bots. Deleting a team never deletes its bots.
+
+### Creating a channel
+
+From a team, **New channel** asks for:
+
+- channel name;
+- optional working folder;
+- optional shared instructions/bulletin;
+- responder policy: one default member, everyone, or mentions only; and
+- roster mode: inherit the team or customize this channel.
+
+The channel opens immediately as an empty project conversation. It does not seed every bot with a greeting.
+
+### Managing membership
+
+Channels inherit their team's roster by default. Team roster changes propagate to inherited channels. A channel can switch to a custom roster; after that, team changes do not silently replace its membership. The UI must label the mode and offer **Reset to team roster**.
+
+Removing a member does not erase messages already attributed to that bot. Re-adding the member restores participation but starts future turns from the visible channel context rather than reviving a hidden stale session.
+
+### Deletion behavior
+
+- Delete a channel: delete only its transcript and channel runtime state.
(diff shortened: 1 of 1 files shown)


maxkongerskov:
Hey Milind — nice to see this stacked on #342. The “same roster, several projects, don’t clone the bots” sentence is the right product.

A few notes from living in the team-switcher model, in case they’re useful before the implementation slices:

**How #342 works today, as a workflow.** You pick a team at the top of the sidebar (or All bots). That team is the place you are standing. New bot and new room both join *that* team — they get its `teamId` on create, and a new room is pre-filled with that team’s people. Switch to another team and the list is a different cast: those bots, those rooms, that search/import/export. You are not taking the same Designer with you. Designer-on-Launch stays on Launch; Research is other people. All bots is the escape hatch if you need the whole workspace. Delete team only unassigns; it does not delete the bots.

That is why the current UI feels straightforward. A team is a separate roster you work inside, not one squad you drag through several project chats.

**Why that loop matches the harness.** Today a bot is one turn and one computer. Create a bot in Launch, then switch teams, and you are looking at different people — so “busy” and “which Designer is this?” stay obvious. You never have to promise that the same agent is in two chats at once. Stand in a team, new bot and new room join it, switch and that cast is gone, All bots if you need the whole house: that is a rule that fits the runtime we actually have. A nested tree that treats a team as a reusable cast across `#` channels is a bigger model; with global `bot.busy` still in place, those channels would still grey the bot out everywhere.

**Switcher vs channels.** #342 is a filter: pick a team, the sidebar shows that roster. Rooms stay a flat list. This plan is a hierarchy: pick a team, then pick a project conversation under it. That’s a real product step, not a small delta on the switcher. I’d treat the nested sidebar as replacing the switcher UX rather than rebasing it in place — otherwise people learn two different meanings of “team.” (Yours: same bots in `#website-launch` and `#q4-kickoff`. #342: switching teams means you left those bots.)

**What `teamId` on a room means.** In #342 a room is tagged for a team when every member already shares that team. Here the team owns the roster and inherited channels follow it (unless customized). Same field, different owner. If #342 lands first, PR 1 is probably a small migration of that contract rather than “extract the foundation unchanged.” Happy to align the store shape if you want that to be a clean handoff.

**Busy vs the promise.** You already call this out: a nested list that still uses global `bot.busy` isn’t the feature. I’d keep that in the delivery order — either isolation with the new nav, or don’t present PR 2 as channels yet. The fake version (two `#` rooms, Designer greyed out because they’re busy next door) is the one worth avoiding. In #342 that case does not come up in the same way: switch team, and you are looking at a different bot, so the busy one is not sitting in the other channel as the same person.

The rest I would keep: `GroupRecord` under the hood, don’t duplicate bots, don’t delete bots with the team, isolate transcripts/folders/computers. That matches what #342 was aiming at.

#342 still has merge conflicts and a couple of integrity follow-ups in flight. If it’s easier, extracting just `TeamRecord` + assignment + events as groundwork (as you already allow for) is totally fine with me.

Thanks for writing it down in the open.

milind-soni:
First usable Channels slice shipped in #390: existing rooms migrate in place to Channels, users can create and move them under Work/Personal/project contexts, and the #381 rename/member editor carries over. The broader same-bot concurrency and resource-isolation phases in this plan remain follow-up work.

milind-soni (commented):
Reviewed — looks good. The strongest part is 'Why this is not only a sidebar feature': calling out that `runGroupMemberTurn`'s busy-guard is load-bearing (interruption, computer backends, cleanup all assume one active turn per bot) is exactly the honest framing that keeps an implementer from shipping crossed streams. Extending `GroupRecord`/`/api/groups` rather than building a parallel engine, and keeping channel-busy state visible instead of hiding it until multi-turn safety actually lands, are the right calls. Planning-only scope, green CI — ready to merge as direction-setting.

maxkongerskov (commented):
The remaining plan has a **compatibility problem** with what already shipped, and a **workflow collision** if the later phases land on that language unchanged. I would not merge this as direction-setting until those two meanings are split.

## What already shipped vs what this doc still promises

**#390** productized rooms as **Channels**: New Channel, context (Work / Personal / a project), move between contexts, same transcript/folder/roster as today’s groups. It explicitly does **not** promise the same bot working in two conversations at once. `bot.busy` stays the safety boundary.

**This plan** still defines a Channel as something else: a project conversation **under a reusable team roster**, with inherited vs custom membership, and a definition of done that requires the same bot to stream in two channels at once.

Users who get #390 will learn: a channel is a room with a context label. If later PRs keep this document’s meaning, they will be taught: a channel is one of several projects for the *same* saved squad, and Mira in `#website-launch` is the same Mira in `#acme-renewal` and can be busy in both. Those are not the same product. Shipping the second on the first’s name is the bug.

**#342** (if it ever lands) is a third meaning of **team**: a filter you stand in. New bot / new room join *that* roster; switch teams and you are looking at different people. This plan’s team is the opposite: one squad you reuse across many `#` chats. Same word, opposite navigation rule. The doc currently says to treat #342 as foundation for PR 1. That would bake the collision into the store (`teamId` on bots *and* on groups, with different owners).

**#384** already files rooms and bots under shared section labels. #390’s contexts sit on that same axis. Nested `Teams → #channels` is a second hierarchy on the same sidebar.

## Workflow that actually breaks

Not “people might not want Slack.” The break is:

1. **Busy is global.** `runGroupMemberTurn` still bails when `bot.busy`. The plan is honest that a nested list without conversation-scoped turns is a fake. #390 shipped the nested-ish *language* (Channel) while keeping that guard. Fine for now. Unsafe if PR 2 of this plan presents inherited channels as “your Launch team, several projects” while Mira greys out in every other channel because she is typing next door. That is a workflow lie: the UI says she is a member of both; the runtime says she can only exist in one.

2. **Inherited roster vs explicit members.** Today a channel/room has `memberIds`. You add/remove people; that is the membership. Live inheritance (“team roster changes propagate”) means adding a bot to Launch silently puts them in `#website-launch` *and* `#q4-kickoff`. That is a different mental model from #390/#381, where you edit *this* conversation’s members. Mixing inherit + custom + “reset to team” on top of the editor users just got will fight that editor.

3. **Same bot, two computers/folders.** Even without parallel *reasoning*, the plan’s later leases exist because one `BotRecord` still owns one Local VM / host-computer grant. Putting that identity in two project folders without leases is how two channels edit one checkout. That is a compatibility issue with the current one-bot-one-computer assumption, not a sidebar issue.

## Proper fix

Do not use this document as the Channels roadmap anymore. Channels already means #390. Rewrite the leftover work as two *separate* follow-ups, neither of which redefines Channel or Team against the shipped UI.

**Follow-up A — roster template (no runtime change).**  
A named list of bot ids. “New channel” can *copy* that list into `memberIds` once. After create, membership is the channel’s, edited with the existing #381 panel. No live inheritance, no `rosterMode`, no nested Teams tree. This is the “don’t clone Mira for the next client” workflow. It fits today’s `bot.busy` and today’s Channel.

**Follow-up B — conversation-scoped turns (only if A is actually painful).**  
`TurnKey = botId:threadId`, then leases for the same folder / this computer. Do this because Scout has two *tasks* or two channels and the user is blocked, not because the sidebar looks like Slack. Until B exists, the UI must not promise parallel work. Status is “Mira is in #website-launch” — same as today.

**Do not:** rebase this plan onto #342 as if `teamId` meant the same thing; nest `#` under Teams in the sidebar while #390 contexts/sections already group channels; keep “same bot, two live streams” in the definition of done for a nav slice.

I would close this planning PR or retitle it to “roster templates + optional parallel turns” with an explicit note that **Channel is already shipped in #390** and must not change meaning. Direction-setting that disagrees with the last merge is how we get two products in one sidebar.

milind-soni:
Closing this earlier architecture plan: channels and separate channel tasks have since landed in #390 and #537 using the current GroupRecord/GroupTaskRecord model. The proposed TeamRecord/roster-inheritance plan no longer matches that implementation. This does not mean its entire parallel-turn/resource-lease wishlist has shipped; remaining gaps should be tracked as focused follow-ups.

PR #376, opened 2026-08-23
feat(harness): redact secrets in tool results before storage

willsigmon:
## Summary

Tool **results** were the one transcript input that bypassed redaction. `Store.appendMessage` scrubbed everything the *bot authored* (reply text, tool titles, card copy) via `redactBotAuthored`, but a `cat .env`-style tool output lands in the persisted thread as an activity chip's tool name — and since rebuilds replay activity into every handed-over context, a secret printed by a command was stored (and re-fed to models) in plaintext, permanently.

This PR closes that gap at the single store choke point rather than at each driver/append call site:

- `server/store.ts`: new `redactToolResult` pass runs inside `appendMessage`, applying the existing `redactSecretsInText` to `tool.name` on activity messages regardless of role — except user-authored ones, which stay exactly as typed (same rule as user text today).
- Value-shape rules for high-signal credential formats already exist in `server/redact.ts` (`AKIA…` AWS key ids, `sk-`/`sk-proj-` API keys, `ghp_`/`gho_`/… and `github_pat_` GitHub tokens, `xox[bposar]-` Slack tokens, PEM `PRIVATE KEY` blocks) — this PR puts those rules to work on tool results too, with tests proving each shape is masked in raw command output.
- Over-redaction guard: no generic long-token/hex rule is added; near-miss fixtures (short prefixes, one-char-short AKIA, plain commit hashes, prose) assert ordinary text survives so debugging usefulness is preserved.
- Marker style is unchanged: the same `«redacted N chars»` convention as every existing call site.

Note: the audit evidence this task was scoped from claimed these shape rules were missing from `redact.ts`; they were in fact already present on origin/main with passing tests. The genuine gap was the missing result-side hook, which is what this PR implements.

## Changes

- `server/store.ts`: add `redactToolResult`; fold it into `appendMessage`; document why it sits beside (not inside) `redactBotAuthored`.
- `server/redact.test.ts`: new "tool-output shapes" suite — positive masks for sk-, AKIA, gho_, github_pat_, xoxp- and PEM bodies in raw tool output; near-miss negatives proving short/partial shapes are left untouched.
- `server/store.test.ts`: end-to-end store-time test (`cat .env` dump containing an sk-proj key gets masked in both the returned message and the persisted record while surrounding prose survives), plus a guard test that user messages with `tool.name` are never rewritten.

## Test Plan

- [x] `export VITEST_CACHE_DIR=$PWD/.vitest && pnpm typecheck` — passes
- [x] `npx vitest run server/redact.test.ts server/store.test.ts` — 61/61 passed (all 59 pre-existing redaction/store tests unmodified)
- [x] `npx oxlint server` — no findings on any added line (repo has a large pre-existing anti-slop baseline on untouched files)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protection for secrets appearing in bot-generated tool results.
  - Redacts recognized OpenAI, AWS, GitHub, Slack, and PEM private-key patterns in stored tool-result titles and content.
  - Preserves ordinary hashes, near-miss values, and user-authored tool messages without unintended changes.
  - Ensures redaction remains applied to persisted messages.

### DIFF
--- a/server/redact.test.ts
+++ b/server/redact.test.ts
@@ -165,4 +165,46 @@ describe("redactSecretsInText", () => {
     expect(out.command).toContain("«redacted");
     expect(out.note).toBe("fine");
   });
+
+  // Tool results are what actually leak: a `cat .env` or `printenv` dumps
+  // real credentials into the transcript as plain output. The same content
+  // rules apply — but a near-miss must survive, because debugging usefulness
+  // matters and a generic "long token" rule would rewrite real code.
+  describe("tool-output shapes (value-shape rules)", () => {
+    const alpha = "abcdefghijklmnopqrstuvwxyz0123456789";
+    it("masks every high-signal secret shape in raw tool output", () => {
+      const cases: Array<[string, RegExp]> = [
+        [`OPENAI_API_KEY=${"sk" + "-"}${alpha}`, /sk-[a-z]/],
+        [`aws ${"AKIA" + "ABCDEFGHIJKLMNOP"} example`, /AKIA[A-Z0-9]{16}/],
+        [`bare openai shape ${"sk" + "-"}${alpha.slice(0, 24)} trailing`, /sk-[a-z]/],
+        [`token: ${"gh" + "o_"}${alpha.slice(0, 36)}`, /gho_/],
+        [`fine-grained ${"github_" + "pat_"}11ABCDEFG0${alpha}`, /github_pat_/],
+        [`slack user token ${"xox" + "p-"}2400-${"1234567890123"}-${alpha.slice(0, 24)}`, /xoxp-/],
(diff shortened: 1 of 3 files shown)


milind-soni:
Thanks for tracing the redaction path. On current `main`, `appendMessage` already passes every bot-authored activity through `redactBotAuthored`, which applies `redactSecretsInText` to `tool.name`; `Message.role` is only `bot | user`. The new store test uses `role: "bot"`, so it is already covered by that existing path and the additional helper does not change behavior.

Closing this as a no-op to avoid maintaining two redaction passes. A follow-up would be welcome if it includes a concrete fixture that leaks on `main` and passes only with the change.

PR #388, opened 2026-08-23
fix(drivers): auto-retry transient failures in claude/codex/grok

willsigmon:
## Summary
A provider hiccup (429 / overloaded / 5xx / connection reset) killed a turn outright in the three main CLI drivers even though box/webhook drivers already rode out transients. This adds one shared retry policy and wires it into claude, codex, and grok.

## Changes
- **`server/drivers/retry.ts`** (new): `classifyError` sorts failures into transient (`rate_limited`, `overloaded`, `server_error`, `connection_reset`, `timeout`) vs terminal (`auth`, `quota`, `unknown_model`, `invalid_request`, `not_found`, user-interrupted); `computeBackoff` is capped 1s/3s/8s with ±25% jitter; `interruptibleDelay` makes the wait cancellable so an interrupt during backoff settles the turn instead of relaunching (no zombie processes).
- **Runtime events**: new `turn.retrying` event in `contracts.ts` carrying `{ attempt, delayMs, reason }`; handled exhaustively across transcript/thread-event/inspector switches so each retry is visible.
- **Drivers**: claude/codex/grok relaunch up to `RETRY_MAX_ATTEMPTS = 3` on transient-only failures. Terminal errors never retry. A turn that already streamed assistant text **never** auto-retries — the duplicate-text hazard outweighs the recovery. Test-scale env knobs (`FAKE_*_RETRY_SCALE`) keep tests instant.
- **Tests**: new `retry.test.ts` (10 cases: classification positives/negatives, backoff bounds + jitter determinism, cancellation) plus per-driver retry suites via the existing fake-CLI/app-server infrastructure: transient-twice-then-success produces exactly two `turn.retrying` events and one final message; terminal errors and partial-output turns produce zero retries.

## Verification
- `pnpm typecheck` — clean
- `npx vitest run server/drivers/retry.test.ts` — 10 passed
- `npx vitest run server/drivers/grok.test.ts` — 7 passed
- `npx vitest run server/drivers/claude.test.ts` — 48 passed, 1 skipped (pre-existing skip)

## Summary by CodeRabbit

* **New Features**
  * Added automatic retries for temporary provider and service failures before response streaming begins.
  * Retry attempts use increasing delays and stop after a defined limit.
  * Retry progress now displays the attempt, delay, and failure reason.
  * Added support for handling interactive tool requests during Codex sessions.
* **Bug Fixes**
  * Stopping a turn now cancels pending retries promptly.
  * Retries no longer occur after assistant content has started streaming or for permanent failures.
  * Failed retries provide clearer final error and interruption handling.

### DIFF
--- a/server/contracts.ts
+++ b/server/contracts.ts
@@ -87,6 +87,14 @@ export type RuntimeEvent = RuntimeEventBase &
     | { type: "session.exited"; reason?: string }
     | { type: "turn.started" }
     | {
+        type: "turn.retrying";
+        /** 1-based: the retry about to be launched (1 = first relaunch). */
+        attempt: number;
+        delayMs: number;
+        /** Why this failure was judged retry-worthy (classifyError's reason). */
+        reason: string;
+      }
+    | {
         type: "turn.completed";
         ok: boolean;
         stopReason?: string | null;
--- a/server/drivers/claude.test.ts
+++ b/server/drivers/claude.test.ts
@@ -157,6 +157,10 @@ describe("ClaudeDriver turns (fake CLI)", () => {
   afterEach(async () => {
     delete process.env.FAKE_CLAUDE_MODE;
     delete process.env.FAKE_CLAUDE_DUMP;
+    delete process.env.FAKE_CLAUDE_TRANSIENTS;
+    delete process.env.FAKE_CLAUDE_PARTIAL_FAILS;
+    delete process.env.FAKE_CLAUDE_STATE;
+    delete process.env.FAKE_CLAUDE_RETRY_SCALE;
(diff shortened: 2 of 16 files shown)


milind-soni:
Rebased onto current main and fixed all seven review findings: retry state/broker cleanup, per-turn Codex interrupt state, orphan app-server retirement, fetch restoration, attempt display, and strict event validation. Added regression coverage for fresh retry budgets and cross-thread interruption.

Local validation: full suite green — 1,636 passed, 12 skipped — plus broker, updater, desktop viewer, and packaged-server smoke tests.

Issue #255, opened 2026-08-19
Feature request: bot credential vault / password manager — needs a bot-invisible login broker


aivsomkar:
## Goal

Let a bot sign into the user's **own** accounts (Gmail → Drive/Sheets, and the long tail of password sites) **inside an isolated computer (Local VM / Cloud box)** — without wiring the OAuth plugin. It must be:

- **Robust / hard to breach** across the threat model: *both* a tricked bot *and* an attacker with machine access.
- **Cross-platform**: macOS, Windows, iOS, Android (companion apps) — not Mac-only.
- **MIT-friendly with no paid dependency**: OS keystores + open-source tools; GPL tools only over a process boundary (link only MIT code).

**Hard invariant:** the model must *never* learn the secret — not in a prompt, tool arg, tool result, transcript, or a screenshot it reads.

## What already exists (PR #246, branch `feat/credential-vault`)

The **storage + management half** is built and tested and is sound:

- `electron/vault.cjs` — `Vault` over Electron `safeStorage` (OS keychain: Keychain / DPAPI / libsecret), `vault.bin` `0600`, separate from `credentials.bin`. `list()` returns metadata only; `reveal()` is Touch-ID gated; `matchForFill({origin, botId, context})` scopes by exact origin + allowed bot + context and refuses lookalike / ambiguous matches.
- `SettingsModal` + `VaultSection.tsx` — the Vault UI (add/edit/remove, scope to bots + contexts, per-row Touch-ID reveal).
- `electron/vault-bridge.cjs` + `server/vault-client.ts` — a **trust boundary**: the secret lives only in Electron main; the harness server may only *request* a fill over a token-guarded loopback endpoint and receives an *outcome only* (never a secret). RFC 6238 TOTP is generated in main from a stored seed.
- `docs/superpowers/specs/2026-08-19-bot-credential-vault-design.md` — full spec (threat model, a `CredentialProvider` port with adapters, per-platform matrix, licensing table, phased plan).

This part can stay. **The delivery mechanism (Phase 1b) is what's blocked.**

## The blocker (why this is a feature request, not a ready PR)

The intended Phase 1b was: *type the secret into the isolated computer's browser from Electron main, so the model never sees it.* **This does not actually hide the secret from the model.**

[...]

PR #397, opened 2026-08-23
feat(mcp): give a bot MCP servers of your own

aivsomkar:
## What changed

OpenMausBot mounts plenty of MCP servers — Composio, the computer, peer comms, dweb — and every one of them is the harness's. There was no way to hand a bot a server *you* chose, which is the ordinary case: a filesystem server rooted at your project, a database server, anything you already run against another MCP client.

A bot now carries a list of stdio MCP servers: a command on this computer with args and env, enabled or not. Enabled ones are mounted into both seams that already assemble `mcpServers` — the Claude driver (where the tools are also pre-allowed, or a headless `acceptEdits` run silently denies them) and the ACP session, in its `{name,value}[]` env shape. There's an MCP servers card in the bot's profile to add, toggle and remove them.

## Relationship to #61

@carbongotfound's PR #61 built this first and was asked for changes. That review is effectively the spec for this one, so it's worth being explicit about how each finding is handled.

Three of the six **do not exist here**, because v1 is stdio-only: there is no URL to strip userinfo credentials from, no remote response body to bound, and no bracketed IPv6 loopback to parse. That is most of why the scope is drawn where it is.

The other three are the design:

- **`id` is the identity.** A name is a label you edit, so nothing that routes a turn derives from it — renaming a server cannot re-point anything at a different one. *(their "updates can change routed IDs")*
- **Two servers may never fold onto one key.** The agent addresses a server by name, so a collision means one silently wins. It is refused when you **save**, where a person reads the error, not at turn time where nobody is looking. *(their "sanitized names can collide")*
- **An env value never leaves the process.** `wireBot` — the single projection every bot payload passes through — replaces values with `true`. The renderer learns which variables are set, never what they hold. Sending `true` back means "keep the stored one", which is what lets you rename a server without its secrets ever having been in the browser. *(their "secrets in snapshots")*

Validation is a zod schema at the boundary, matching `bot-profile.ts`, so API errors are the schema's message rather than raw exception text.

## Why

Requested directly: the app has no custom MCP or plugin support, and every other MCP client has it.

## How it was verified

- **9 unit tests** for the registry (`server/mcp-servers.test.ts`) — key folding, collision refusal, id-survives-rename, the `true` placeholder resolving against storage, the placeholder with nothing behind it being refused, and every malformed shape a hand-written PATCH can send. Written against a stub and watched failing first.
- **A driver test** proving the Claude seam mounts a custom server, pre-allows `mcp__<key>`, and keeps the value off argv. Verified it genuinely bites by stashing the wiring: `expected undefined to match object { command: 'npx', … }`.
- **An API test** proving `env` arrives as `{ TOKEN: true }`, that the secret appears nowhere in any payload, that a rename preserves the id, and that a collision is a 400.
- **End to end in a browser**: added a server through the card with `API_TOKEN=sup3rs3cret`. It persists to `bots.json` with the real value; the API payload and the DOM contain neither. *(My first probe was wrong — it typed into the bot's Name field and renamed the bot — which is exactly why this was checked against stored state rather than the screenshot.)*
- `pnpm typecheck` clean, 1700 tests pass, `oxlint` count **identical to main** (1405).

## Known limitation

Env values live in `bots.json` beside the credentials already there — consistent with how the app stores secrets today, but plaintext on disk. The credential vault (#255) is the real answer.

## Not in scope

Remote HTTP servers, health inspection / `tools/list`, global (all-bot) servers, a catalog. Those carry the remaining risks from the #61 review and belong in their own change.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added per-bot local MCP server management in Settings.
  * Create, edit, enable, disable, and remove servers with commands, arguments, and environment variables.
  * Enabled custom servers are now available during bot interactions.
* **Security & Validation**
  * Environment values are protected when displayed or returned.
  * Added validation for server names, commands, arguments, environment variables, duplicates, and malformed configurations.
* **Bug Fixes**
  * Preserved server identities when names are changed.

### DIFF
--- a/server/contracts.ts
+++ b/server/contracts.ts
@@ -194,6 +194,11 @@ export interface SendTurnInput {
     /** dweb network daemon: an MCP proxy exposing dweb status, repo, and
      * opencode model access as tools. url is the dweb HTTP base. */
     dweb?: { url: string };
+    /** The user's own stdio MCP servers for this bot, already filtered to
+     * the enabled ones and keyed by `mcpKey(name)`. Unlike every other
+     * entry here the harness does not own these processes' behaviour — it
+     * only spawns what the user configured. */
+    custom?: Array<{ key: string; command: string; args: string[]; env: Record<string, string> }>;
   };
   cwd?: string;
 }
--- a/server/drivers/acp/core.ts
+++ b/server/drivers/acp/core.ts
@@ -247,6 +247,10 @@ export function createAcpDriver(support: AcpSupport): ProviderDriver<AcpConfig> 
         if (agents) {
           servers.push({ name: "agents", command: agents.command, args: agents.args, env: acpEnv(agents.env) });
         }
+        for (const custom of turn.integrations?.custom ?? []) {
+          if (servers.some((server) => server.name === custom.key)) continue;
+          servers.push({ name: custom.key, command: custom.command, args: custom.args, env: acpEnv(custom.env) });
+        }
         const composio = turn.integrations?.composio;
         if (composio) {
           servers.push({
--- a/server/drivers/claude.test.ts
+++ b/server/drivers/claude.test.ts
@@ -343,6 +343,32 @@ describe("ClaudeDriver turns (fake CLI)", () => {
     expect(seen.argv[seen.argv.indexOf("--allowedTools") + 1]).toContain("mcp__dweb");
   });
 
+  it("mounts the user's own MCP servers and pre-allows their tools", async () => {
+    await create();
+    const dump = join(scratch, "dump.json");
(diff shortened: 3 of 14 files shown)


milind-soni:
Closing this older implementation as superseded by the shipped custom MCP manager in #697. Main now has Plugins → MCP servers, persisted command/args/env, write-only secret handling, enable/disable/test/delete, and the provider bridge. Thank you for the original per-bot design and validation work. Per-bot assignment is not claimed as shipped by #697; if we add it, it should be a focused extension of the existing registry rather than a second competing MCP configuration model.

PR #410, opened 2026-08-23
fix(rooms): hold the turn ceiling while a human decides

### LINKED ISSUE #289: Room turns are capped at a flat 5 minutes, even while streaming and even while parked on an approval

OWConnoi:
Fixes #289 — the sharper half, in the direction that issue's fallback suggests and that #284 already committed to: keep the configurable room ceiling, but stop counting time the turn spends parked on a person.

## What changes

The room timer is no longer a bare `setTimeout` over wall time. `RoomTurnDeadline` (in `server/room-turn-timeout.ts`, replacing `scheduleRoomTurnTimeout`) runs the configured budget as **work time**:

- `request.opened` on the room thread holds the clock; however long the human takes to answer an approval or question, none of it counts.
- `request.resolved` resumes the budget where it left off.
- Streaming, tool runs, and provider latency still burn the budget exactly as before, so a busy-but-slow member still hits the same configured ceiling.
- A turn that goes silent entirely remains the stall watchdog's job (`OMB_TURN_STALL_MS`, waiting-on-human exempt) — unchanged.

The wiring lives in the existing bus subscription inside `runGroupMemberTurn`; nothing else about room turn settlement (busy state, chained mentions, stall grace release) moves. Requests are counted with a clamping depth rather than a boolean, because stale `request.resolved` events can arrive for cards a given turn never opened.

## Why this shape

#284 chose to keep a hard, configurable ceiling for rooms rather than delete it in favor of the watchdog. Given that, this PR takes #289's minimum ask: exempting the wait-on-human window stops the cap from manufacturing stranded approval cards — the exact state #274/#292 had to build recovery paths for. The timeout note itself is unchanged and stays truthful: it now fires after N minutes of actual work.

## Tests

- `server/room-turn-timeout.test.ts`: deadline precision, cancel-on-settle, hold/resume across one and several sequential approvals (only the last resolve re-arms), stale resolves ignored without firing early, `stop()` during a held clock, plus the existing stall-registry and message-formatting coverage.
- `pnpm typecheck` clean; full `pnpm test` green (162 files, 1,693 passed, 12 skipped).
- Note: `pnpm lint` fails on current `main` independent of this branch (pre-existing anti-slop findings in driver files untouched here); the files in this PR lint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Turn timeouts now pause while awaiting human approval or responses and resume afterward.
  * Prevented stale, duplicate, or canceled timeout actions from completing turns incorrectly.
  * Timed-out turns now record the timeout and finish with a clear timed-out result.
  * Overdue turns now time out immediately when active work resumes, improving consistency for delayed interactions.


OWConnoi:
Addressed the review finding in 1270104: `hold()` now expires immediately when the card event lands after the deadline already passed, so a human's wait can never extend an exhausted budget. Added a regression test that advances the clock past the deadline without running timer callbacks, opens the card, and asserts exactly one expiry.

Issue #310, opened 2026-08-20
Apple Silicon: pinned trycua/xfce-cua ARM64 image has zero-byte OpenSSL libraries


tek123168:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Computer use (cloud box / local)

## Summary

I'm seeing this on an Apple Silicon Mac (arm64) with Docker Desktop. OpenMausBot's Local VM setup fails while preparing the CUA desktop because the pinned ARM64 base image appears to contain zero-byte OpenSSL libraries.

The user-visible error is:

~~~text
Prepare Cua desktop -> docker build -> curl fails
curl: error while loading shared libraries: /lib/aarch64-linux-gnu/libssl.so.3: file too short
~~~

## Reproduction

OpenMausBot currently pins:

~~~text
docker.io/trycua/xfce-cua@sha256:274eb636f5cf3fc58f705916ee72b7a701270b3877369d08533a385c5325be9b
~~~

The generated Dockerfile starts from that digest and installs cua-driver 0.20.0. On Apple Silicon it selects the aarch64 wheel.

To rule out a local cache problem, I removed the image, ran docker builder prune -af, confirmed that docker image inspect returned “No such image,” and pulled the digest again with --platform linux/arm64. The fresh image still contains:

~~~text
/lib/aarch64-linux-gnu/libssl.so.3 = 0 bytes
/lib/aarch64-linux-gnu/libcrypto.so.3 = 0 bytes
~~~

So curl cannot even start, before cua-driver installation begins.

## Compatibility concern

For comparison, trycua/cua-xfce:0.1.15 has a healthy ARM64 Ubuntu 22.04 filesystem, with libssl.so.3 at about 616 KB. However, it is roughly 5.3 GB versus roughly 1.79 GB for the pinned image, and curl is not installed by default.

I would not change only the FROM line without checking compatibility first. In particular, could you confirm the newer image still provides the cua user, /opt/venv/bin/python, supervisord, DISPLAY=:1/X11, xset, the expected noVNC/VNC ports and startup, and /home/cua/workspace?

## OpenMausBot context

~~~text
CUA_DRIVER_VERSION = 0.20.0
BASE_IMAGE_REPOSITORY = docker.io/trycua/xfce-cua
BASE_IMAGE_DIGEST = sha256:274eb636f5cf3fc58f705916ee72b7a701270b3877369d08533a385c5325be9b
IMAGE_LAYER_VERSION = 3
[...]

PR #411, opened 2026-08-23
fix(localvm): reject a zero-byte OpenSSL base image before the wheel fetch

OWConnoi:
Addresses the mitigation half of #310 — the base-image sanity check with a clearer error before cua-driver installation, which the issue explicitly asks for.

## What changes

`managedImageDockerfile()` now opens its install RUN with a gate that, for the target architecture's multiarch triplet, rejects any present-but-empty `libssl.so.3` / `libcrypto.so.3` in the base layer:

~~~text
pinned base image is defective on aarch64: /lib/aarch64-linux-gnu/libssl.so.3 is zero bytes,
so curl cannot start — re-pull or replace the base image instead of debugging the wheel download
~~~

Today that same condition surfaces as `curl: error while loading shared libraries ... file too short` mid-build, which reads as a network fault or a broken wheel mirror. The gate names the actual problem at the one place equipped to act on it, and it costs one shell loop inside the existing RUN — no new steps, no host tooling.

Deliberately out of scope:

- **Missing** libraries are not flagged; the dynamic loader's own message already names them precisely, and guessing at alternate layouts risks blocking a healthy future base.
- **The digest bump itself** needs what the issue says it needs: pulling both images, confirming libssl sizes, and walking the compatibility surface (cua user, `/opt/venv/bin/python`, supervisord, `:1`/xset, noVNC ports, `/home/cua/workspace`) before repinning `BASE_IMAGE_DIGEST`/`IMAGE_LAYER_VERSION`. That belongs in a change with real ARM64 build evidence behind it, which I'd rather not fake.

## Tests

- New case asserts the generated Dockerfile checks both triplets' libraries for zero-byte content and that the gate precedes `curl -fsSL` in the same RUN.
- The extracted RUN was executed under plain `sh`: zero-byte lib → exit 1 with the message above; healthy libs → proceeds to the fetch; absent libs → unchanged behavior.
- `pnpm typecheck` clean; full `pnpm test` green (162 files, 1,694 passed). No new lint findings versus current main (the pre-existing anti-slop findings in these two files are untouched and unshifted beyond line offsets).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved room turn timeouts to pause while awaiting human approval or questions and resume afterward.
  * Added safeguards against duplicate timeout handling and stale responses.
  * Managed images now detect missing OpenSSL libraries early and provide actionable diagnostics.

* **Tests**
  * Expanded coverage for paused, resumed, nested, stopped, and expired room turns.
  * Added validation for OpenSSL library checks before driver installation.


OWConnoi:
The branch was originally cut on top of another open PR's branch, so this diff briefly carried that change too — which is what the description/diff mismatch flag picked up. It is now rebased onto current main and carries only the image-gate change (2 files).

PR #412, opened 2026-08-23
fix(desktop): saved workspace keys were destroyed by the next launch

OWConnoi:
Saving an API key in the packaged app worked for the session and was then **gone after a restart** — the key had to be entered again every launch. Reproduced on 0.1.32/macOS; the on-disk fingerprint is a config.json whose credential sections end up as empty objects (e.g. `"opencodeGo": {}`).

## Root cause — two features reading the same field with opposite meanings

1. `credential:set` (electron/main.mjs) commits the real secret to the OS-encrypted store, then calls `PUT /api/config?secretStorage=external`. That route **tombstones every supplied credential as `""`** in config.json so no plaintext survives (server/index.ts, external-secret block).
2. The next boot, `migrateWorkspaceCredentials` (electron/workspace-credentials.mjs) swept plaintext into the store using semantics written for dev-mode files: non-empty = newest intent, **empty = "the user cleared this — drop the stored secret too"**, absent = keep.

So every packaged save left `apiKey: ""` behind, and the following launch obeyed it by deleting the just-saved key from credentials.bin. Composio's project key was spared only because its older sweep ignores empty fields; the five workspace credentials were not.

## Fix

The migration now treats an empty plaintext field as *no information*: the store stays authoritative, only a non-empty value overwrites it. This is lossless in every direction:

| disk field | store had | before | after |
|---|---|---|---|
| non-empty | old/newer | overwrite | overwrite (unchanged) |
| `""` tombstone after a packaged save | fresh key | **deleted (bug)** | kept |
| `""` after a UI clear | already removed | delete (no-op) | kept/no-op |
| absent | anything | keep | keep (unchanged) |

Clearing still works end-to-end because `credential:set("", …)` removes the store entry directly before the tombstone is persisted — there is no case where an empty field is the only evidence of a clear.

Also: `loadSecureCredentials()` returned `{}` silently when the OS store was unavailable at boot, which is indistinguishable from "never saved anything" — it now logs that specific condition.

## Tests

- Rewritten: empty-field migration keeps stored secrets and still sweeps the field.
- New: the full packaged cycle (plaintext migrate-in → external save leaves tombstone → next boot) asserts the key survives.
- Existing overwrite/idempotency/junk cases unchanged and passing.
- `pnpm test` green (162 files, 1,690 passed); `pnpm check:electron` clean (40 modules).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when secure credential storage is unavailable, preventing startup failures and allowing the app to continue without stored credentials.
  * Preserved saved credentials when configuration fields are empty or omitted.
  * Improved credential migration and removal behavior to prevent accidental loss of secrets.
  * Ensured credentials remain intact across application restarts after migration or configuration cleanup.

### DIFF
--- a/electron/main.mjs
+++ b/electron/main.mjs
@@ -68,8 +68,15 @@ let secureCredentials = {};
 const CREDENTIALS_FILE = path.join(app.getPath("userData"), "credentials.bin");
 
 async function loadSecureCredentials() {
+  if (!fs.existsSync(CREDENTIALS_FILE)) return {};
+  if (!(await safeStorage.isAsyncEncryptionAvailable())) {
+    // Returning nothing here looks identical to "the user never saved keys".
+    // Say why, so an OS-store hiccup is diagnosable instead of reading as
+    // wiped credentials.
+    slog("OS credential store unavailable; saved credentials are not loaded this launch");
+    return {};
+  }
   try {
-    if (!fs.existsSync(CREDENTIALS_FILE) || !(await safeStorage.isAsyncEncryptionAvailable())) return {};
     const decrypted = await safeStorage.decryptStringAsync(fs.readFileSync(CREDENTIALS_FILE));
     return JSON.parse(decrypted.result);
   } catch (error) {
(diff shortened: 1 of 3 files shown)


milind-soni:
Thanks for the fix

PR #414, opened 2026-08-23
fix(connectors): keep connection state truthful and durable across launches

OWConnoi:
Follow-up to the restart-persistence family (#412): two gaps where managed Connected Apps could *look* or become disconnected without anything actually being wrong — or get truly orphaned by a bad launch.

## 1. A failed inventory fetch read as "nothing connected"

`refreshConnectedStatus` swallowed every `/api/connectors/connected` failure (`.catch(() => ({}))`). After a restart against a slow or erroring broker, the Connected tab rendered its "No connected apps yet" empty state even though all connections were intact server-side — bots kept using them, which made the UI look simply wrong.

Now a failed fetch keeps last-known state and renders a retryable alert:

> Couldn't load your connected apps (Connected apps: HTTP 502…). Your connections are not lost — check your connection and try again.

Success clears it; the misleading empty state is suppressed while the alert is up; Marketplace tab is unaffected. Per-slug background polls keep their silent catch on purpose.

## 2. A fresh broker installation could be registered over an unreadable credential store

`ensureManagedComposioCredentials` treats "empty in-memory credentials" as "nothing registered yet". But `loadSecureCredentials` also returns empty when credentials.bin **exists yet could not be read** this launch (OS store briefly unavailable, decrypt failure). Registration then minted a new installation identity and persisted it over the unreadable file — permanently orphaning every connection bound to the old installation.

The load now records *why* it came back empty, and registration is skipped for that launch with a log line explaining why Connected Apps state is deferred, instead of manufacturing a new identity.

## Tests

- New cases for `inventoryFailureMessage`: retryable phrasing, service detail preserved, no "disconnected" implication.
- Existing merge/generation race tests unchanged and passing.
- `pnpm typecheck`, full `pnpm test` (1,692 passed), `pnpm check:electron` green; no new lint findings versus main.

Note for reviewers: open PR #346 also touches inventory-request staleness as part of its much larger Ubuntu scope — happy to drop or rebase this slice if that one lands first.

## Summary by CodeRabbit

- **Bug Fixes**
  - Connected-app inventory failures now display a clear, retryable warning with error details.
  - Existing connections remain visible when inventory loading fails, instead of appearing lost or empty.
  - Credential handling now distinguishes missing stores from unreadable encrypted stores to prevent unintended credential replacement.
  - Managed app registration waits until saved credentials can be read successfully.

### DIFF
--- a/electron/main.mjs
+++ b/electron/main.mjs
@@ -94,6 +94,16 @@ async function loadSecureCredentials() {
 }
 
 async function saveSecureCredentials(credentials) {
+  if (secureCredentialsUnreadable) {
+    // A store that exists but failed to read must never be replaced by a
+    // partial one: the file may hold recoverable credentials this launch
+    // cannot see. Every writer (boot migrations, credential:set) goes
+    // through here, so refusing covers them all; a launch that can read
+    // the store again is the recovery path.
+    throw new Error(
+      "The saved credential store could not be read this launch. Restart OpenMausBot before saving credentials so nothing is overwritten.",
+    );
+  }
   if (!(await safeStorage.isAsyncEncryptionAvailable())) {
     throw new Error("The operating-system credential store is unavailable");
   }
--- a/electron/main.mjs
+++ b/electron/main.mjs
@@ -64,16 +64,31 @@ app.on("second-instance", () => {
 let serverProc = null;
 let serverReady = true;
 let secureCredentials = {};
+// Set when credentials.bin exists but could not be read this launch (OS store
+// unavailable, decrypt/parse failure). Downstream boot steps check this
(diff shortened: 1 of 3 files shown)


OWConnoi:
Both findings addressed: every store write now goes through a guard that refuses to replace an unreadable credentials.bin (boot migrations and credential:set included — the migration callers already log-and-skip their config rewrite when the store write throws), and the empty-state suppression is limited to the Connected tab so Marketplace search still reports no matches.

milind-soni:
Superseded by #470, which carries these connection-truth and unreadable-credential-store protections plus request ordering and authoritative cache handling. Closing this duplicate without merging—thank you for identifying the original failure modes.

PR #415, opened 2026-08-23
feat(voice): built-in Mac voices — calls with no ElevenLabs key

OWConnoi:
Today voice is all-or-nothing on an ElevenLabs key: calls show *"Add an ElevenLabs API key so the bot can speak during calls"* and there is no other engine. This adds a second one — the voices already installed on the Mac — so calls and spoken replies work with zero accounts, zero keys, and zero network.

## What it is

`tts.provider` picks the engine:

- **`elevenlabs`** (default) — exactly as before, byte-for-byte behavior when no provider is set.
- **`system`** — `/usr/bin/say`, the same engine the Spoken Content pane uses. One utterance per request to match the existing client pipeline, rendered to 22 kHz mono WAV (`--data-format=LEI16@22050`, no conversion step), temp dir cleaned up after every call. Voices are listed by parsing `say -v ?` with its localized header treated as junk; multi-word voice names ("Bad News") parse correctly.

The renderer keeps playing opaque audio bytes from `POST /api/tts/speak`, so nothing client-side changes shape or learns which engine answered. Utterance splitting, the 500-char ceiling, the speaker singleton, and interruption semantics are all untouched — only the bytes' origin changes.

## Wiring

- `describeVoice` reports `provider`, and `configured`/`ready` become provider-aware: the system engine needs a picked voice, not a key, so calls unblock the moment a voice is chosen.
- Voice settings gain an engine picker, offered only where the desktop capability contract reports Darwin (never inferred from a user agent). The ElevenLabs key row appears only in ElevenLabs mode; switching engines re-fetches the voice list, and per-bot voice choices persist across the switch.
- Call-gating copy no longer assumes ElevenLabs; the no-key hint now mentions the built-in voices as the alternative.
- Platform-gated like everything macOS: elsewhere the picker simply doesn't appear and ElevenLabs behaves exactly as today.

Deliberately out of scope: provider-native speech (using the selected chat model's own voice) — different plumbing per driver, worth its own discussion; this PR is the zero-key path that works on every Mac today.

## Tests

- New suite: voice-table parsing (localized header ignored, multi-word names), synthesis through a fake `say` runner asserting exact argv (`-o`, `--data-format`, `-v`, text) and temp-dir cleanup, no-key readiness, and the missing-voice error staying distinct.
- Existing ElevenLabs stub tests unchanged and passing; `describeVoice` assertion updated for the new `provider` field.
- `pnpm typecheck` clean; full `pnpm test` green (162 files, 1,693 passed); no new lint findings versus main.

## Summary by CodeRabbit

* **New Features**
  * Added built-in Mac system voices alongside ElevenLabs.
  * Added a voice-engine selector with provider-specific guidance and automatic voice-list refresh.
  * Call setup now supports system voices and checks voices for all participants.
  * Added provider selection to text-to-speech configuration.
  * Added secret-request cards, message replies, team-map navigation, and VPS auto-start bot settings.

* **Bug Fixes**
  * Improved voice configuration checks and error messages.
  * Added cleanup and fallback handling during system voice generation.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -83,8 +83,10 @@ const appConfigSchema = z.object({
   vps: vpsConfigSchema.optional(),
   /** Optional OpenCode key; persisted write-only and passed only to its child. */
   opencodeGo: z.object({ apiKey: optionalText }).optional(),
-  /** Voice credentials and the selected voice id. */
-  tts: z.object({ key: optionalText, voice: optionalText }).optional(),
+  /** Voice credentials and the selected voice id. `provider` picks the
+   * engine: "elevenlabs" (default; needs a key) or "system" (the Mac's
+   * built-in voices, no key). */
+  tts: z.object({ key: optionalText, voice: optionalText, provider: z.enum(["elevenlabs", "system"]).optional() }).optional(),
   /** OpenAI key used only by the in-process avatar image generator. */
   imageGen: z.object({ key: optionalText }).optional(),
   /** Non-secret profile details shown in the sidebar. */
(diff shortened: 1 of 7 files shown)


milind-soni:
Interesting, will test it out! Thanks

PR #416, opened 2026-08-23
fix: keep the Agent profile model picker inside the sidebar

maxkongerskov:
## What changed

Opening the model picker from **Agent profile** used a 380px popover aligned to the chip. That ran out of the right sidebar, so the Cloud/Local rail sat under the chat.

In the profile the picker now expands **in-flow** under the Model row, full width of the card. The chat-header picker is unchanged.

## Why

You could not see or click the provider rail while setting up a bot from its profile.

## How it was verified

Clicked **Claude Sonnet 5** in Dense's Agent profile on the packaged app after the UI rebuild: Cloud/Local rail is fully visible inside the sidebar.

## Screenshots (UI changes)

**Before** — rail clipped by the chat:

Model picker overflowing the Agent profile into the chat

**After** — picker stays in the profile:

Model picker contained in the Agent profile sidebar

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a contained model selector layout for use within settings cards and other inline contexts.
  * Added optional labels for model selection controls.
  * Improved model selector presentation with compact styling and full-width menu support.

* **Style**
  * Updated the settings panel layout with a labeled card-based model selection section.
  * Improved panel layering to ensure controls display correctly.


milind-soni:
Love it. I do think we need to design the model picker itself to be better as well.

PR #419, opened 2026-08-24
fix(computer): make live desktop non-modal so control can be handed back

kaushik9999:
## Problem

Taking control of a bot's computer — **Take control** on a Cloud/Box or Local VM computer — opens the live desktop as a **modal** window (`openDesktopViewer`, `modal: true` in `electron/main.mjs`). A modal window disables the main app window while it's open, but the only **Hand control back** button lives in the Computer panel *in that main window* (`src/components/ComputerPanel.tsx`). So once the live desktop is open:

- you can't reach **Hand control back** — the app is blocked by the modal, and
- closing the viewer window doesn't release control either — the `closed` handler only clears the "viewer open" UI flag, it never releases the hold, so the bot's hands stay paused.

Net effect: after taking control you can get stuck — unable to hand control back or to cleanly close the live desktop.

## Fix

- **Non-modal viewer** — the live desktop keeps `parent` (so it still floats above the app) but is no longer `modal`, so the app stays usable and **Hand control back** is reachable while you watch/drive.
- **Closing the viewer hands control back** — the window's `closed` broadcast now releases control. `release()` is idempotent, so this is safe.
- **`desktopViewer.close()` bridge** — the panel's **Hand control back** button now also closes the viewer window, keeping the button and the window in sync.

Because the viewer is a separate window that outlives the panel, the close→release is reconciled at a scope that's always mounted rather than in the panel itself:

- **App-wide reconciliation** — the close→release listener lives in `Shell` (always mounted), so control is handed back even if the Computer panel was closed, or switched to another bot, while the viewer stayed open.
- **Context-changing replacement** — when one bot's viewer is replaced by another bot's, an explicit close is emitted for the previous bot so its control is released too. A same-bot reopen stays suppressed to avoid a spurious release.
- **Remount/switch initialization** — a (re)mounted panel initializes its viewer-open state from the live viewer, so switching to or reopening a panel while the viewer is open keeps the screenshot poll paused instead of resuming it.
- **Scoped close** — the **Hand control back** close request is scoped to the requesting bot, so a bot that holds control without a viewer of its own (e.g. a self-hosted VPS) can't close — and thereby release — another bot's open viewer.

## Testing

- `pnpm typecheck` passes
- `pnpm test:desktop-viewer`, `pnpm test:updater`, and the `computer-control` suite pass
- Traced OS-chrome close, panel-button close, panel-closed-while-viewer-open, cross-bot replacement, same-bot reopen, VPS (no viewer), and plain-web (no bridge) paths

## Summary by CodeRabbit

- **New Features**
  - Added controls to close the desktop viewer and check its current status.
  - The viewer now reflects the active bot context and updates control state when closed.
  - Desktop viewing is non-modal, allowing continued interaction with the app.
  - Returning control to the bot now closes its desktop viewer.

- **Bug Fixes**
  - Improved handling when switching between bot contexts.
  - Prevented stale viewer references and preserved screenshot polling behavior across remounts.

### DIFF
--- a/electron/main.mjs
+++ b/electron/main.mjs
@@ -410,12 +410,25 @@ function openDesktopViewer(owner, rawUrl, rawTitle, contextId) {
   const titleCandidate = Object.prototype.toString.call(rawTitle) === "[object String]" ? rawTitle.trim() : "";
   const title = titleCandidate ? titleCandidate.slice(0, 80) : "Live desktop";
 
+  const nextContextId =
+    Object.prototype.toString.call(contextId) === "[object String]" ? contextId.slice(0, 120) : null;
+
   // Desktop URLs contain rotating access tokens. A newly minted URL replaces
   // the old viewer instead of being retained anywhere after its window closes.
-  if (desktopViewerWindow && !desktopViewerWindow.isDestroyed()) desktopViewerWindow.close();
+  // Clear the ref first so the stale window's close handler no-ops; on a bot
+  // change, tell the previous bot to release (same-bot reopen stays quiet).
+  if (desktopViewerWindow && !desktopViewerWindow.isDestroyed()) {
+    const previous = desktopViewerWindow;
+    const previousOwner = desktopViewerOwner;
+    const previousContextId = desktopViewerContextId;
+    desktopViewerWindow = null;
+    previous.close();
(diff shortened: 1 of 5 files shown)


milind-soni:
Looks pretty good. Much better experience with VM use

PR #421, opened 2026-08-24
fix: recover VPS computers after image upgrades

cyburke:
## Summary

- detect a managed VPS container whose image no longer matches the current image layer before auto-provisioning
- show an explicit **Replace VPS computer** recovery action instead of issuing a Provision request that is guaranteed to return 409
- require confirmation before removing the disposable container, then remove and reprovision it through the existing guarded endpoints
- add a regression test for the stale-managed-container decision

## Reproduction

After updating OpenMausBot 0.1.27 → 0.1.32, an existing BYO-VPS bot still had `driver-0.20.0-v3`, while the app required `driver-0.20.0-v4`.

The Computer panel automatically called Provision. The server correctly refused to overwrite the existing incompatible container, but the panel exposed no Remove/replace action. The user was left at:

```text
Prepare the pinned OpenMausBot Cua image on the VPS (Driver 0.20.0)
```

The only recovery was an out-of-band API sequence:

```text
POST /computer/remove → POST /computer/provision
```

This PR makes that already-supported recovery discoverable and confirmed in the Computer panel.

## Safety

- replacement is offered only when the container is app-managed, present, and image-incompatible
- no automatic deletion: the user must click Replace and confirm the disposable-container warning
- unowned containers remain protected by the existing server-side ownership gate

## Verification

- `pnpm typecheck`
- `vitest run src/lib/vps-computer.test.ts server/vps-computer.test.ts` — 19/19 passed
- full `pnpm test` — 1,699 passed, 12 skipped (1,711 total)
- live recovery on a BYO-VPS installation: v4 image labels matched, container healthy/hardened/private, `desktopReady: true`, and a 1280×900 Cua screenshot succeeded

## Lint note

`pnpm lint` currently reports existing repository-wide anti-slop findings in untouched files; TypeScript and all test suites pass.

## Summary by CodeRabbit

* **New Features**
  * Added VPS compatibility checks for managed computers.
  * Displays a clear warning when a computer requires replacement.
  * Added a confirmation flow to replace and reprovision incompatible computers.
  * The desktop viewer now closes automatically when control is handed back.

* **Bug Fixes**
  * Prevents incompatible container images from continuing to be used.

* **Tests**
  * Added coverage for compatible, incompatible, missing, managed, and unmanaged computers.

### DIFF
--- a/src/components/ComputerPanel.tsx
+++ b/src/components/ComputerPanel.tsx
@@ -37,6 +37,7 @@ import {
   localComputerDisabledReason,
   localComputerSelectable,
 } from "@/lib/local-computer";
+import { vpsComputerNeedsReplacement, type VpsComputerStatus } from "@/lib/vps-computer";
 
 async function api(path: string, init?: RequestInit): Promise<any> {
   const res = await fetch(path, { headers: { "content-type": "application/json" }, ...init });
@@ -53,6 +54,7 @@ type Phase =
   | "vm"
   | "vm-unavailable"
   | "vps-unconfigured"
+  | "vps-incompatible"
   | "vps-stopped"
   | "local"
   | "local-unavailable"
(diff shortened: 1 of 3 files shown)


cyburke:
Live post-repair canary completed successfully on the original affected BYO-VPS bot: OpenMausBot provisioned `driver-0.20.0-v4`, reported `imageMatches: true`, `desktopReady: true`, and the bot used `computer___launch_app`, accessibility inspection, and `computer___page` to open https://example.com in Firefox and report the title **Example Domain**. The turn settled idle with no Cua image error.

milind-soni:
Appreciate you testing out and fixing the BYO VPS! Thanks a lot for the contribution. Will merge

PR #438, opened 2026-08-24
feat(pi): image attachments and reasoning-effort levels

cjpraia:
## What

Two small capability gaps between the pi engine and the other CLI engines:

### Images
The composer gates image paste on `capabilities.images`, and pi reads images fine — flip the flag so pi bots accept attachments like Claude, Codex, and Antigravity. No native plumbing needed: images already ride the prompt as `<attached-image path>` refs the agent opens with its own read tool.

### Reasoning effort
Declare `effortLevels` and pin each turn's reasoning effort via pi's `set_thinking_level` RPC, sent right after `set_model` (the supported level set is model-dependent). The harness and pi level sets match one-for-one except the lowest rung's name — harness `none` is pi's `off`. A rejected/unsupported level (e.g. `xhigh` on a model that lacks it) keeps the engine default instead of failing the turn.

## Testing

- Fake CLI extended to answer `set_thinking_level` and record the pinned level in the `FAKE_PI_DUMP` stream.
- New tests: capabilities advertise images + all six effort levels; a turn with `effort: "high"` pins exactly one level; `none` maps to `off`; a plain turn sends no thinking-level command.
- `pnpm typecheck` clean; full suite green (173 files, 1781 passed).

## Summary by CodeRabbit

* **New Features**
  * Pi now supports configurable effort levels, including disabling extended thinking.
  * Pi advertises image input support.
  * Thinking-level settings are applied automatically after model selection.
* **Bug Fixes**
  * Unsupported or unsuccessful thinking-level updates now safely preserve Pi’s default behavior.

### DIFF
--- a/server/drivers/pi.test.ts
+++ b/server/drivers/pi.test.ts
@@ -255,6 +255,51 @@ describe("PiDriver turns (fake CLI)", () => {
     expect(instance.adapter.hasSession("t-exit")).toBe(false);
   });
 
+  it("advertises images and every harness effort level", async () => {
+    await create();
+    expect(instance.adapter.capabilities.images).toBe(true);
+    expect(instance.adapter.capabilities.effortLevels).toEqual(["none", "low", "medium", "high", "xhigh", "max"]);
+  });
+
+  it("pins reasoning effort via set_thinking_level after the model", async () => {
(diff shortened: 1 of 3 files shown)


milind-soni:
Thanks for making the pi integration better! Appreciate it

PR #440, opened 2026-08-24
fix(drivers): scope local-computer approvals to real desktop-control tools only

johnsonAyo:
## What changed

`server/drivers/claude.ts`: the permission broker stamped `approvalScope: "local-computer"` onto **every** `request.opened` / `request.resolved` event whenever a bot had "This computer" enabled — including plain `Bash`, `Edit`, and `Write` asks. The server (`server/index.ts`) suppresses `allowKey` whenever an approval carries a scope, which is correct for desktop control but meant the **"Always allow" button never rendered for any tool** on any bot with local computer-use enabled.

Now:
- only tools whose name starts with `mcp__computer` (the real desktop-control namespace) carry the scope on `request.opened`;
- `request.resolved` looks up the ask's tool from a small per-turn map, so resolved events carry the same scoped logic.

## Why

Desktop-control approvals must stay one-at-a-time (no remembered grants), but ordinary CLI/file asks should be rememberable. Conflating them made a core UX affordance unreachable in a very common configuration.

Closes nothing — filed independently as a bug fix.

## How it was verified

- `pnpm typecheck` ✅
- `pnpm test` — full suite green: 163 files, 1,699 passed, 12 skipped ✅
- Updated `server/drivers/claude.test.ts` contract test: a `Bash` ask now asserts `approvalScope === undefined` (so the UI can offer a remembered grant), and a new `mcp__computer__screenshot` ask asserts the scope is still `"local-computer"` on both `opened` and `resolved`.
- Manually verified end-to-end on macOS against a live bot: "Always allow" renders for Bash/Edit cards and remembered grants stick; computer-use cards still prompt every time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission approval handling for desktop-control tools.
  * Standard command-line tools no longer receive unnecessary computer-access approval prompts.
  * Follow-up approval requests now consistently use the correct access scope and status.
  * Improved handling of approval requests when tool information is incomplete or invalid, helping prevent incorrect access prompts or unresolved approval states.

### DIFF
--- a/server/drivers/claude.test.ts
+++ b/server/drivers/claude.test.ts
@@ -715,18 +715,38 @@ describe("ClaudeDriver turns (fake CLI)", () => {
       tool: "Bash",
       summary: "rm -rf scratch",
       requestId: "ask-1",
-      approvalScope: "local-computer",
     });
+    // a plain CLI tool never carries the desktop-control approval scope,
+    // so the UI can offer a remembered grant for it
+    expect(opened).toHaveProperty("approvalScope", undefined);
 
     // the outcome names exactly what was granted: this action, once
     await expect(instance.adapter.respondToRequest("t-perm-abc", "ask-1", { behavior: "allow" })).resolves.toBe("allowed-once");
     expect(await answered).toMatchObject({ behavior: "allow" });
     const resolved = await recorder.until((e) => e.type === "request.resolved");
-    expect(resolved).toMatchObject({
-      behavior: "allow",
-      source: "user",
-      approvalScope: "local-computer",
+    expect(resolved).toMatchObject({ behavior: "allow", source: "user" });
+    expect(resolved).toHaveProperty("approvalScope", undefined);
+
+    // a real desktop-control tool keeps the local-computer scope, which
+    // suppresses remembered grants — desktop actions must be approved
+    // one at a time
+    const answered2 = new Promise<{ behavior: string }>((resolve) => {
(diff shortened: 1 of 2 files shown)


johnsonAyo:
Rebased onto current `main` — clean, full suite green (1,819 passing).

Flagging why this matters a bit more after #442 landed: that merge deliberately kept the **Always allow** button, and right now that button cannot render at all on any bot with computer use enabled. The chain in `main` today:

1. `server/drivers/claude.ts:685,697` — `approvalScope: controlsHost ? "local-computer" : undefined`, so *every* tool on a computer-enabled bot is labelled desktop-control, including plain `Bash` and file edits.
2. `server/index.ts:909` — `allowKey: event.approvalScope ? undefined : approvalKey(...)`, so a scoped card carries no `allowKey`.
3. `src/components/PendingApproval.tsx:119` — `{bot && pending.allowKey && ( … Always allow … )}`, so with no `allowKey` the button is never rendered.

Net effect: turn on computer use and the feature you kept in #442 silently disappears for that bot, even for approvals that have nothing to do with driving the desktop.

This PR narrows the label to tools that actually control the desktop (`mcp__computer*`), so genuine desktop-control approvals stay unrememberable — the original intent — while ordinary tool approvals can be remembered again. No behaviour change for bots without computer use.

Happy to adjust the matching rule if you'd rather scope it differently.

PR #447, opened 2026-08-24
Let Antigravity models control computers (mount the computer MCP)

milind-soni:
## Root cause

Turn assembly gates every computer destination on adapter capabilities (`server/index.ts` ~1516–1557): `computerMcp` enables the cloud box (`integrations.computer`), the Local VM, and the self-hosted VPS mounts; `localComputerMcp` enables the host desktop. `server/drivers/antigravity.ts` declared neither — only `{ sessionModelSwitch, images }` — so any Antigravity model on a bot with a computer destination threw "this model engine cannot control this computer — choose Claude or an ACP engine" (or the VM/VPS variants of that error). The driver simply never mounted `turn.integrations` at all.

## Design

**How the mount works.** `agy` has no per-turn MCP flag and no project-level MCP config. Probed empirically against agy 1.1.19 on this machine:

- the binary's embedded MCP docs list exactly two config locations — global `~/.gemini/config/mcp_config.json` and per-plugin `mcp_config.json` — and no project location;
- `agy mcp list` run from cwds seeded with `.gemini/config/mcp_config.json`, `.gemini/mcp_config.json`, `.gemini/settings.json`, `.agents/mcp_config.json`, and a bare `mcp_config.json` sees none of them;
- `agy mcp add/remove/disable` reads and writes only the global file (shape `{"mcpServers": {name: {command, args, env}}}`).

So project-scoped config provably doesn't work, and the driver falls back to upserting the **global** file: one key, `openmausbot-computer`, written synchronously right before each spawn. The upsert preserves every other byte of the user's config (other servers, unknown top-level keys) and tolerates malformed JSON by starting from a fresh object — the same discipline as `ensureOpenCodeInjectModel` in `server/drivers/acp/opencode-go.ts`. A computer-less turn removes the key, so tools and box/control tokens cannot leak into later turns or the user's own interactive `agy` sessions.

**What gets mounted.** Cloud boxes mount OpenMausBot's REST-to-MCP proxy: `command = process.execPath`, `args = [SPAWNED_PROXIES.computer]` (resolved through `server/proxy-paths.ts`, never relative to the module — the 0.1.24 lesson), `env = ELECTRON_RUN_AS_NODE=1` + `computerProxyEnv()` (box id/token plus the who-is-driving control endpoint) — the same spec `claude.ts` and `codex.ts` build. Local VM and VPS connections arrive as ready-made Cua Driver stdio commands and pass through unchanged.

**Capability gating.**

- `computerMcp: config.fullAuto` — only full-auto instances (the default) advertise a computer. Print mode has no interactive approval channel, and outside `--dangerously-skip-permissions` agy auto-denies tools that would need a prompt (its own mode help: accept-edits = "auto-approve file edits, prompt for commands"; print mode soft-denies what would prompt). A non-fullAuto mount could therefore never fire, so per the review rule it is gated rather than shipped dead.
- `localComputerMcp` intentionally stays unset. `contracts.ts` requires it only "when local MCP calls can reach the human approval channel" and requires full-auto/bypass instances to leave it false — agy print mode cannot deliver per-action approvals in any mode. **Result: Cloud box, Local VM, and VPS now work for Antigravity; "This computer" (host desktop) still refuses, by design.** Host control returns with the native ACP path (agy issue #31).

**Known quirk, pre-existing, not touched here:** `instanceSupportsLocalComputer` in `src/lib/local-computer.ts` treats `computerMcp` as implying local-desktop support in the UI, so a full-auto instance can select "This computer" and only fails at turn time — Antigravity full-auto now joins ACP-fullAuto and Claude-bypass instances in that bucket. The OR is deliberate and test-pinned (commit 848fea8), so aligning it with the server gate is left as a follow-up. The ComputerPanel mirror copy at `src/components/ComputerPanel.tsx:227` renders only for instances where nothing works (non-fullAuto Antigravity), where its current wording is still accurate, so it is unchanged.

**Trade-offs:** the global file is machine-wide, so between the write and the spawned agy reading its config there is an ms-scale window in which a concurrent Antigravity turn with a *different* computer could rewrite the entry; the write is kept immediately before the spawn to minimize it. Between a computer turn and the next computer-less turn the entry persists in the user's global agy config (their own interactive agy would see the computer tools until then). Both are consequences of agy offering no per-session MCP config.

## Verification

- `pnpm typecheck` clean.
- `npx oxlint` on both touched files adds no findings vs origin/main (driver 22 → 22, test file 6 → 6; new code uses zod at the JSON boundary, no typeof narrowing, no assertions).
- New tests in `server/drivers/antigravity.test.ts` (18 total, all pass):
  - upsert preserves the user's other servers and unknown top-level keys, and overwrites our key in place on a later turn;
  - survives malformed JSON by rebuilding a valid file with the mount;
  - a computer-less turn removes only our key, and never creates the file just to remove;
  - the cloud-box spec pins `command/args/env` at `process.execPath` + `SPAWNED_PROXIES.computer` + `ELECTRON_RUN_AS_NODE`/box/control env; Local VM/VPS specs pass through; no computer → null;
  - `computerMcp` exposed on fullAuto instances only; `localComputerMcp` never;
  - a fake-CLI turn asserts the settings file content at spawn time (write is synchronous before the spawn `sendTurn` resolves behind) and the removal on the next computer-less turn.
- **Mutation checks** on the upsert-preserve behavior: (1) replacing the merged `{ ...config.mcpServers }` with an empty object → 3 tests fail; (2) dropping unknown top-level keys on write (`{ mcpServers }` instead of `{ ...config, mcpServers }`) → the preserve test fails. Both mutations reverted; suite green again.
- Full `pnpm test` green (includes the packaged-server smoke; no bare imports outside the bundler's reach — the driver is inlined into the index bundle like claude.ts/codex.ts, which already import the same modules).
- CLI probes (`agy mcp list/add/remove`, embedded docs, project-level location matrix) run live against agy 1.1.19 on macOS.

## Not verified

- **An end-to-end live computer turn.** `agy` on this machine is signed out, so no real print-mode turn can run. "agy discovers the mounted MCP server and calls its tools in a real turn" has not been exercised live — the mount file format matches agy's own `mcp add` output and its embedded schema docs, and the stream-json turn handling is unchanged, but live tool-calling is unproven here.
- Whether accept-edits print mode denies or allows MCP tool calls — unprovable while signed out; the capability is gated to fullAuto so no mount ships where it might never fire.
- Probes ran on agy 1.1.19; the driver was originally validated against 1.1.12. The MCP config location/schema is asserted for 1.1.19.

## Summary by CodeRabbit

* **New Features**
  * Added computer-use MCP support for Antigravity in full-auto and accept-edits modes.
  * Supports computer-use servers hosted in the cloud, local virtual machines, or VPS environments.
  * Reports available computer-use capabilities and automatically manages integration setup.

* **Bug Fixes**
  * Improved recovery from malformed MCP configuration files.
  * Preserves unrelated and concurrent configuration changes.
  * Improved security for temporary configuration files.
  * Prevented overlapping operations from interfering with each other.
  * Ensured configurations are restored or cleaned up after operations complete.
  * Improved cleanup when operations or background processes do not exit normally.

### DIFF
--- a/server/drivers/antigravity.test.ts
+++ b/server/drivers/antigravity.test.ts
@@ -444,4 +444,47 @@ describe("Antigravity computer MCP config", () => {
       rmSync(home, { recursive: true, force: true });
     }
   });
+
+  it("reaps a child that hangs after result, restores the mount, and unblocks the next turn", async () => {
+    ensureDirs();
+    chmodSync(FAKE_CLI, 0o755);
+    const home = mkdtempSync(join(tmpdir(), "omb-agy-mcpreaper-"));
+    const firstDump = join(home, "first.json");
+    const secondDump = join(home, "second.json");
+    const first = await AntigravityDriver.create({
+      instanceId: "agy-mcp-zombie",
+      displayName: undefined,
+      environment: { HOME: home, FAKE_AGY_MCP_DUMP: firstDump, FAKE_AGY_POST_RESULT_DELAY_MS: "10000" },
+      enabled: true,
+      config: { cli: FAKE_CLI, fullAuto: true },
+    });
+    const second = await AntigravityDriver.create({
+      instanceId: "agy-mcp-after-zombie",
+      displayName: undefined,
+      environment: { HOME: home, FAKE_AGY_MCP_DUMP: secondDump },
+      enabled: true,
+      config: { cli: FAKE_CLI, fullAuto: true },
+    });
+    const firstRecorder = recordEvents(first.adapter);
+    const secondRecorder = recordEvents(second.adapter);
+    try {
+      await first.adapter.sendTurn({ threadId: "t-mcp-zombie", text: "first", integrations: boxIntegrations });
+      await firstRecorder.until((event) => event.type === "turn.completed");
+      expect(readConfig(home).mcpServers[ANTIGRAVITY_COMPUTER_MCP_KEY]).toEqual(boxEntry());
+
+      const secondTurn = second.adapter.sendTurn({ threadId: "t-mcp-after-zombie", text: "second" });
+      await secondTurn;
+      await secondRecorder.until((event) => event.type === "turn.completed");
+
+      expect(JSON.parse(readFileSync(firstDump, "utf8")).mcpServers[ANTIGRAVITY_COMPUTER_MCP_KEY]).toEqual(boxEntry());
+      expect(JSON.parse(readFileSync(secondDump, "utf8"))?.mcpServers?.[ANTIGRAVITY_COMPUTER_MCP_KEY]).toBeUndefined();
+      await expect.poll(() => existsSync(configPath(home))).toBe(false);
+    } finally {
+      firstRecorder.stop();
+      secondRecorder.stop();
+      await first.dispose();
+      await second.dispose();
+      rmSync(home, { recursive: true, force: true });
+    }
+  }, 10_000);
 });
--- a/server/drivers/antigravity.test.ts
+++ b/server/drivers/antigravity.test.ts
@@ -4,7 +4,7 @@
 //
 // The fake CLI is a shebang script Windows cannot exec directly;
 // spawnCli resolves it to `node <script>`, so these run everywhere.
-import { chmodSync, existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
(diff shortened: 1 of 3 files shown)


milind-soni:
Addressed the three review findings in 4db9403: the global MCP mount now has a process-lifetime lease and restores the prior config, the effective environment is shared with the spawned CLI, and config permissions are tightened to 0700/0600. Added regression tests for overlapping turns, spawn-time HOME resolution, restoration, and permissions. Full local suite: 1,802 tests passed (12 skipped), plus all package smoke suites.

Issue #505, opened 2026-08-26
Voice: the phone blames ElevenLabs when the configured provider is the system voice


KesleyDavid:
## What happens

With `tts.provider = "system"`, synthesis runs through the Mac's own voices and no ElevenLabs key exists or is needed. The app still explains the state in ElevenLabs terms — it tells the user to add a key that would change nothing.

The voice list and the preview work correctly under this provider. Only the explanation is false.

## What the server sends

`server/tts/index.ts:57-63` returns the provider explicitly:

```ts
export function describeVoice(cfg: AppConfig) {
  return {
    configured: providerConfigured(cfg),
    ready: voiceConfigured(cfg),
    voice: cfg.tts?.voice ?? "",
    provider: voiceProvider(cfg),
  };
}
```

and it documents that `configured` changed meaning along with it:

```ts
/** The system provider needs no credential — it is only ever offered where
 * the platform actually has it, so "configured" means "this engine can
 * speak", not "a key is on file". */
```

That reaches the phone: `server/index.ts:2568` puts it in `/api/config`, which the companion allowlists (`companion/src/routes.ts:56`).

## What iOS does

`ios/Sources/CompanionCore/Models.swift:455-458` has no `provider` field, and `apiKeyConfigured` is still documented as the presence of a credential. The copy in `ios/App/AgentProfileView.swift` is provider-specific:

- `:138` — "ElevenLabs is not configured"
- `:145` — "Add the shared ElevenLabs key in this agent's profile on the computer."
- `:147` — "synthesis still uses the shared ElevenLabs key on your computer."

The worst of the three is `:145`: the user is told to add a key to fix a state that has nothing to do with keys. Under the system provider, "not configured" means the platform reports no voices.

## Suggested fix

Decode `provider` and name the engine actually in use — or drop the brand from the sentence when the provider is not ElevenLabs. The remedy sentence has to change with the meaning of `configured`, not just the label.

## Note from the Android side

The Android companion (#241) mirrors this copy verbatim, deliberately — the port's rule is not to diverge from published iOS behaviour on its own. I'll follow whatever wording you settle on here, on both clients.

Issue #504, opened 2026-08-26
Connected apps: an unreadable credential store makes the phone show every account as disconnected


KesleyDavid:
## What happens

When Composio's credential store cannot be read, `GET /api/connectors/connected` answers `200` with `services: {}` and `credentialStore: "unavailable"`. iOS writes that empty map straight into its view state, so accounts that are still connected render as **Connect**.

Nothing is actually disconnected — the integration on the computer is untouched. The phone shows the opposite of the truth and invites the user to re-authorize an account that never went away.

## The server already separates the two cases

`server/index.ts:4996-5007` distinguishes "we know nothing is connected" from "we could not find out", and says so in place:

```ts
// `credentialStore` is what stops the panel treating this empty list
// as authoritative: an unreadable store means we do not KNOW what is
// connected, which is not the same as knowing nothing is.
```

The web panel honours it — `src/components/PluginsPanel.tsx:60-71` keeps the last confirmed inventory and marks the answer non-authoritative:

```ts
if (response.credentialStore === "unavailable") {
  return { services: readCachedInventory()?.services ?? {}, authoritative: false };
}
```

## What iOS does instead

`ios/Sources/CompanionCore/Models.swift:757-760` does not model the field:

```swift
public struct ConnectorStatuses: Codable, Sendable {
    public var configured: Bool
    public var services: [String: ConnectorStatus]
}
```

and `ios/App/ConnectedAppsView.swift:147-152` replaces state unconditionally:

```swift
guard let response = await session.loadAllConnectorStatuses() else { return }
statuses = response.services
```

`refreshStatuses()` also runs on scene activation (`ConnectedAppsView.swift:61`) and on pull-to-refresh (`:59`), so a transient unreadable store can blank the list on any foregrounding, not only on first load.

## Suggested fix

Decode `credentialStore` and, on `"unavailable"`, keep the last authoritative inventory rather than replacing it — the rule the panel already applies.

## Note from the Android side

I'm building the Android companion (#241) and it has the same gap — `ConnectorStatuses` there carries no `credentialStore` either. I'd rather align it with whatever you decide here than invent a divergent behaviour in one client, so I'm reporting it instead of quietly fixing only Android.

Issue #479, opened 2026-08-25
Failover: the trust ratchet does not survive a restart — a superseded local route leads again


KesleyDavid:
The trust ratchet holds inside one `CandidateRotation` value, but not across a restart. A hand-typed local route that has already been upgraded away from leads the rotation again on the next launch, and the long-lived bearer goes with it.

### The sequence

Every step is reachable through normal use.

1. Someone types a LAN address in Settings. `Session.updateAddress` (`ios/App/Session.swift:535-541`) builds it with `priority: 0` and puts it at the head:

```swift
updated.promote(endpoint)
updated.endpoints = [endpoint] + existingRoutes.filter { $0.url != endpoint.url }
```

2. That route fails. `CandidateRotation.advance` upgrades to a hosted or tailnet route and drops the local one from the rotation — the ratchet working as intended.

3. The protected route says hello, and `rememberWorkingRoute()` (`Session.swift:478-484`) calls `promote(winner)`. But `promote` (`Failover.swift:307-319`) only assigns `activeEndpoint` and **appends** the winner if it was missing:

```swift
public mutating func promote(_ winner: CompanionEndpoint) {
    activeEndpoint = winner
    host = winner.host
    port = winner.port
    if let existing = endpoints,
       !existing.contains(where: { $0.url == winner.url }) {
        endpoints = existing + [winner]
    }
    ...
}
```

It never reorders and never reprioritises, so the `priority: 0` local route is still first in `endpoints`.

4. `orderedEndpoints` (`Failover.swift:196-214`) sorts by priority and original offset only, and appends `activeEndpoint` **only when it is missing** — it is never pulled to the front:

```swift
candidates = candidates.enumerated().sorted {
    $0.element.priority == $1.element.priority
        ? $0.offset < $1.offset
        : $0.element.priority < $1.element.priority
}.map(\.element)
```

5. If the best-effort endpoint refresh 404s, fails, or is cancelled — all deliberately tolerated — that order is what persists.

6. Next launch, `CandidateRotation(endpoints: saved.orderedEndpoints)` (`Session.swift:155`) starts from the cleartext route again, treats it as the explicit local choice, and hands it the stored bearer.

### Why it matters

The ratchet exists so that a connection which has reached a protected route does not walk back onto cleartext carrying a long-lived token. That guarantee currently lives in the lifetime of a `CandidateRotation` value. A process restart — or anything that rebuilds the rotation from the store — resets it.

[...]

Issue #469, opened 2026-08-25
Session.linger(): a stale 25s sleeper can end the next background window early


KesleyDavid:
`Session.linger()` starts a detached 25-second sleeper whose only guard is whether *some* linger is active, not whether it is its own. A second background within the first window ends early as a result.

`ios/App/Session.swift:310-321`:

```swift
func linger() {
    guard streamTask != nil, lingerTask == .invalid else { disconnect(); return }
    lingerTask = UIApplication.shared.beginBackgroundTask(withName: "companion.linger") { [weak self] in
        self?.disconnect()
    }
    Task { [weak self] in
        try? await Task.sleep(for: .seconds(25))
        guard let self, self.lingerTask != .invalid else { return }
        self.disconnect()
    }
}
```

`connect()` calls `endLinger()` (`Session.swift:234`), which sets `lingerTask = .invalid` and ends the UIKit background task — but the detached `Task` above is never cancelled and keeps sleeping.

### The sequence

1. `t = 0` — app backgrounds. `linger()` takes background task **A** and starts sleeper **A**.
2. `t = 24` — user comes back. `connect()` → `endLinger()` sets `lingerTask = .invalid`. Sleeper A is still asleep.
3. `t = 24.5` — user leaves again. `linger()`'s guard sees `lingerTask == .invalid`, so it proceeds and takes background task **B**, starting sleeper **B**.
4. `t = 25` — **sleeper A wakes**, checks `lingerTask != .invalid` — which is true, because it now holds **B** — and calls `disconnect()`.

The second window is cut to half a second instead of 25. The sleeper cannot tell its own window from a later one, because there is no generation token to compare.

### Why it matters

The grace period exists so a turn that finishes just after the user leaves still produces a notification. Glancing at the app and leaving again — which is a natural thing to do while waiting for a bot — is exactly the gesture that truncates the next window, and the closer to the end of a window the glance happens, the shorter the next one gets.

It is a modest bug: the worst case is losing up to ~25 seconds of grace, and only after an out-and-back. Nothing is corrupted and no state is lost.

### Suggested shape of a fix

Give each window a token the sleeper can compare against — an incrementing generation stored on the session, captured by value in the `Task`, and checked alongside `lingerTask` on wake — or hold the `Task` and cancel it in `endLinger()`. Either makes a stale sleeper a no-op.

### How it was found

[...]

PR #513, opened 2026-08-26
An Android companion app

KesleyDavid:
# An Android companion app

Closes the proposal in #241, and supersedes #349 — that branch was `android/core` before the last two parity deltas, and it still carried the three pairing security defects those deltas closed.

A native Kotlin + Jetpack Compose client in `android/`, next to `ios/`. Same
product, same sidecar, same allowlist — so the companion works from either
phone.

```
  phone ──LAN / Tailscale / hosted──▶ companion :8810 ──loopback──▶ harness :8799
                                       token, allowlist, Origin refused
```

`server/` is untouched. `companion/` is untouched. The only thing this adds
outside `android/` is a CI job.

## One PR instead of five

#241 proposed splitting this five ways. You said you would rather test the
whole thing than review it piece by piece, so this is one PR. The commits are
still the passes, each one reviewed before it landed, so `git log` reads as the
five-way split would have.

## Parity target

This tracks the iOS app **as of upstream `667af71`**, not upstream `main` as it moves.
Anything that lands in `ios/` after that point is a later Android version, not
a gap in this one. Five parity deltas were ported to reach that point; the last
one closed three security gaps in pairing, which are the three commits below
that touch `android/core`.

Behaviour mirrors iOS exactly. Visual treatment goes native where Apple has no
counterpart — Material instead of glass and `DisclosureGroup`, a custom `Layout`
instead of a `Form`. The mascot, the Messages-shaped chat, the roster and the
transcript are ports of behaviour, not of painting.

## What the phone can and cannot do

The computer still owns agents, keys, transcripts and computers. That is
enforced on this side too, not just assumed:

- **No credential reaches disk.** The QR credential and the 6-digit code live in
  process memory only; a failed pairing burns them. Only the long-lived token
  persists, in the Keystore-backed store.
- **The route you chose is the route it uses.** Automatic pairing carries hosted
  HTTPS alone; Tailscale is an explicit choice; a local pin authorizes one exact
  LAN origin. The endpoint list the sidecar returns is availability information,
  not new consent — the phone will not move a bearer to an authority you did not
  pick.
- **Provider credentials never come back to the phone**, and the phone is never
  asked for an API key. Revoking stays on the computer.
- **No foreground service.** Notifications arrive while connected; a 25s linger
  window covers ask-and-leave, which is what the iOS `session.linger()` does.

## Tested

**1133 JVM tests, 0 failures** — 366 in `:core`, 767 in `:app` — run with
`:core:cleanTest :app:cleanTestDebugUnitTest` so the suites genuinely re-execute.
(`cleanTest` alone does not clean `:app:testDebugUnitTest`; the task is named
`testDebugUnitTest` in an application module, so `:app:cleanTest` quietly
resolves to nothing.)

Every rule added along the way is pinned by a mutation: the production rule is
replaced with the wider version someone would plausibly write, and the intended
test is confirmed to go red. Deleting a rule only proves some test depends on
it; substituting it proves the scope is right. That distinction found six real
defects that were living behind a green suite.

Hardware:

| Device | API | What it covered |
|---|---|---|
| moto g32 (physical) | 33 | Pairing over real Wi-Fi, mDNS discovery, runtime permissions at the API 33 boundary, notification shade, approvals, release APK |
| Emulator | 26 | `minSdk` floor, release APK install and first run |

Two things worth calling out, because they were found on hardware and not in a
test:

- iOS bug #312 (a three-option approval card recording `deny` when you tap
  "Always allow") **does not** reproduce here — verified on the phone, which
  recorded `allow` with `alwaysAllow: ['ask_bot:<botId>']`.
- The notification shade rendered LTR text with RTL paragraph direction under an
  RTL locale. Proven on the device, then fixed. Compose defaults differ from
  `TextView` here in a way that is easy to miss.

## Screenshots

Both devices, on the release build.

### First run — Android 8 (API 26), clean install

| | |
|---|---|
|  |  |
| The first frame explains before it asks. No permission dialog fires here — that was the point of the last pass. | "Not now" leads somewhere useful rather than a dead end, with Settings reachable and the pairing instruction pointing at **Settings → Phone**. |

### Pairing — API 26

QR first. "Other ways to connect" starts collapsed, and local discovery — with the network permission it needs — begins only when someone opens it. Before that tap there is not a single NSD line in the log; "Looking…" appears after.

### Daily use — moto g32, Android 13 (API 33)

| | |
|---|---|
|  |  |
| Groups, bots and the mascot. This phone was already paired and upgraded in place — it went straight here, with no welcome screen and no repeat permission prompt. | The transcript: tool calls, an approval card, and a markdown table that reads across rather than down. |

### What the phone is not allowed to do

| | |
|---|---|
|  |  |
| Avatar and identity are editable from here; image generation is not, and says why — provider keys cannot be added from a phone. | Settings states what notifications cover **and what they do not**: closed-app push needs a relay that does not exist yet, rather than implying it works. |

## The APK

`./gradlew :app:assembleRelease` writes `app-release-unsigned.apk` when no key
material is configured, which is the artifact for you to sign with your own key:

```sh
apksigner sign --ks <your-keystore>.jks --ks-key-alias <alias> \
  --out app-release.apk app-release-unsigned.apk
```

It is already zip-aligned, so no `zipalign` and no `jarsigner`. `android/README.md`
has the full flow, including how to have Gradle sign directly if you ever want
that, and why R8 is off (kotlinx.serialization reaches 66 generated serializers
reflectively; a minified build installs and opens and then fails on the first
frame off the socket, so turning it on means writing keep rules *and* testing
them against a real pairing).

Signing keys never enter this repository; `.gitignore` refuses `*.jks`,
`*.keystore` and `*.p12` repo-wide.

## Found in your code while porting

Filed rather than worked around, each verified against `origin/main` before
filing:

- **#467** — mDNS sets the cache-flush bit on shared PTR records, which RFC 6762
  §10.2 forbids; two computers running the companion tend to appear as one.
  Fixed in **#471**.
- **#469** — `linger()` has no generation token, so a round trip near the end of
  one window can cut the next one short.
- **#479** — the failover trust ratchet does not survive a restart.
- **#504** — an unreadable credential store makes the phone show connected
  accounts as disconnected. Shared with the published iOS client.
- **#505** — the phone blames ElevenLabs when the configured provider is the
  system voice. Also shared; the Android side follows whatever wording you
  settle on.

## Known limits

- **Closed-app push does not exist.** Notifications arrive while connected, plus
  a 25s linger window for ask-and-leave. A push relay is a separate piece of
  work, and Settings says so rather than implying otherwise.
- **No foreground service**, deliberately.
- **R8 is off**, for the reason above.
- Two screen-invocation lines in the pairing flow are covered by source pins
  rather than by a composition test; the rules they invoke are covered. Closing
  them needs Robolectric plus a harness for the public APK, which did not look
  worth it for this release. Recorded rather than claimed closed.
- `UNRESOLVED` permission state is unreachable from the Android UI, which reads
  authorization synchronously; the rule is pinned in `:core` instead.
- The API-33 charset overload is guarded by a source scan rather than a test,
  because no JVM test can reproduce that crash — Robolectric does not shadow
  `java.*`. The scan stands down on its own once `minSdk` reaches 33.

## Summary by CodeRabbit

* **New Features**
  * Added a complete Android companion app with onboarding, secure pairing via QR codes, deep links, manual addresses, and network discovery.
  * Added chat, task, room, routine, computer, and settings experiences with persistent drafts and notification navigation.
  * Added voice dictation, voice previews, avatar selection and generation, animated mascots, Markdown, transcript cards, reactions, search, and transcript sharing.
  * Added local notifications, connection recovery, cloud desktop access, accessibility support, and light/dark themes.
* **Documentation**
  * Added Android setup, testing, building, signing, and release guidance.
* **Tests**
  * Added extensive coverage for pairing, networking, lifecycle, storage, notifications, media, and UI behavior.


KesleyDavid:
06-boas-vindas-api26
07-home-sem-vinculo-api26
09-outras-formas-api26
01-roster
02-chat
10-perfil-agente
05-ajustes

milind-soni:
Maintainer integration pass is now on this branch.

- Synced the complete Android app over current `main`.
- Fixed the secure-token durability edge case: pairing now fails if the encrypted token commit does not land.
- Replaced custom parsers/readers/caches with Android/Okio/Compose primitives and removed dead permission/preferences/compatibility APIs (net -773 lines in the hardening commit).
- Fixed all six Android lint errors, including notification permission races and API 31/34 guards.
- Preserved managed HTTPS pairing, route consent, protected-route failover, multi-computer storage, sharing, and the rest of the contributor implementation.

Local verification: 1,247 JVM/Robolectric tests, 0 failures; `lintDebug`; debug APK; unsigned release APK. A physical-device cellular/managed-HTTPS pass is still the final real-world validation once an Android phone is connected.

PR #514, opened 2026-08-27
feat(claude): add per-bot tool scoping

### LINKED ISSUE #478: Per-bot tool scoping: no way to withhold a built-in engine tool from one bot
Engine child processes are spawned with `{ ...process.env, PATH: augmentedPath() }`,
and I couldn't find any per-bot env config. So every bot inherits every credential the
app has — `GH_TOKEN`, cloud tokens, whatever happens to be in the environment.

That matters once you run bots with different trust levels. A bot that crawls live
sites or reads an inbox holds the same GitHub token as a bot that writes code. With the
Claude engine that token is often broadly scoped (`repo` covers every repository on the
account, `workflow` covers CI), so a prompt injection from a crawled page reaches a lot
further than the bot needs.

The same gap exists for tools. Every bot gets the engine's full toolset, so a browsing
bot can list and modify account-level cloud routines it has no business touching —
including ones created by unrelated work on the same account. Credentials and tools are
really the same problem: right now a bot cannot be less privileged than the app.

You already apply this thinking on the Box path — `noEnv: true`, *"so API keys cannot
silently appear inside the guest."* Local engines could use the same idea.

## Concrete ask

Optional per-bot overrides in bot settings:

- **env** — a small map merged over `process.env` at spawn. Even just being able to set
  `GH_TOKEN` per bot would cover most of it.
- **tools** — an allowlist (or denylist) so a bot can be handed fewer capabilities than
  the engine offers.

Today the only lever is launching the whole app with a different environment, which
hits every bot at once and is all-or-nothing.

## Environment

OpenMausBot 0.1.34, macOS arm64.

lightcloud00:
## Summary

- add per-instance Claude `tools[]` and `disallowedTools[]` configuration
- narrow and validate both arrays when decoding instance config
- emit the exact Claude CLI flags without changing other provider drivers
- document how to allow a subset or explicitly withhold built-in tools per bot

## Verification

- 56 focused Claude driver tests pass, 1 skipped
- Node 24 / pnpm 10 typecheck passes
- production build passes
- exact diff and working tree are clean

Closes #478.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring Claude’s allowed and disallowed tool scopes per instance.
  - Tool selections and exclusions are applied to each Claude interaction, including the option to disable all built-in tools.
  - Configuration values are validated, trimmed, and deduplicated.

- **Documentation**
  - Added setup guidance covering tool patterns, defaults, and how tool scopes differ from MCP integrations and permissions.

### DIFF
--- a/apps/docs/content/docs/getting-started/configuration.mdx
+++ b/apps/docs/content/docs/getting-started/configuration.mdx
@@ -32,6 +32,26 @@ Local agent chat works without these credentials.
 
 Source and headless runs can set supported environment variables before starting the harness. The app-level settings are preferable for normal packaged use because they validate inputs and keep secret values out of the renderer.
 
+### Per-instance Claude tool scope
+
+Source and headless installations can give separate Claude instances different built-in tool sets in `~/.openmausbot/config.json`:
+
+```json
+{
+  "instances": {
+    "claude-browser": {
+      "driver": "claudeAgent",
+      "config": {
+        "tools": ["Read", "WebFetch", "WebSearch"],
+        "disallowedTools": ["Bash(git *)", "Edit", "Write"]
+      }
+    }
+  }
+}
+```
+
+`tools` selects Claude's available built-ins; an explicit empty array disables every built-in, while omitting it keeps Claude's default set. `disallowedTools` applies Claude tool-name patterns as an additional deny list. These settings do not grant or pre-approve MCP integrations, which remain controlled by the instance's mounted integrations and OpenMausBot permission flow.
+
 <Callout type="warn" title="Do not put secrets in prompts">
   A prompt becomes conversation history and may be sent to an agent provider. Use App Settings or environment configuration for credentials.
 </Callout>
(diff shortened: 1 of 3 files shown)


lightcloud00:
## Exact source handoff complete

[PR #514](https://github.com/milind-soni/OpenMausBot/pull/514) is open and mergeable at exact head `21d36fae6ef22e14851bb0f5b1e2148938ef709e`.

Deterministic admission is green: the full local Node 24 fleet gate, macOS, Ubuntu, Windows, control-plane/workerd, Linux package-smoke, Swift/iOS, and CodeRabbit all passed. CodeRabbit reported no actionable comments. The Vercel authorization status is unrelated to this source-only change.

The upstream merge was attempted once and GitHub denied it because `lightcloud00` has read-only access to `milind-soni/OpenMausBot`. The remaining action is maintainer merge of PR #514; its body closes this issue automatically.

PR #519, opened 2026-08-27
fix(ios): a linger sleeper may not close a window it does not own

### LINKED ISSUE #469: Session.linger(): a stale 25s sleeper can end the next background window early

KesleyDavid:
Fixes #469.

## What happens

`linger()` starts a 25-second sleeper and `endLinger()` does not cancel it. A round trip near the end of one window — leave, come back, leave again — lets the first window's sleeper wake up during the second one and end it early. The stream drops while the user believes they still have the grace period, so an approval or a finished-turn notification arriving in those seconds never lands.

The UIKit expiration handler had the same shape: it captured no window identity, so an expiration belonging to a finished window could disconnect a live one.

## The fix

A generation counter, incremented only when a window actually opens. Both the sleeper and the expiration handler capture it and act only if it is still current and a window is still open. A sleeper from window A wakes during window B, sees a generation that is no longer its own, and returns.

`Session` is `@MainActor`, so there is no suspension point between the guard and `disconnect()` — the check cannot be raced. The old `Task` is deliberately left to finish its sleep rather than being stored and cancelled: with the guard in place it is inert, and keeping a handle would add state to buy nothing but an earlier wake-up.

## About tests

There are none, and I would rather say so than imply otherwise.

`Session.swift` lives in `ios/App/`, and `ios/Package.swift` builds and tests only `ios/Sources/CompanionCore`, so `swift test` cannot reach it. The `xcodebuild` step compiles it but does not exercise it. Moving the lifecycle policy into `CompanionCore` purely to make it testable would be a much larger change than the fix, in a shipped app, and that is your call rather than mine.

The scenario a test would need to drive, if that ever happens: open window A and suspend its sleeper near 25s; call `connect()` to end A; open window B before A's sleeper resumes; resume A and assert the stream and B are still alive; then let B expire on its own schedule and assert only that disconnects.

Found while porting this behaviour to Android, where the same rule is covered by tests because the equivalent code sits in a testable module.

https://claude.ai/code/session_01SkWZ3y7iKQMjH6tHNFBte7

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved background session handling to prevent older expiration events from disconnecting an active session.
  * Increased reliability when transitioning between consecutive background tasks.

### DIFF
--- a/ios/App/Session.swift
+++ b/ios/App/Session.swift
@@ -376,6 +376,9 @@ final class Session: ObservableObject {
     }
 
     private var lingerTask: UIBackgroundTaskIdentifier = .invalid
+    /// Identifies the background window currently allowed to end the stream.
+    /// A sleeper from an earlier window may wake after a newer one has opened.
+    private var lingerGeneration = 0
 
     /// Leaving the screen: keep the stream alive for the grace period iOS
     /// allows (~30 s) rather than cutting it at once, so an approval that
@@ -384,13 +387,22 @@ final class Session: ObservableObject {
     /// the cursor is written down at a known point.
     func linger() {
         guard streamTask != nil, lingerTask == .invalid else { disconnect(); return }
+        lingerGeneration += 1
+        let generation = lingerGeneration
         lingerTask = UIApplication.shared.beginBackgroundTask(withName: "companion.linger") { [weak self] in
             // time is up before our own timer — the system wants us gone now
-            self?.disconnect()
+            guard let self,
+                  self.lingerGeneration == generation,
+                  self.lingerTask != .invalid
+            else { return }
+            self.disconnect()
         }
         Task { [weak self] in
             try? await Task.sleep(for: .seconds(25))
-            guard let self, self.lingerTask != .invalid else { return }
+            guard let self,
+                  self.lingerGeneration == generation,
+                  self.lingerTask != .invalid
+            else { return }
             self.disconnect()
         }
     }


KesleyDavid:
Closing this — #499 was already open for #469 and gets there first. I should have looked before opening a second one; my apologies for the noise.

Having read that diff: cancelling the stored `Task` and guarding on `Task.isCancelled` does close the defect as filed. The sleeper cannot survive `endLinger()`, which is the whole of the sequence in the issue.

One thing I noticed while working on the same code, offered as a question rather than a claim, because I could not check it here — no Swift toolchain on this machine:

The UIKit expiration handler in `linger()` still captures no window identity and calls `disconnect()` unconditionally. My assumption is that `endBackgroundTask(lingerTask)` deregisters it, so a handler belonging to a finished window can never fire and there is nothing to guard. If that assumption is right, #499 is complete and this is nothing. If iOS can still invoke an expiration handler after its task has been ended, then the same stale-window problem exists on that path too, and the fix there would be the same shape as the one you already wrote.

Worth thirty seconds of someone's attention who knows that API better than I do; not worth another PR from me either way.

Found while porting the companion to Android, where the equivalent code is in a testable module and the rule is covered by tests. That port is #513.

PR #526, opened 2026-08-27
feat(composer): cancel a queued 1:1 message with the white X

maxkongerskov:
## What

1:1 queued sends get a white **X**, same as rooms already had for a local draft.

Clicking it drops the pending chip and `DELETE /api/bots/:id/queue/:queueId`, so drain never fires that send. The transcript's "Queued — sends when this turn finishes" row has the same control.

## Demo

Queued message with a white cancel X

## Testing

- `pnpm typecheck`
- `pnpm exec vitest run src/state/store.test.ts server/steer-queue.test.ts` (27 passed)

## Summary by CodeRabbit

* **New Features**
  * Added the ability to cancel queued messages before they are sent.
  * Added cancellation controls for queued messages in bot and group conversations.
  * Cancelled messages are removed from the queue immediately and remaining messages continue processing.

* **Bug Fixes**
  * Improved queue state updates so cancelled messages disappear without waiting for another queue event.


maxkongerskov:
Closing as duplicate of #524.

The queued 1:1 cancel-X (composer + transcript row, `DELETE /api/bots/:id/queue/:queueId`) already ships in that chat-presence PR, and the two PRs touch the same files (`server/index.ts`, `server/steer-queue.ts`, `ChatView.tsx`, `Composer.tsx`, `store.tsx`). Leaving both open would just fight on merge.

Keeping #524 as the single review surface.

PR #549, opened 2026-08-28
fix: remote live stream bubbles + sidebar grouping

matt5115:
## Why

I host OpenMausBot on one Mac and use it from another over Tailscale. Sends hit the host, but the remote window stayed empty until I quit or opened a browser on the host. The SSE socket still *looked* connected.

The server was pinging with SSE comments (`: keepalive`). Browser `EventSource` ignores comments, so a half-dead hop never fired `onerror` and never reconnected. Chat bubbles only arrive on that stream (or after a full hydrate).

## The fix

- Emit a `kind: ping` **data** frame the client can see (keep the comment too for TCP).
- Reconnect with `?since=` when pings go quiet, and on focus / online / visible.
- Return the user message from `POST /api/bots/:id/messages` (and group sends) so the bubble does not wait on SSE.
- Bind Vite on `0.0.0.0`, `changeOrigin` the `/api` proxy, disable proxy timeouts, and disable buffering on `text/event-stream`.
- Disable Node `requestTimeout` / `headersTimeout` on the harness so long-lived SSE is not killed.

## Also in this PR (sidebar / copy)

- **Bot Chats** auto-files bot-to-bot DMs (they were inheriting the sender’s team, e.g. Agents). Delete from the chat menu.
- **Pinned** is its own collapsible section.
- **Channels** renamed **Projects**.
- Drag a section header to reorder (no up/down arrows). Empty sections keep their saved place.
- Copy on bubbles uses a clipboard fallback that works on this local/Electron path.
- ⌘1–6: Chief of Staff, then pinned, then the rest.

Not merged from this seat. No `.env` or credentials in the diff.

## Test plan

- [ ] Idle bot: send from a remote UI; user bubble appears without waiting for the assistant.
- [ ] Kill/stall the SSE hop; client reconnects; missed frames resume or hydrate.
- [ ] Sidebar: drag Projects below Agents, reload, order sticks; start a bot-to-bot DM, it lands under Bot Chats.
- [ ] Copy a bubble on desktop Electron / `http://127.0.0.1`.

## Summary by CodeRabbit

* **New Features**
  * Added collapsible, reorderable sidebar sections with saved layout preferences.
  * Introduced dedicated Bot Chats organization for bot-to-bot conversations.
  * Added Codex Micro slots, with keyboard shortcuts for quick access.
  * Improved live updates with automatic reconnection and keepalive handling.
  * Added reliable copy actions with clipboard fallback support.
* **Bug Fixes**
  * Bot conversations are now consistently placed in Bot Chats.
  * New messages appear immediately after sending.
  * Inspector updates are more reliable during completed or failed turns.
* **UI Improvements**
  * Renamed the Channels section to Projects.
  * Added clearer chat-specific room actions.

### DIFF
--- a/server/comms-visibility.test.ts
+++ b/server/comms-visibility.test.ts
@@ -1,0 +1,41 @@
+import { rmSync } from "node:fs";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import { BOT_CHATS_SECTION } from "../shared/sidebar-layout.ts";
+import { DATA_DIR } from "./config.ts";
+import type { ModelSelection } from "./contracts.ts";
+import { getOrCreateChannel } from "./comms-visibility.ts";
+import { closeMessageDb } from "./message-db.ts";
+import { Store } from "./store.ts";
+
+const selection = (): ModelSelection => ({ instanceId: "claude", model: "fake-model" });
+
+describe("getOrCreateChannel", () => {
+  beforeEach(() => {
+    rmSync(DATA_DIR, { recursive: true, force: true });
+  });
+  afterEach(() => {
+    closeMessageDb();
+    rmSync(DATA_DIR, { recursive: true, force: true });
+  });
+
+  it("files a new bot chat under Bot Chats, not the sender's team section", () => {
+    const store = new Store(selection);
+    const from = store.createBot({ name: "Forge", section: "Agents" });
+    const target = store.createBot({ name: "Quarry", section: "Agents" });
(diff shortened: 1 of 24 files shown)


milind-soni:
Thanks @matt5115 — the live-delivery diagnosis, visible heartbeat, and canonical POST-response ideas were valuable. I brought that work into main through #555 as a focused rewrite with replay-safe snapshots, idempotent retries, task-race protection, and draft recovery, while leaving out the unrelated sidebar changes and unsafe non-loopback dev-server exposure.

PR #571, opened 2026-08-29
Add Avatar Lab for generated mascot bodies

LukichevDsgn:
## What changed

Adds an Avatar Lab on top of the current official OpenMaus mascot-body generator instead of introducing a second avatar engine.

The old 33-file / 36-commit implementation was replaced with one current-main commit touching five files.

## Behavior

- One live preview and static choice cards all render through the same `MausAvatar` / `CursorAvatar` path.
- Body, resting face, and color can be previewed together and saved atomically.
- Randomize only selects visible supported choices and avoids immediately repeating each current value.
- Escape, backdrop click, Cancel, and focus restoration are supported.
- Runtime state changes do not erase unsaved choices while the dialog is open.
- The profile card shows one current-style entry point instead of a second inline body catalog.

## Shape and face consistency

- Uses the official generated body catalog, shared face scale, solved safe zones, gradients, and native animation engine.
- Cursor appears exactly once.
- Moon is not present.
- Hexagon is hidden from new choices, while old stored `hexagon` values remain valid so existing bots are not corrupted.
- Alert glyph morphs, blinking, state motion, and success confetti remain native `CursorAvatar` behavior.

## Verification

- Focused avatar/profile suite: 28 tests passed.
- Client and server TypeScript checks passed.
- Production renderer build passed.
- Electron syntax check: 90 modules.
- Targeted lint and `git diff --check` passed.
- Added-line secret scan: 405 lines passed.
- Full Vitest run: 2,784 passed, 101 skipped, 1 unrelated Windows Electron sandbox test failed.
- The same `electron/browser-closed-shadow.electron.test.mjs` failure reproduces on untouched official main with exit code `2147483651`.

## Live UI check

Checked the current local app at 1000×720: opened profile → Avatar Lab, selected Diamond + Curious + Coral, verified preview/card parity, then cancelled without changing bot data.

## Maintainer follow-up

Vercel authorization was not attempted, and no CodeRabbit command was triggered.

Accepted head: `b7c67da38fe3f2db177995d1b36a9351f9bfe2c4`.

### DIFF
--- a/src/components/AvatarLabDialog.test.ts
+++ b/src/components/AvatarLabDialog.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from "vitest";
+
+import { MAUS_COLOR_NAMES, PICKABLE_STATES } from "@/lib/mascot";
+import { AVATAR_LAB_BODY_IDS, randomizeAvatarLabDraft } from "@/lib/avatar-lab";
+import { MASCOT_BODY_IDS } from "../../shared/mascot-bodies";
+
+describe("Avatar Lab catalog", () => {
+  it("keeps exactly one cursor and omits the retired hexagon and nonexistent moon", () => {
+    expect(AVATAR_LAB_BODY_IDS.filter((id) => id === "cursor")).toHaveLength(1);
+    expect(AVATAR_LAB_BODY_IDS).not.toContain("hexagon");
+    expect(AVATAR_LAB_BODY_IDS).not.toContain("moon");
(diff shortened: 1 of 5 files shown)


LukichevDsgn:
Addressed the CodeRabbit mouth-control finding in the latest commits. `showMouth` and `mouthStroke` now flow through `CanonicalFaceRig`; defaults remain unchanged. Focused avatar tests: 9/9 passed, plus app/server TypeScript checks.

PR #572, opened 2026-08-29
Add a bounded streaming secret-diff scanner

LukichevDsgn:
## What changed

Adds a portable streaming scanner for added lines in staged or branch diffs.

- Detects high-signal provider keys, private-key headers, AWS access keys, and literal credential assignments.
- Ignores exact placeholders, environment references, and expression-shaped credential lookups.
- Parses the diff incrementally with a 16 MiB fail-closed cap.
- Limits retained diagnostics to 20 findings and never prints matched secret values.
- Validates base revisions and uses `--end-of-options` / `--` to prevent Git option injection.
- Exposes `pnpm check:secrets` for staged changes and `pnpm check:secrets -- --base <revision>` for a branch diff.

## Why

This gives contributors a bounded, redacted pre-publish check without uploading source or credentials to another service.

## Verification

- 11 scanner/parser/CLI tests passed.
- The scanner passed against its own staged diff.
- Targeted lint passed.
- Server TypeScript check passed.
- `git diff --check` passed.
- Rebased as one commit on official `main` at `cb0d376`.
- Current head: `a73bc73e5bb79e853862aa2a15f65e9e8e187ee5`.

## Scope

This PR adds the scanner command and tests. It does not add a new GitHub Actions job or consume a third-party review quota. Maintainers can wire the command into CI separately if desired. No UI is changed, so screenshots are not applicable.


LukichevDsgn:
Addressed all CodeRabbit findings in the latest commits: hunk-aware `+++` parsing, a 20-item bounded redacted sample with exact total/omitted counts, and accurate finding-file counts. Secret scanner tests: 10/10 passed; base mode scanned all 506 added lines; app/server TypeScript checks passed.

milind-soni:
The streaming diff parser is the hard part here and it holds up: a fail-closed byte cap, `StringDecoder` so a secret split across chunks still matches, findings that carry file/line/rule and never the matched value, and `--end-of-options` on the user-supplied revision.

What blocks it:

**1. The credential-assignment rule matches shapes, not secrets** (`scripts/check-secrets.mjs:21`). It fires on type annotations and plain reads. Running `scanAddedLine` over main's `server/` and `shared/`: 492 hits across 123 files, none of them a real credential — `token: string;` (`server/box-create-idempotency.ts:47`), `const apiKey = process.argv.includes("--api-key")` (`scripts/verify-claude-auth-settings.ts:9`). Feeding the last 12 merged PR diffs through `scanDiffStream`, three exit 1 (#943, #960, #1103), all false. The fix is to gate on the value: a minimum length plus entropy, and skip bare identifiers, type names, and member expressions.

**2. `npm_[A-Za-z0-9_-]{20,255}` (line 17) matches npm env var names.** `npm_config_update_notifier: "false"` in `server/engine-install.ts:82` is reported as `provider-key`, and the inline marker can't suppress it — `TEST_FIXTURE_PATH` only covers fixture and test paths. Real npm tokens are `npm_` plus 36 chars; excluding the `npm_config_` prefix also works.

**3. The new tests never execute.** `package.json` adds `test:secrets`, but `test` is unchanged and `.github/workflows/ci.yml:50` runs `pnpm test`. Appending `&& pnpm test:secrets` wires the 11 tests in.

**4. CI has not run on this head.** Both runs sit at `action_required` from Sep 1 — I'll approve them. The red Vercel check is fork-deploy authorization and is red on every fork PR, not something you caused. The branch is 508 commits behind main.

Could you rebase, tighten 1 and 2 until `--base origin/main` comes back clean against a few recent merges, wire the tests into `test`, and push?

PR #611, opened 2026-08-31
Add portable OpenMausBot diagnostics

LukichevDsgn:
## What changed

Adds a portable `pnpm doctor` command for checking a local OpenMausBot instance without reading user configuration or credentials.

It reports:

- loopback endpoint validation;
- `/api/health` reachability, app identity, readiness, and PID;
- optional status/capsule identity and source/build digest agreement;
- opt-in SHA-256 hashes for explicitly supplied regular files;
- human-readable, `--json`, and `--quiet` output;
- deterministic exit codes: 0 healthy, 1 warning, 2 failure.

## Safety boundaries

- accepts only HTTP loopback hosts;
- rejects URL credentials;
- never follows redirects;
- uses bounded request timeouts;
- rejects conflicting digest aliases;
- does not print supplied watch paths when they fail;
- does not read `~/.openmausbot`, tokens, or account data;
- validates missing CLI argument values instead of consuming the next flag.

## Verification

Validated after rebasing onto current official `main`:

- 12 focused tests passed
- real smoke check against a running packaged OpenMausBot instance passed
- missing-argument negative control returned exit code 2
- `git diff --check` passed
- staged additions were checked for private emails and common secret-token formats

## Scope

- 1 commit
- 3 files changed
- no UI changes
- no dependency or lockfile changes
- no credentials or private account data included

Current head: `86be30484814d7af3c929e9bbf04ce85b8ab44af`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "packageManager": "pnpm@10.33.0",
   "scripts": {
     "clean": "node scripts/clean.mjs",
+    "doctor": "node scripts/openmaus-doctor.mjs",
     "lint": "oxlint .",
     "check:contrast": "node scripts/check-skin-contrast.mjs",
     "i18n:check": "node scripts/generate-locale.mjs --check",
--- a/scripts/openmaus-doctor.mjs
+++ b/scripts/openmaus-doctor.mjs
@@ -1,0 +1,267 @@
+import { createHash } from "node:crypto";
+import { createReadStream } from "node:fs";
+import { lstat } from "node:fs/promises";
+import { resolve } from "node:path";
+import { pathToFileURL } from "node:url";
+
+const DEFAULT_ENDPOINT = "http://127.0.0.1:8799";
+const STATUS_PATHS = ["/api/status-capsule", "/api/status", "/api/capsule"];
+const DIGEST = /^(?:sha256:)?[a-f0-9]{64}$/i;
+
+/**
+ * Creates one normalized doctor check result.
+ * @param {string} level Check severity.
+ * @param {string} name Check name.
+ * @param {string} message Human-readable check message.
+ * @param {Record<string, unknown>} [details] Additional structured fields.
+ * @returns {Record<string, unknown>} The normalized check result.
+ */
+function check(level, name, message, details = {}) {
+  return { level, name, message, ...details };
+}
+
+/**
+ * Validates and normalizes a local HTTP endpoint.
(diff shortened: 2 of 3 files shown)


milind-soni:
The test file is the strongest part of this. `openmaus-doctor.test.mjs` proves *response effect* rather than TCP reachability — the "local HTTP negative control" binds a real server answering `{app:"not-openmausbot"}` and asserts exit 2 — and the loopback-only endpoint check, `redirect: "error"`, and the "never echo secret-shaped values" test make this safe to hand a user. I ran it: 12/12 pass, clean under main's stricter `oxlint --deny-warnings` gate.

What blocks it:

**1. The status block probes routes that don't exist.** `STATUS_PATHS` (`scripts/openmaus-doctor.mjs:8`) tries `/api/status-capsule`, `/api/status`, `/api/capsule`. A grep over `origin/main` finds no route for any of the three, so all three land on the catch-all 404 at `server/index.ts:14784` and the loop continues. Against a stub of main's real surface the doctor prints `INFO status: status/capsule endpoint unavailable` and exits 0 — meaning `digestPair`/`inspectStatus` (lines 92-120) and three of the twelve tests can never run against a real server. The shape you built to is real: `source_sha256`/`dual_view_sha256` in `server/openmaus-status-capsule.ts`. But it's read from a cache file and folded into a system prompt (`index.ts:1443`), never served over HTTP. Cutting lines 92-120 and 183-198 leaves health + watch, which works today.

**2. Rebase.** The branch sits on cb0d376 (Sep 2), 508 commits back, and conflicts on one line: main changed `"lint": "oxlint ."` to `"oxlint --deny-warnings ."`, directly under where you insert `"doctor"`.

**3. No real CI signal.** The one red check is Vercel, "Authorization required to deploy" — fork deployment authorization, not caused by your change. The Kotlin/Swift/build workflows never ran on this head because it predates them; the rebase will start them.

Could you rebase onto main and drop the status block? If you'd rather keep the capsule check, say so and I'll take the "serve those digests over HTTP" question separately — that's a server decision and shouldn't ride on this PR.

milind-soni:
Triage note: current main now provides pnpm control:omb doctor, backed by the shared MCP health and engine-status checks and the isolated verification launcher. This proposal adds a second doctor implementation and probes status-capsule/status/capsule endpoints that are not part of the current verification contract. Holding this version; any still-useful portable checks should extend the shared doctor surface with current-main fixture coverage rather than add a parallel diagnostics CLI.

PR #612, opened 2026-08-31
Add an isolated Codex permission reviewer

LukichevDsgn:
## What changed

Adds a dedicated Codex permission-review path instead of reusing a normal bot turn.

The reviewer now:

- starts a fresh ephemeral `codex app-server` process for each review;
- receives the review prompt over stdin, never through argv;
- runs from a disposable empty working directory;
- uses `sandbox: "read-only"` and `approvalPolicy: "never"`;
- disables inherited MCP servers and passes no dynamic tools;
- receives explicit developer instructions not to call tools or inspect files;
- constrains the final answer with the existing strict `{ allow, reason }` JSON contract;
- inherits only the small runtime environment allowlist required to launch Codex;
- enforces abort handling, a 60-second process bound, and a 1 MB output bound.

Malformed, multiple, oversized, missing, aborted, or timed-out responses fail closed and leave the human approval card open.

## Why

Permission-review prompts may contain untrusted command and tool text. They must not run inside the original bot session with its integrations, working directory, provider credentials, or conversation state.

## Verification

Validated after rebasing onto current official `main`:

- `server/drivers/codex.test.ts`
- `server/auto-review.test.ts`
- 42 focused tests passed
- server TypeScript check passed
- `git diff --check` passed
- staged additions were checked for private emails and common secret-token formats

## Scope

- 1 commit
- 2 files changed
- no UI changes
- no dependency or lockfile changes
- no credentials or private account data included

Current head: `9153c82db1d1d48f31861d166f07ee901df6ed98`.

### DIFF
--- a/server/drivers/codex.test.ts
+++ b/server/drivers/codex.test.ts
@@ -5,18 +5,70 @@
 //
 // The fake is a shebang script — the same constraint codex.cmd itself
 // hits on Windows. resolveCliSpawn covers both, so these run everywhere.
-import { chmodSync, mkdtempSync, readFileSync } from "node:fs";
+import { chmodSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
 import type { ProviderInstance } from "../contracts.ts";
+import { parseReviewVerdict } from "../auto-review.ts";
+import { PROVIDER_CREDENTIAL_ENV } from "../config.ts";
 import { recordEvents, type EventRecorder } from "../testing/events.ts";
-import { CodexDriver } from "./codex.ts";
+import { CODEX_REVIEW_TIMEOUT_MS, CodexDriver, runCodexPermissionReview } from "./codex.ts";
 import { removeTempDir } from "../testing/cleanup.ts";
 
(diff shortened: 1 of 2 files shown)


milind-soni:
The isolation mechanics are the right shape, and the test that asserts the prompt never lands in `argv` and that none of `PROVIDER_CREDENTIAL_ENV` reaches the child is the part worth keeping from this work in any form.

What blocks it, in order:

**1. The approach was superseded after you opened this.** #1089 (merged Sep 11) removed the app-side approval guards — the unattended downgrade and the model reviewer — in favour of passing the provider's own approval modes straight through. `server/auto-review.ts` and `server/auto-review.test.ts` no longer exist on main. That was a deliberate call, so please hold off on rebasing until we settle whether to revisit it.

**2. The test cannot compile against main.** `server/drivers/codex.test.ts:15` imports `parseReviewVerdict` from `../auto-review.ts`, which is now deleted; typecheck and vitest both fail on rebase. Nothing red here is yours — CI never ran on this branch, every run sits at `action_required` waiting on fork approval.

**3. Nothing on main calls `reviewPermission`.** `server/drivers/codex.ts:901` wires it onto the instance, and the only downstream effect is `server/harness/registry.ts:258` setting `capabilities.approvalReview`. That flag is declared at `src/state/store.tsx:579` and read nowhere else, so the change has no visible behaviour today.

What I am asking for: no push. This needs a decision from me, not more code from you. If you see a gap that provider-native review leaves open — untrusted permission text running inside the bot's own session with its credentials and cwd — say so on this thread and I will reopen the question.

PR #647, opened 2026-08-31
fix(linux): keep updates from breaking Ubuntu installs

### LINKED ISSUE #645: Linux: in-app update renames the AppImage out from under its launcher, and runs dpkg under the live app for .deb
The in-app updater is not gated by platform, so it runs on both Ubuntu artifacts. On each one it does something the platform does not want:

- **AppImage:** the update lands under a *new filename* and the old file is deleted, breaking every launcher that points at it.
- **`.deb`:** the app runs `dpkg -i` itself, behind a polkit root prompt, while the app is still running.

This also contradicts the shipped documentation, which states that automatic updates are macOS/Windows only.

All references are to `main` at `24b58040`.

---

## 1. AppImage: the update renames the file and unlinks the old one

`electron/vendor/electron-updater.cjs:14950-14954` (upstream `AppImageUpdater.doInstall`):

```js
if (path2.basename(installerPath) === existingBaseName || !/\d+\.\d+\.\d+/.test(existingBaseName)) {
  destination = appImageFile;                                                   // in place
} else {
  destination = path2.join(path2.dirname(appImageFile), path2.basename(installerPath));  // rename
}
```

with `unlinkSync(appImageFile)` a few lines above.

`installerPath` comes from the feed, and our published `latest-linux.yml` names the versioned asset:

```yaml
path: OpenMausBot-0.1.43-x86_64.AppImage
```

So the branch taken depends entirely on what the **user** named their copy:

| User downloaded | Existing basename | Result |
|---|---|---|
| `OpenMausBot.AppImage` (README / docs link) | no semver | overwritten in place — fine |
| `OpenMausBot-0.1.43-x86_64.AppImage` (releases page) | has semver | **renamed to the new version, old path deleted** |

The release publishes both names, so the second case is reachable by anyone browsing the releases page rather than following the README button.

**Impact:** every `.desktop` entry, symlink, dock pin, and AppImageLauncher registration pointing at the versioned path breaks. The app disappears from the launcher. The already-running process keeps executing from the deleted inode, so it keeps reporting an update is available and never gets it.

Observed on Ubuntu 24.04.4, updating 0.1.27 → 0.1.43:

```
[info] New version 0.1.43 has been downloaded to ~/.cache/openmausbot-updater/pending/OpenMausBot-0.1.43-x86_64.AppImage
[info] Install on explicit quitAndInstall
[info] Executing: ~/.local/opt/openmausbot-test/OpenMausBot-0.1.43-x86_64.AppImage with args:
```

The `0.1.27` file at that directory no longer existed afterwards, while the `.desktop` entry still referenced it.

---

## 2. `.deb`: the app installs a system package itself

`electron/updater.mjs` has no platform gate — `startUpdater` wires the updater for any packaged build. And the `.deb` ships `resources/package-type` containing `deb` (verified in the installed 0.1.43 package), which selects `DebUpdater` at `electron/vendor/electron-updater.cjs:15909`.

So a `.deb` user gets the banner, and "Restart to update" runs (`DebUpdater.doInstall` → `LinuxUpdater.runCommandWithSudoIfNeeded`, `spawnSync` with `shell: true`):

```
pkexec --disable-internal-agent /bin/bash -c 'dpkg -i ~/.cache/openmausbot-updater/pending/OpenMausBot-X-amd64.deb'
```

Four separate problems:

1. **A polkit root prompt appears out of a chat app.** Cancelling, or having no polkit agent registered, fails the install and surfaces a raw error in the banner.
2. **`dpkg -i` does not resolve dependencies.** Any release that adds a dependency fails and falls back to `apt-get install -f -y` — wrapped in the same sudo helper, so it raises a **second** password prompt, and `-f` may remove the half-configured package.
3. **The install runs before the app quits.** `BaseUpdater.quitAndInstall` calls `install()` synchronously and only then `setImmediate(() => app.quit())`. `/opt/OpenMausBot` is replaced underneath the live process; anything read lazily from `extraResources` (the CUA runtime, `cloudflared`) is read from the new package while the process is still the old one.
4. **A root install from a user-writable path.** The staged `.deb` sits in `~/.cache/openmausbot-updater/pending/` and is handed to `dpkg` as root. Anything running as the user can replace it between download and authorization. Worth weighing in an app that runs agent processes with tool access.

---

## 3. The docs describe different behaviour

`apps/docs/content/docs/troubleshooting/updates.mdx`:

> Automatic updates are available on macOS and Windows. Ubuntu packages are updated by installing a newly downloaded package.

The shipped code attempts an automatic install on both Ubuntu artifacts. Either the doc or the code is wrong; they should be reconciled deliberately.

---

## 4. No CI coverage

`.github/workflows/package-linux.yml:61` reproduces a `.deb` upgrade by invoking `dpkg` **directly as root**. That validates the package, not the updater. Nothing exercises `AppImageUpdater.doInstall`'s destination choice or the `DebUpdater` path, which is why both went unnoticed.

---

## Proposed fix

Split the two artifacts, matching what each ecosystem expects:

- **AppImage — keep one-click updates.** Patch the vendored bundle so the destination is always the running file's path, whatever the user named it. Applied by `scripts/bundle-updater.mjs` at bundle time so a regenerated bundle cannot silently lose it, and asserted on the committed artifact by a test.
- **`.deb` (and rpm/pacman) — download, then hand off.** Detect the package type in `electron/updater.mjs` the same way electron-updater does, and route the install action to the system package handler (`shell.openPath` on the downloaded file) instead of `quitAndInstall`. No `pkexec`, no `dpkg` under a live app, and the documented contract becomes true again.
- Add coordinator tests for the hand-off path and a test pinning the AppImage in-place invariant.

Happy to open a PR with this — I have it implemented and testing locally.

## Environment

- Ubuntu 24.04.4 LTS, x64
- Reproduced going 0.1.27 → 0.1.43 on both the `.deb` and the AppImage

KesleyDavid:
Fixes #645.

The in-app updater ran unchanged on both Ubuntu artifacts, and on each one it did something the platform does not want. This splits the two, matching what each ecosystem expects.

## AppImage — the update stops renaming the file

`AppImageUpdater.doInstall` only overwrites in place when the running file has no version in its name; otherwise it writes the release asset's filename and unlinks the old one. Our feed names assets after the version, so anyone running `OpenMausBot-<version>-x86_64.AppImage` ended up with a new file and **nothing at the old path** — every `.desktop` entry, symlink, dock pin and AppImageLauncher registration pointing at it broke, and the app vanished from the launcher.

`scripts/patch-appimage-updater.mjs` rewrites that branch at bundle time, so the update always lands on the path the user launches, whatever they named it. The patch throws when upstream's shape moves — a regenerated bundle cannot silently lose it.

## `.deb` — the app stops installing system packages

`resources/package-type` selects `DebUpdater`, so **Restart to update** ran:

```
pkexec --disable-internal-agent /bin/bash -c 'dpkg -i ~/.cache/openmausbot-updater/pending/OpenMausBot-X-amd64.deb'
```

A polkit root prompt out of what the user experiences as a chat app; `dpkg -i` resolving no dependencies (with a **second** prompt on the `apt-get install -f -y` fallback); and `/opt/OpenMausBot` replaced while the process is still running, since `quitAndInstall` installs first and quits after.

`electron/updater.mjs` now detects the package type the way electron-updater does (`resources/package-type`, `APPIMAGE` as fallback) and hands the downloaded package to the desktop instead of installing it. The banner says **Install** rather than **Restart to update**, and a new `handed-off` state tells the user where to finish.

This also makes the shipped documentation true again — it already promised that Ubuntu packages are updated by installing a downloaded package.

## Tests

Nothing covered either path: the CI deb upgrade runs `dpkg` directly as root, which validates the package, not the updater.

- `electron/vendor-updater.node-test.mjs` — runs the **shipped** `AppImageUpdater.doInstall` against real files and asserts the update lands on the launched path, the staged download is consumed, and no extra file appears. Plus textual invariants on the committed bundle and fail-closed coverage of the patch itself.
- `electron/updater-coordinator.node-test.mjs` — the hand-off receives the staged paths and ends in `handed-off`; a failed hand-off surfaces as an error instead of spinning; the restart path still calls `quitAndInstall`.
- `scripts/verify-linux-package.mjs` — asserts the `.deb` ships `package-type: deb` and the AppImage ships **no** marker. That marker is what routes the two artifacts; if packaging ever inverts it, this fix dies silently.

## Verification

| Check | Result |
|---|---|
| `node --test electron/*.node-test.mjs` | 69/69 pass |
| `pnpm typecheck` | clean |
| `pnpm lint` | no findings in changed files |
| `pnpm vitest run` | 2753 pass, 4 fail |

The 4 vitest failures are `server/skills.test.ts` (3, symlink handling) and `electron/browser-closed-shadow.electron.test.mjs` (1). They reproduce identically on a pristine `main` with this branch's changes stashed — pre-existing on this host, unrelated to these files.

## What is not covered

I have not run a real packaged update end to end — that needs a published release newer than the installed build. The AppImage install path is covered behaviourally against the shipped bundle; the `.deb` hand-off is covered at the coordinator boundary, with `shell.openPath` itself left to the desktop.

On a stock Ubuntu 24.04 the registered handler for `application/vnd.debian.binary-package` is App Center. If it opens but cannot complete the install, the package is already downloaded and the troubleshooting doc carries the `apt-get` command; when no handler is registered at all, OpenMausBot reveals the file instead.

## Deliberately left alone

`DebUpdater` stays in the vendored bundle — we simply never route to it. Keeping the patch surface minimal, with `verify-linux-package.mjs` guarding the routing.

## Summary by CodeRabbit

* **New Features**
  * Linux package updates now copy an installation command and open a terminal for completion.
  * Update prompts display the installation command and whether the terminal opened successfully.
  * Ubuntu 24.04 guidance now uses `apt-get` to resolve dependencies.
  * AppImage updates preserve the existing filename and launch path.

* **Bug Fixes**
  * Improved guidance when a terminal does not open.
  * Improved Linux update reliability and error handling.
  * Linux updates now support package paths containing special characters.

### DIFF
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -63,6 +63,11 @@ jobs:
           deb=(release/*.deb)
           test "${#deb[@]}" -eq 1
           sudo --preserve-env=CI,RUNNER_TEMP node scripts/smoke-deb-upgrade.mjs "${deb[0]}"
+      - name: Prove an AppImage update lands on the launched path
+        # Installs the current published release over a copy named after a
+        # version, which is the shape that used to orphan the launcher. Runs
+        # against the real feed, so it also covers the differential download.
+        run: pnpm smoke:linux-update
       - name: Configure Chromium sandbox for the unpacked app
         run: |
           sudo chown root:root release/linux-unpacked/chrome-sandbox
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -63,6 +63,15 @@ jobs:
           deb=(release/*.deb)
           test "${#deb[@]}" -eq 1
           sudo --preserve-env=CI,RUNNER_TEMP node scripts/smoke-deb-upgrade.mjs "${deb[0]}"
+      - name: Prove the handed-over Ubuntu command installs
+        # The .deb path ends with a command on the user's clipboard, so the
+        # command is the deliverable. Runs it as root on a clean Ubuntu through
+        # a real shell, and checks that the `dpkg -i` it replaced still fails
+        # there — the reason the app must not run that itself.
+        run: |
+          deb=(release/*.deb)
+          test "${#deb[@]}" -eq 1
+          pnpm smoke:deb-command "${deb[0]}"
       - name: Prove an AppImage update lands on the launched path
         # Installs the current published release over a copy named after a
         # version, which is the shape that used to orphan the launcher. Runs
--- a/apps/docs/content/docs/troubleshooting/updates.mdx
+++ b/apps/docs/content/docs/troubleshooting/updates.mdx
@@ -4,7 +4,14 @@ description: Recover from a stuck download or restart and understand supported u
 icon: RefreshCw
 ---
 
-Automatic updates are available on macOS and Windows. Ubuntu packages are updated by installing a newly downloaded package.
+OpenMausBot checks for updates in the background and offers the download in a card at the bottom of the window. How the download is applied depends on what you installed.
+
+| Build | What the button does |
+|---|---|
+| macOS, Windows, Linux AppImage | **Restart to update** — the app replaces itself and reopens. An AppImage is always rewritten at the path you launch, so your launcher, symlink or dock pin keeps working. |
+| Ubuntu `.deb` (and rpm/pacman) | **Install** — the downloaded package opens in your system package manager and you finish there. The app never installs a system package itself, so it never asks for your administrator password. |
+
+Reopen OpenMausBot after the package manager finishes to be running the new version.
 
 ## The download keeps spinning
 
@@ -13,6 +20,18 @@ Leave the app open long enough for one complete attempt, then quit and reopen it
 ## Restart to update does nothing
 
 Quit OpenMausBot completely and launch it again. If the version did not change, install the latest release manually. Your conversations and settings live outside the application bundle and are not removed by replacing the app.
+
+## The package manager did not open, or could not install
+
+On Ubuntu the **Install** button hands the downloaded `.deb` to whatever your desktop has registered for Debian packages — App Center on a stock Ubuntu 24.04. If nothing is registered, OpenMausBot opens the containing folder instead.
+
+Either way the package is already downloaded, so you can always finish from a terminal:
+
+```bash
+sudo apt-get install -y ~/.cache/openmausbot-updater/pending/OpenMausBot-*-amd64.deb
+```
+
+Use `apt-get`, not `dpkg -i` — only `apt-get` resolves dependencies.
 
 ## Verify the installed version
 
--- a/apps/docs/content/docs/troubleshooting/updates.mdx
+++ b/apps/docs/content/docs/troubleshooting/updates.mdx
@@ -9,9 +9,15 @@ OpenMausBot checks for updates in the background and offers the download in a ca
 | Build | What the button does |
 |---|---|
 | macOS, Windows, Linux AppImage | **Restart to update** — the app replaces itself and reopens. An AppImage is always rewritten at the path you launch, so your launcher, symlink or dock pin keeps working. |
-| Ubuntu `.deb` (and rpm/pacman) | **Install** — the downloaded package opens in your system package manager and you finish there. The app never installs a system package itself, so it never asks for your administrator password. |
+| Ubuntu `.deb` (and rpm/pacman) | **Install** — OpenMausBot downloads and verifies the package, copies the install command, and opens a terminal for you to paste it into. The app never installs a system package itself, so it never asks for your administrator password. |
 
-Reopen OpenMausBot after the package manager finishes to be running the new version.
+Ubuntu packages are installed from the releases repository by hand, so the app hands you the exact line rather than guessing at a package handler:
+
+```bash
+sudo apt-get install -y '/home/you/.cache/openmausbot-updater/pending/OpenMausBot-<version>-amd64.deb'
+```
+
+Reopen OpenMausBot once it finishes to be running the new version.
 
 ## The download keeps spinning
 
@@ -21,17 +27,15 @@ Leave the app open long enough for one complete attempt, then quit and reopen it
 
 Quit OpenMausBot completely and launch it again. If the version did not change, install the latest release manually. Your conversations and settings live outside the application bundle and are not removed by replacing the app.
 
-## The package manager did not open, or could not install
+## No terminal opened
 
-On Ubuntu the **Install** button hands the downloaded `.deb` to whatever your desktop has registered for Debian packages — App Center on a stock Ubuntu 24.04. If nothing is registered, OpenMausBot opens the containing folder instead.
-
-Either way the package is already downloaded, so you can always finish from a terminal:
+The install command is on your clipboard either way. Open a terminal yourself and paste it. If the clipboard was overwritten, the package is already downloaded and you can run:
 
 ```bash
 sudo apt-get install -y ~/.cache/openmausbot-updater/pending/OpenMausBot-*-amd64.deb
 ```
 
-Use `apt-get`, not `dpkg -i` — only `apt-get` resolves dependencies.
+Use `apt-get`, not `dpkg -i` — only `apt-get` resolves dependencies, and on Ubuntu 24.04 ours are satisfied through virtual packages that `dpkg` alone will not follow.
 
 ## Verify the installed version
 
--- a/electron/fixtures/updater-handoff.cjs
+++ b/electron/fixtures/updater-handoff.cjs
@@ -1,0 +1,81 @@
+"use strict";
+
+// Proves the two steps of the Ubuntu hand-off that only exist at runtime: the
+// command really reaches the system clipboard, and a terminal is really
+// spawned. The unit tests inject a fake runner and never spawn anything, so
+// this runs the production path with nothing stubbed — a recording script
+// placed on PATH stands in for the terminal emulator, and is discovered and
+// executed by the real child_process.
+
+const { chmodSync, existsSync, mkdtempSync, rmSync, writeFileSync } = require("node:fs");
+const { tmpdir } = require("node:os");
+const { join } = require("node:path");
+const { app, clipboard } = require("electron");
+
+if (process.platform === "linux") app.commandLine.appendSwitch("no-sandbox");
+
+const STAGED = "/home/tester/.cache/openmausbot-updater/pending/o'brien pkg.deb";
+
+function say(marker) {
+  process.stdout.write(`${marker}\n`);
+}
+
+async function main() {
+  const { handOffDownloadedPackage } = await import("../updater.mjs");
+  const { packageInstallCommand } = await import("../package-install-command.mjs");
+  const handOff = handOffDownloadedPackage("deb");
+  const expected = packageInstallCommand("deb", STAGED);
+
+  const workspace = mkdtempSync(join(tmpdir(), "omb-handoff-fixture-"));
+  const realPath = process.env.PATH;
+  try {
+    // A terminal emulator that records being launched. openBlankTerminal tries
+    // x-terminal-emulator first on Linux, and resolves it through PATH.
+    const receipt = join(workspace, "launched");
+    const terminal = join(workspace, "x-terminal-emulator");
+    writeFileSync(terminal, `#!/bin/sh\necho "$@" > ${JSON.stringify(receipt)}\n`);
+    chmodSync(terminal, 0o755);
+
+    clipboard.writeText("something else entirely");
+    process.env.PATH = workspace;
+    const result = await handOff([STAGED]);
+
+    if (clipboard.readText() === expected) say("clipboard-holds-the-install-command");
+    else say(`clipboard-mismatch:${JSON.stringify(clipboard.readText())}`);
+
+    if (existsSync(receipt)) say("terminal-really-launched");
+    if (result.terminalOpened === true) say("hand-off-reported-the-terminal");
+    if (result.command === expected) say("hand-off-returned-the-command");
+
+    // Nothing on PATH: the terminal cannot open, but the command must still be
+    // on the clipboard — the docs tell the user to paste it themselves.
+    const empty = mkdtempSync(join(tmpdir(), "omb-handoff-empty-"));
+    clipboard.writeText("something else entirely");
+    process.env.PATH = empty;
+    const withoutTerminal = await handOff([STAGED]);
+    if (withoutTerminal.terminalOpened === false) say("no-terminal-is-reported-honestly");
+    if (clipboard.readText() === expected) say("clipboard-written-even-without-a-terminal");
+    rmSync(empty, { recursive: true, force: true });
+
+    // A download that vanished must be reported, not turned into a command
+    // that installs nothing.
+    process.env.PATH = realPath;
+    await handOff([]).then(
+      () => say("missing-download-was-not-reported"),
+      (error) => {
+        if (/no longer available/.test(error.message)) say("missing-download-is-reported");
+      },
+    );
+  } finally {
+    process.env.PATH = realPath;
+    rmSync(workspace, { recursive: true, force: true });
+  }
+
+  say("fixture-complete");
+  app.exit(0);
+}
+
+app.whenReady().then(main).catch((error) => {
+  process.stdout.write(`fixture-failed:${error.stack ?? error}\n`);
+  app.exit(1);
+});
--- a/electron/fixtures/updater-handoff.cjs
+++ b/electron/fixtures/updater-handoff.cjs
@@ -14,8 +14,6 @@ const { app, clipboard } = require("electron");
 
 if (process.platform === "linux") app.commandLine.appendSwitch("no-sandbox");
 
-const STAGED = "/home/tester/.cache/openmausbot-updater/pending/o'brien pkg.deb";
-
 function say(marker) {
   process.stdout.write(`${marker}\n`);
 }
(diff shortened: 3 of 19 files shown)


KesleyDavid:
Pushed a follow-up that changes the `.deb` hand-off, and verified the whole thing against real packages.

**The App Center delegation was wrong.** OpenMausBot is installed from the releases repository by hand, not from a store, so there is nothing to delegate to. On a stock Ubuntu 24.04 the registered handler for `application/vnd.debian.binary-package` is App Center, which may not install a local `.deb` at all — and `shell.openPath` resolves as soon as the handler *launches*, so a dead end would still have reported success.

It now does what the app already does for engine installs: copies the exact command and opens a blank terminal via `openBlankTerminal`, with the command shown on the card. Nothing is executed for the user, so no renderer-reachable text becomes a process argument.

`electron/package-install-command.mjs` isolates the one string this path delivers, so it can be tested: `apt-get` rather than `dpkg -i`, and quoting that survives a shell round-trip for paths with spaces, apostrophes, semicolons or command substitution.

## Verified against real artifacts

Built both packages locally (`pnpm package:linux:offline`) and checked what actually ships:

- `node scripts/verify-linux-package.mjs` → **OK** on the real `0.1.44` `.deb` and AppImage, including the new `package-type` assertions.
- Extracted `electron/vendor/electron-updater.cjs` from **both** shipped `app.asar` files: the rename branch is gone from each. The packaged AppImage cannot rename itself out from under its launcher.
- `package-type` is `deb` in the package and **absent** from the AppImage, so the two artifacts route to different install paths.
- Ran the command the app generates, in simulate mode, on Ubuntu 24.04:
  ```
  sudo apt-get install -y '/…/release/OpenMausBot-0.1.44-amd64.deb'
  → 1 to upgrade, 0 to newly install, 0 to remove
  → Inst openmausbot [0.1.43] (0.1.44 local-deb [amd64])
  ```
- `pnpm build:updater` regenerates the vendored bundle **byte-identical** to the committed one, so the patch is reproducible.

One thing the packaging log makes worth keeping the `verify-linux-package.mjs` assertion for:

```
• building target=AppImage …
• building target=deb …
• adding autoupdate files for: deb  resourceDir=release/linux-unpacked/resources
```

`package-type` is written into the **shared** unpacked resources, after the AppImage is sealed. The AppImage is clean only because it is built first. If target order ever changes, the AppImage inherits `deb`, routes to the hand-off, and this fix silently stops applying — which is exactly what that assertion catches.

## Still not covered

I have not run a released-to-released update, which needs a published release newer than an installed build. The AppImage install is covered behaviourally against the shipped bundle; the `.deb` command is covered by executing it against a real package.

Unrelated observation while extracting the asar: `electron/*.node-test.mjs` files are packaged into the app. Pre-existing, and not touched here.

KesleyDavid:
Closed the gap I had flagged as not covered: the update cycle is now proven end to end, against the real feed.

`scripts/smoke-linux-update.mjs` loads the updater **out of the packaged `app.asar`**, points it at the real `openmausbot-releases` feed, and installs the published release over a copy named `OpenMausBot-0.0.1-x86_64.AppImage` — a version in the filename, with a `.desktop` pinned to it. That is the exact shape that orphaned launchers.

The trick that makes it runnable without a newer release: the current version comes from the app adapter, not the file, so the harness claims `0.0.1` and the published release becomes the upgrade. It has to be set before the updater is constructed, since the version is parsed once in the constructor.

```
[smoke-linux-update] feed offers 0.1.43
    Downloading update from OpenMausBot-0.1.43-x86_64.AppImage, …
    Full: 159,810.07 KB, To download: 1,984.85 KB (1%)
[smoke-linux-update] files left in the install directory: OpenMausBot-0.0.1-x86_64.AppImage
  ✔ the update stayed on the launched path
  ✔ no second AppImage appeared
  ✔ the launcher still points at a real file
  ✔ the file now holds the published build
  ✔ the relaunch uses the launched path
```

It exercises the differential download too — 1% of the payload, the same path in the original report.

## The failing case is part of the coverage

A smoke that cannot fail proves nothing, so `--updater=` points the same harness at another bundle. Against the **published 0.1.43**, whose bundle is unpatched:

```
[smoke-linux-update] files left in the install directory: OpenMausBot-0.1.43-x86_64.AppImage
  ✖ the update stayed on the launched path
  ✔ no second AppImage appeared
  ✖ the launcher still points at a real file
  ✖ the file now holds the published build
  ✖ the relaunch uses the launched path
[smoke-linux-update] 4 check(s) failed
```

The launched file is gone and a differently-named one took its place — the original report, reproduced by the test that guards against it.

Wired into `package-linux.yml` after packaging, so every Linux package build proves it.

## Coverage now

| Path | Proven by |
|---|---|
| AppImage, full cycle | This smoke, real feed, real download, real install |
| AppImage, install step | The shipped `doInstall` against real files, in `electron/vendor-updater.node-test.mjs` |
| The two artifacts route apart | `verify-linux-package.mjs` on the real `.deb` and AppImage |
| The `.deb` command installs | Executed in simulate mode against a real package |
| The `.deb` command's quoting | Shell round-trip over paths with spaces, apostrophes, `;`, `$( )` |
| The card's state machine | `updater-coordinator.node-test.mjs` |

What is still unproven: the `.deb` cycle end to end, which would mean installing a package as root inside the test. The command it hands over is executed against a real package, and the app is no longer the thing that installs — so the untested remainder is the clipboard write and the terminal launch, both already covered for engine installs.

KesleyDavid:
The `.deb` cycle is no longer the untested remainder. Both Linux paths now have a discriminating end-to-end proof.

## The handed-over command, run as root

`scripts/smoke-deb-command.mjs` executes the command the app copies, as root on a clean `ubuntu:24.04`, through a real shell, exactly as pasted. The package is staged at a path with a space **and** an apostrophe, so the quoting is exercised rather than assumed:

```
command under test:
    sudo apt-get install -y '/root/pending dir/o'\''brien/OpenMausBot.deb'

  ✔ the package installed and configured
  ✔ the app binary is in place
  ✔ `dpkg -i` alone fails where the handed-over command works
```

That third check is the control, and it is the concrete evidence for §2 of #645: `dpkg -i` — what `DebUpdater` ran in-app — **fails on that same clean image**, because it resolves no dependencies. A user who clicked the old button on a release that added one was left with a broken package. It now fails CI if that ever stops being true.

## The routing had no test at all

`linuxPackageType` decided everything and was covered by nothing. It moved next to the command builder so it could be tested: a marker reading `deb` selects the hand-off, an absent marker with `APPIMAGE` set selects the in-place restart, and the marker wins over the env probe the way electron-updater orders them.

Writing those tests surfaced a real one. A missing `resourcesPath` reached the AppImage fallback by *throwing into the catch* rather than deciding anything — and a `.deb` taking that route would go back to `quitAndInstall` and the in-app `dpkg`. `process.resourcesPath` is always defined under Electron, so this was latent, but the failure direction was wrong. It answers explicitly now.

## Where the coverage stands

| Path | Proven by | Control that fails |
|---|---|---|
| AppImage, full cycle | Real feed, real differential download, real install | The published unpatched bundle fails 4 of 5 checks |
| AppImage, install step | Shipped `doInstall` against real files | Patch throws if upstream's shape moves |
| `.deb` command | Executed as root on a clean Ubuntu | `dpkg -i` fails on the same image |
| `.deb` command quoting | Shell round-trip: spaces, `'`, `;`, `$( )` | — |
| Routing between artifacts | `linuxPackageType` unit tests | `verify-linux-package.mjs` on real artifacts |
| Card state machine | `updater-coordinator.node-test.mjs` | `quitAndInstall` asserts if reached in hand-off |

What is left: the clipboard write and the terminal launch, which are `clipboard.writeText` and the already-tested `openBlankTerminal` this app uses for engine installs. Everything specific to updates is covered.

`pnpm smoke:deb-command` and `pnpm smoke:linux-update` run both locally; both are wired into `package-linux.yml`.

KesleyDavid:
The last two items I had listed as "argued, not shown" are now covered, and both were checked by mutation.

I had waved these off as already tested via the engine-install path. That was too generous: `terminal-launch.test.mjs` injects a fake runner, so it proves the branching and never spawns anything, and the clipboard write had no test at all.

`electron/fixtures/updater-handoff.cjs` runs the production hand-off inside a real Electron process with nothing stubbed, following the `*.electron.test.mjs` pattern already here:

- A recording script placed on `PATH` stands in for the terminal emulator, so `openBlankTerminal` resolves and spawns it through the real `child_process`.
- The command is read back off the real X11 clipboard, after seeding it with a sentinel so a no-op write cannot pass.
- With nothing on `PATH`: the card reports no terminal opened **and** the clipboard still holds the command — what the troubleshooting doc tells the user to rely on.
- A download that vanished is reported, not turned into a command that installs nothing.

## Both assertions fail when the behaviour is removed

```
- clipboard.writeText(command)
  → clipboard-mismatch:"something else entirely"
    expected to contain 'clipboard-holds-the-install-command'

- terminalOpened: await openBlankTerminal()  →  terminalOpened: false
  → no-terminal-is-reported-honestly
    expected to contain 'terminal-really-launched'
```

Each mutation is caught by the assertion that claims it, not by an unrelated one, and the green returns when reverted. Worth doing: the vitest run finished in about a second, which was fast enough for a spawned Electron process to be suspicious — the mutations are what showed the test was real rather than accidentally green.

## Coverage

Everything in the update path now has a case that fails when the rule is broken:

| Behaviour | Proven by | Fails when |
|---|---|---|
| AppImage update stays on the launched path | Real feed, real differential download, real install | Unpatched published bundle: 4 of 5 checks fail |
| The shipped `doInstall` overwrites in place | Packaged bundle against real files | The patch throws if upstream's shape moves |
| The handed-over command installs | Root, clean Ubuntu, in a container | `dpkg -i` fails on the same image |
| Its quoting survives a shell | Round-trip: spaces, `'`, `;`, `$( )` | — |
| The command reaches the clipboard | Real Electron, real X11 clipboard | Mutation: write removed |
| A terminal is really launched | Recording script resolved through `PATH` | Mutation: launch replaced with `false` |
| The artifacts route apart | `linuxPackageType` units + `verify-linux-package.mjs` | Marker in both artifacts is rejected |
| The card never quits on a system package | `updater-coordinator.node-test.mjs` | `quitAndInstall` asserts if reached |

KesleyDavid:
Two things, one of them a correction to what I claimed above.

## Correction: the new CI steps are not running yet

I said both smokes were "wired into `package-linux.yml`" and pointed at a green `package + smoke (Ubuntu 24.04 x64)`. The wiring is in the diff, but it did not execute. The steps this run actually performed are exactly `main`'s workflow:

```
… Reproduce and verify an in-place DEB upgrade
… Configure Chromium sandbox for the unpacked app
… Launch packaged app and verify lifecycle
```

Neither `Prove the handed-over Ubuntu command installs` nor `Prove an AppImage update lands on the launched path` is there. GitHub runs the base repository's workflow for a fork pull request, so workflow changes proposed from a fork stay inert until merged.

So that green means the **existing** Linux package checks pass against these changes — worth having, but not what I said it was. Both smokes pass locally, and they will start running in CI on merge. Worth a look on your side, since I cannot verify them in this PR.

## A data-loss bug in the same function

Reading `doInstall` again for the ordering, the removal happens before anything is validated and before the replacement exists:

```js
unlinkSync(appImageFile);
const installerPath = this.installerPath;
if (installerPath == null) { dispatchError(…); return false; }   // app already deleted
execFileSync("mv", ["-f", installerPath, destination]);
```

Anything failing past that first line leaves the user with **no application at all** — including that guard, which returns having already deleted it. The staged download also sits in the updater cache, which can be on another filesystem, so `mv` may fall back to copy-then-unlink and write into the destination while it is still a running executable.

The patch now lands the update in two steps: `mv` into a sibling of the destination absorbs any cross-filesystem copy, then `rename(2)` within one directory replaces the old file atomically and works over a running binary. Nothing is removed until the replacement is complete.

Covered by a failing install — a staged path that does not exist, standing in for a full disk or a severed mount — asserting the launched AppImage is left byte-for-byte intact. Restoring the unlink fails both that test and the assertion on the committed bundle:

```
✖ the install stages the replacement instead of deleting first
✖ the running AppImage is never removed before its replacement is in place
```

Rebuilt and re-ran everything after the change: `verify-linux-package.mjs` OK on the new artifacts, and the real-feed smoke still updates `0.0.1 → 0.1.43` in place with no residue in the directory. 78 unit tests pass.

KesleyDavid:
One consequence worth deciding on before this ships, because it affects release notes rather than code.

**The update to the first patched release is performed by the unpatched updater.** `startUpdater` runs in the installed app, so the build that downloads and installs a release is the one already on disk. A user sitting on 0.1.43 gets 0.1.43's `doInstall`, whatever the new release contains.

The control in `smoke-linux-update.mjs` already shows what that does. Run against the published 0.1.43 bundle:

```
files left in the install directory: OpenMausBot-0.1.43-x86_64.AppImage
  ✖ the update stayed on the launched path
  ✖ the launcher still points at a real file
```

So every AppImage user whose file carries a version in its name will have their launcher orphaned **once more**, on the update into the first patched release, and be safe from then on. Users who took the README link (`OpenMausBot.AppImage`, no version) are unaffected throughout — that name never triggered the rename.

No code in this PR can prevent that, since the old code runs the transition. What can help is saying so in the release notes for whichever version lands this: AppImage users may need to re-point a `.desktop` entry, symlink or dock pin once, and the file will be next to the old path under the new version's name.

The same generational lag applies to `.deb`: a user on an unpatched build gets one more `pkexec` prompt when updating into the patched release. That path at least completes when no new dependency is involved, so it is less disruptive.

Happy to draft the release-note wording if useful.

---

CI on the latest commit: `package + smoke (Ubuntu 24.04 x64)` and `typecheck + test (ubuntu-latest)` both green. As noted above, the two smoke steps this PR adds still do not execute here — the base repository's workflow is what runs for a fork PR — so they need a look after merge.

KesleyDavid:
Closed the three review nits.

- Inherited names (`toString` / `constructor` / `__proto__`) are no longer treated as package types.
- A staged path that no longer exists is rejected, so the card asks for a re-download instead of handing over a command that installs nothing.
- The installing state now says **Opening a terminal…**, which is what the hand-off actually does.

PR #653, opened 2026-08-31
Route configured Codex models through NVIDIA and OpenRouter

LukichevDsgn:
## What changed

Adds two explicitly configured Codex app-server provider routes:

- `nvidia::z-ai/glm-5.2` through NVIDIA NIM.
- `openrouter::z-ai/glm-5.2` through OpenRouter.

A row is exposed only when its corresponding instance environment key is configured:

- `OPENMAUSBOT_NVIDIA_API_KEY`
- `OPENMAUSBOT_OPENROUTER_API_KEY`

For each turn, the driver removes both provider keys from the inherited child environment, restores only the selected provider key, and passes only provider metadata in argv. Key values never enter argv.

## Scope

This PR now contains only provider routing and its focused test. The duplicated permission-review implementation was removed; that subsystem remains isolated in #612.

No settings UI, credential store, permission reviewer, or unrelated Codex refactor is included.

## Verification

- Focused Codex driver suite: 31 passed.
- Client and server TypeScript checks passed.
- Production Vite build passed.
- Electron syntax check passed across 90 modules.
- Streaming secret-diff scan passed across 122 added lines.
- Rebuilt from official main `6dd974c`.
- Accepted head: `86fbe65667dc717227efba8990235787ae3c783a`.

Tests verify that:

- only the selected provider key reaches the child environment;
- the other provider key is absent;
- neither key value appears in argv;
- both configured model rows are exposed.

## External checks

Maintainer workflow approval and Vercel authorization remain outside this contribution. Neither was triggered or authorized by this PR update.

## Summary by CodeRabbit

* **New Features**
  * Added support for connecting directly to NVIDIA NIM and OpenRouter models.
  * Available models now reflect configured provider access.
  * Requests use the selected provider and securely apply its credentials.
  * Direct provider connections report authenticated and metered status.

* **Bug Fixes**
  * Prevented credentials for unselected providers from being sent or exposed in request arguments.

### DIFF
--- a/server/drivers/codex.test.ts
+++ b/server/drivers/codex.test.ts
@@ -329,6 +329,43 @@ describe("CodexDriver turns (fake app-server)", () => {
     expect(seen.env.OPENMAUSBOT_LOCAL_UNSLOTH_API_KEY).toBe("unsloth-secret");
   });
 
+  it("routes direct NVIDIA and OpenRouter models with only the selected key", async () => {
+    const environment = {
+      OPENMAUSBOT_NVIDIA_API_KEY: "nvidia-secret",
+      OPENMAUSBOT_OPENROUTER_API_KEY: "openrouter-secret",
+    };
+    await create({ environment });
+    expect(instance.models.options.map((option) => option.id)).toEqual(
+      expect.arrayContaining(["nvidia::z-ai/glm-5.2", "openrouter::z-ai/glm-5.2"]),
+    );
+    const nvidiaDump = join(scratch, "nvidia.json");
+    process.env.FAKE_CODEX_DUMP = nvidiaDump;
+    await instance.adapter.sendTurn({ threadId: "t-nvidia", text: "route", model: "nvidia::z-ai/glm-5.2" });
+    await recorder.until((event) => event.type === "turn.completed");
+    const nvidia = JSON.parse(readFileSync(nvidiaDump, "utf8"));
+    expect(nvidia.env.OPENMAUSBOT_NVIDIA_API_KEY).toBe("nvidia-secret");
+    expect(nvidia.env.OPENMAUSBOT_OPENROUTER_API_KEY).toBeUndefined();
+    expect(nvidia.argv).toContain('model_providers.nvidia.base_url="https://integrate.api.nvidia.com/v1"');
+    expect(nvidia.argv).toContain('model_providers.nvidia.env_key="OPENMAUSBOT_NVIDIA_API_KEY"');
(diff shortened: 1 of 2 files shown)


milind-soni:
The environment discipline is the hard part and you got it right: the key rides in the child env, argv carries only the `env_key` name, and the unselected provider's key is deleted before launch. The test proves all three.

Three things block it.

**1. The catalog probe loses PATH.** `server/drivers/codex.ts:179` replaces `const catalogEnv = childEnv()` with `{ ...input.environment }`. `readCodexModelCatalog` spawns the CLI (`codex-catalog.ts:71`), and Node's `env` option replaces the environment instead of merging, so that child gets no PATH and no HOME. Spawning the test fake with the PR's env exits 127; with `childEnv()` it stays up. Tests stay green because `child.on("error"/"close", () => finish(null))` swallows it and falls back to `STATIC_CODEX_MODELS` — so every user quietly loses their real catalog. Keep `catalogEnv = childEnv()` and read the two keys off `input.environment` separately.

**2. `directCodexProviderArgs` (codex.ts:95) emits the `-c model_providers.<id>.*` overrides even when no key is configured** — the key is consulted at line 103 only for the env. Anyone with their own `[model_providers.openrouter]` in `~/.codex/config.toml`, which `readCodexModelCatalog` already surfaces as `openrouter::<model>` rows, gets their `base_url`/`env_key` overwritten and pointed at an unset `OPENMAUSBOT_OPENROUTER_API_KEY`. Return `[]` when the key is absent.

**3. `snapshot()` at codex.ts:693 reports `authenticated: true` whenever either key exists**, before `codex login status` runs. `EngineSetup.tsx:31` reads that as "no setup needed", so a user who never signed in to ChatGPT loses the sign-in card for the official GPT-5.6 rows too. It also drops `update` (the Astra nudge) and flips `billing` to `metered` engine-wide. snapshot has no model in scope, so it cannot stand in for a per-model route.

One question before you spend more time: `server/drivers/local-inject.ts` already does this shape generically — `LOCAL_HOSTS` rows carry `baseUrl`/`apiKeyEnv` and `codexLocalProviderArgs` emits the same three `-c` args with the secret in env, shared across drivers. Would two rows there cover this without a second mechanism in codex.ts? If not, tell me why and I'll take the codex.ts version.

CI is green; the one red check is Vercel's "Authorization required to deploy", which is our org setting, not your branch. main is 506 commits ahead and codex.ts moved in 17 of them, so a rebase is needed regardless. Asking for: your answer on local-inject, then a rebase with 1-3 fixed.

PR #717, opened 2026-09-02
Notify when a turn dies before it starts

Viewofmind:
## What changed

Adds a `turn-failed` notification kind and raises it when a direct turn fails during dispatch.

## Why

`notify.ts` encodes one rule: *"a bot that is **blocked on you** is worth a buzz, and a bot that **finished** is worth one if you asked for it."*

A routine that fails to dispatch already follows it — `routine-failed` fires from the routine path. An interactive turn that fails to dispatch did not: it appended an `error:` activity row and set the bot idle. The row renders fine in chat, so this is not invisible if you are looking at the thread. The problem is that nothing tells you when you are not.

That gap matters most exactly when the bot is unattended, and the causes are mostly configuration a person has to change before any retry can work:

- an unconfigured cloud box (`Cloud box is not configured — add a Box API key or choose Local VM`)
- a Local VM that was never created (`Create the Local VM (App Settings → Local VM)`)
- an engine that cannot reach the selected computer destination

I hit the second one with a bot on `computer: "vm"` whose container had gone missing. The turn produced a single activity row and stopped. Nothing surfaced anywhere until I went looking in the thread, and `server.log` had nothing either.

The notification fires for **every** dispatch failure, including transient ones a retry would clear. That is deliberate — classifying failures into "retryable" and "needs you" is a larger judgement call, and I did not want to make it inside a bugfix. Happy to narrow it if you would rather.

## How it was verified

macOS, Node 24.20.0. `pnpm typecheck` and `pnpm test` both pass on this branch.

- `server/notify.test.ts` — extended to cover the new kind: its title, and that it stays silent when `notifications: false`.
- New API smoke test in `server/index.test.ts` — drives a real dispatch failure over the HTTP API and waits for the `notify` frame on `/api/events`. Confirmed it **fails without** the `notify(...)` call (times out after 5.5s waiting for the frame) and passes with it, so it is not a test that would pass either way.

The smoke test clears the Box token itself rather than relying on file order, because the existing `"saves config keys write-only and reports booleans"` test sets `box_good` and does not restore it.

## Note on the base

This branch is cut from my fork's `main`, which is behind upstream. The diff is the four files below and nothing else, but say the word and I will rebase onto current `main` before you merge.

## Screenshots (UI changes)

None — server-side only. Delivery is unchanged; clients already render whatever frames `notify` emits, so this needs no renderer work.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added notifications when an interactive bot turn cannot start, identifying the affected bot.
  * Notifications respect bot notification preferences and can be delivered without a message body.

* **Bug Fixes**
  * Prevented turn-failed alerts for background card-continuation actions, which report errors through their own failure handling.
  * Prevented duplicate notifications for routine and delegated turn failures.

* **Tests**
  * Expanded coverage for interactive, routine, delegated, and card-continuation failure notifications.

### DIFF
--- a/server/index.test.ts
+++ b/server/index.test.ts
@@ -2565,6 +2565,64 @@ describe("harness HTTP API", () => {
     }
   });
 
+  it("leaves a failed credential-card continuation on the card without buzzing twice", async () => {
+    let botId: string | undefined;
+    let stream: Awaited<ReturnType<typeof openSse>> | undefined;
+    try {
+      expect((await api("PUT", "/api/config", { box: { token: "" } })).status).toBe(200);
+      const bot = (await api("POST", "/api/bots")).body.bot;
+      botId = bot.id;
+      expect((await api("PATCH", `/api/bots/${bot.id}`, {
+        modelSelection: { instanceId: "claude", model: "claude-sonnet-5" },
+      })).status).toBe(200);
+      stream = await openSse(`${BASE}/api/events`);
+      await stream.until((frame) => frame.kind === "hello");
+
+      rmSync(fakeClaudeDump, { force: true });
+      expect((await api("POST", `/api/bots/${bot.id}/messages`, { text: "stay active" })).status).toBe(202);
+      const dump = await readJsonFileWhenReady<{
+        mcpConfig: { mcpServers: { agents: { env: { OMB_COMMS_TOKEN: string } } } };
+      }>(fakeClaudeDump);
+      const token = dump.mcpConfig.mcpServers.agents.env.OMB_COMMS_TOKEN;
+      const requested = await fetch(`${BASE}/api/internal/request-credential`, {
+        method: "POST",
+        headers: {
+          authorization: `Bearer ${token}`,
+          "content-type": "application/json",
+        },
+        body: JSON.stringify({
+          fromBotId: bot.id,
+          fromThreadId: bot.threadId,
+          credentialId: "openaiImageApiKey",
+          reason: "needed for the task",
+        }),
+      });
+      expect(requested.status).toBe(201);
+      const { messageId } = (await requested.json()) as { messageId: string };
+
+      expect((await api("POST", `/api/bots/${bot.id}/interrupt`)).status).toBe(200);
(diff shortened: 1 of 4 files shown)


Viewofmind:
Thanks — the first finding was a real bug, not a nit.

**Double notification on routine failures.** Routines call the same `startTurn` and pass `onDispatchError`, which raises `routine-failed`. My unconditional `notify(...)` meant one routine dispatch failure rang twice. That made routine failures *noisier*, which is the opposite of what this PR is for.

Fixed by gating the buzz to turns a person started themselves — no `automationSource`, no `commsDepth`:

- routines keep `routine-failed` alone
- a delegated sub-turn is already reported to the bot that asked for it, in its own thread, so it needs no second channel
- the interactive case this PR exists for still buzzes

Added the regression you asked for, `"reports a failed routine once, not twice"`. It asserts the notification **count**, not just that `routine-failed` appears — a presence-only test still passes with the bug. Confirmed it fails without the guard:

```
AssertionError: expected [ 'turn-failed', 'routine-failed' ] to deeply equal [ 'routine-failed' ]
```

**Second finding** — the Box token is now cleared in a `finally` too. Worth noting the token is write-only, so there is no prior value to restore; the cleanup leaves the box unconfigured rather than half-set for whatever runs next.

Also rebased onto current `main` (v0.1.49) since the branch was cut from an older base. `pnpm typecheck` clean, both new tests pass, `notify.test.ts` green.

Viewofmind:
Note on the red Windows check — I believe it's a flake rather than this change.

It fails in `server/room-chat-wait.e2e.test.ts` > *"waits for a teammate summoned by a chained @mention who is busy elsewhere"*, with `Matcher did not succeed in time` after ~18s. That's a timing-sensitive multi-bot e2e losing a race on a slow runner; this PR only adds a `notify(...)` call on an already-existing dispatch-failure path.

macOS and Ubuntu pass here. On #724 the mirror happens — Windows and Ubuntu pass, macOS fails on an unrelated Electron GPU/scroll fixture. `main` is green on all three. Different platforms failing on unrelated changes points at flake.

Say the word if you'd rather I chase it anyway.

Issue #718, opened 2026-09-02
Per-bot depth profile: make the output contract a real setting


Viewofmind:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Harness server / API (server/)

### Problem

Bots inherit their host CLI's answer style. `claude` and `codex` are tuned for short coding replies, so a bot asked for analysis often returns a few bullets even when it did substantial work underneath.

Nothing in prompt assembly says what a finished piece of work should contain. The persona says who the bot is. `chiefOfStaffSystemPrompt` says who its teammates are. Neither says how deep to go, or what a report should include.

### Proposal

A per-bot **depth profile** — Quick / Standard / Deep report — injected as a prompt layer alongside persona and roster, the way `chiefOfStaffSystemPrompt` already injects dynamic context per turn.

Deep report would state the shape explicitly: what was examined, what was found, what is verified versus inferred, what is still open, and sources.

Kept byte-stable per bot so the host CLI's prompt cache stays warm. Default Standard, so no existing bot changes behaviour.

**Why I think this is worth doing rather than left to each user:**

One of my bots is an SEO specialist whose `description` field happens to contain an output contract I typed by hand — score 0-10 with justification, and keep "technical foundation readiness" separate from "traffic achieved so far."

I asked it a deliberately plain question, no formatting instructions of any kind. It returned a structured audit in 160 seconds: live Search Console figures, a crawler-visible SSR comparison against three competitors, two scores it explicitly refused to blend, and an unprompted "no invented metrics" caveat where a data source was unavailable. It followed the contract exactly.

So the mechanism already works, on an engine the app already supports. It just isn't systematic — it exists on one bot because a person typed it into a free-text field, and only that bot benefits.

### Alternatives considered

**Leave it in `description`.** Works, and is what I did. But it is undiscoverable, it competes with the persona text for the same field, and every user has to reinvent it.

**Ship it as a bundled skill.** Skills are selected per task; an output contract should apply to every turn, so the volatile-per-turn injection point seems more right than the skill index.

**Do nothing and let users prompt for depth each time.** Works for one-off asks, but not for routines or unattended turns, which is exactly where a consistent report shape matters most.

### Questions before I write any code

1. Per-bot setting, or per-section alongside section context?
2. A separate prompt layer, or folded into the existing persona text?
3. Fixed profiles, or user-editable text with the three as presets?

Happy to implement whichever shape you prefer — I'd rather agree the approach here first, per CONTRIBUTING.

PR #724, opened 2026-09-02
Add a per-bot depth profile

Viewofmind:
Implements #718. Opening it as code because the shape is easier to argue about when you can read it — happy to reshape or drop it if you'd rather answer the design questions there first.

## What changed

A per-bot `depth` setting with three values, injected as a layer of the system prompt alongside persona and roster.

- `quick` — answer at the length the question deserves, lead with the answer, no preamble
- `standard` — **the default, and emits nothing at all**
- `deep` — states what a report contains: what was examined, what was found, what is verified as against inferred, what could not be checked, and what is still open

New file `server/depth-profile.ts`, one field on `BotRecord`, one enum check on `PATCH /api/bots/:id`, one call at the prompt assembly site.

## Why

A bot's persona says who it is and the section context says what the team is doing. Neither says what a finished answer looks like, so the shape of a reply is whatever the underlying CLI defaults to — and the coding CLIs default to terse, which is wrong for a bot whose job is analysis.

The evidence that this works is in #718: I have a bot whose `description` field happens to contain an output contract I typed by hand, and it returns structured, sourced reports from plain questions. This makes that a setting instead of something each user has to reinvent in a free-text field.

One clause in `deep` comes from watching that same bot: it wrote 4.8KB to chat and saved a fuller 7.3KB write-up to a file in its workspace, so the person who asked never saw most of the work. Hence *"Put the report in your reply … save a file as well only when they asked for one."*

## Two deliberate choices

**`standard` emits an empty string.** Every bot that predates the setting resolves to it, so this changes no existing behaviour until someone opts a bot in. It also costs no tokens for bots that do not use it.

**The text is fixed per profile.** It rides the cached system prompt, so a value that varied per turn would break the provider's prefix cache on every message. `depth-profile.test.ts` pins that.

## How it was verified

macOS, Node 24.20.0. `pnpm typecheck` clean, `pnpm lint` clean on the new files.

Five unit tests in `server/depth-profile.test.ts` cover the enum, the silent default, and that an unknown stored value falls back rather than reaching a prompt.

The test I would actually look at is the end-to-end one in `server/index.test.ts`: it runs a real turn through the fake Claude CLI, reads the system prompt the engine received, and asserts the contract is present on `deep` and absent on `standard`. A setting that never reaches the provider is worse than no setting, because the UI would claim it is on.

`pnpm test` — 3104 pass. One pre-existing failure, `server/control-omb.test.ts > launches, drives a real fake-engine turn`, which fails identically on a clean `origin/main` on my machine with none of this applied. I believe it is environment-dependent (it expects `availableEngines` to equal exactly `["claude"]`; I also have `grok` configured).

## Screenshots (UI changes)

None — there is no UI in this PR. The setting is API-only on purpose, so the prompt wording can be settled before a control is designed around it. If you want the dropdown in the same PR, say so and I will add it with before/after shots.

## Checklist

- [x] `pnpm typecheck` and `pnpm test` pass locally (one pre-existing failure, noted above)

## Summary by CodeRabbit

* **New Features**
  * Added configurable answer-depth profiles: Quick, Standard, and Deep.
  * Added an Answer depth setting for each bot, with validation and persistence.
  * Deep responses provide additional guidance to verify reports and identify unresolved items.
  * Existing bots default to Standard behavior without additional instructions.

* **Tests**
  * Added coverage for profile selection, persistence, validation, and prompt behavior.

### DIFF
--- a/server/depth-profile.test.ts
+++ b/server/depth-profile.test.ts
@@ -1,0 +1,70 @@
+// The setting is one enum and one string, so the tests are mostly about the
+// two things that would hurt: a default that is not silent, and a stored
+// value that is not one of the three still reaching a prompt.
+import { describe, expect, it } from "vitest";
+
+import {
+  DEFAULT_DEPTH,
+  DEPTH_PROFILES,
+  depthProfileSystemPrompt,
+  isDepthProfile,
+  resolveDepthProfile,
+} from "./depth-profile.ts";
+
+describe("depth profiles", () => {
+  it("offers exactly three, defaulting to the silent one", () => {
+    expect([...DEPTH_PROFILES]).toEqual(["quick", "standard", "deep"]);
+    expect(DEFAULT_DEPTH).toBe("standard");
+    // the whole point of the default: an existing bot's prompt is unchanged
+    expect(depthProfileSystemPrompt("standard")).toBe("");
+    expect(depthProfileSystemPrompt(undefined)).toBe("");
+  });
+
+  it("falls back rather than letting an unknown value reach a prompt", () => {
+    // values arrive from bots.json on disk and from the API, so "not one of
+    // the three" has to mean standard, not a crash and not a passthrough
+    for (const bad of [null, 42, "DEEP", "verbose", "", {}, [], true]) {
+      expect(resolveDepthProfile(bad)).toBe("standard");
+      expect(depthProfileSystemPrompt(bad)).toBe("");
+    }
+    expect(isDepthProfile("deep")).toBe(true);
+    expect(isDepthProfile("Deep")).toBe(false);
+  });
+
+  it("asks quick for brevity and deep for evidence", () => {
+    const quick = depthProfileSystemPrompt("quick");
+    const deep = depthProfileSystemPrompt("deep");
+    // quick's whole job is to cap length and kill preamble
+    expect(quick).toMatch(/one-line question gets a one-line answer/i);
+    expect(quick).toMatch(/no preamble/i);
+    // and it must not smuggle in the report contract
+    expect(quick).not.toMatch(/still open/i);
+    // both are appended to a running system prompt, so each needs its own
+    // leading separator or it welds onto the previous sentence
+    expect(quick.startsWith(" ")).toBe(true);
+    expect(deep.startsWith(" ")).toBe(true);
+  });
+
+  it("tells a deep bot to keep the report in the reply", () => {
+    // the failure this exists to prevent: a bot writes the real work to a
+    // file in its workspace and leaves a summary in chat, so the person who
+    // asked never sees it
+    expect(depthProfileSystemPrompt("deep")).toMatch(/put the report in your reply/i);
+    expect(depthProfileSystemPrompt("deep")).toMatch(/verified/i);
+    expect(depthProfileSystemPrompt("deep")).toMatch(/still open/i);
+  });
+
+  it("carries nothing dynamic that would move between turns", () => {
+    // This rides the cached system prompt, so the text has to depend on the
+    // profile and nothing else. The way that breaks in practice is someone
+    // interpolating a timestamp, a bot name, or a count into the guidance —
+    // all of which show up as digits or a template hole, and none of which
+    // any of the three profiles legitimately contains.
+    for (const profile of DEPTH_PROFILES) {
+      const text = depthProfileSystemPrompt(profile);
+      expect(text).not.toMatch(/\d/);
+      expect(text).not.toContain("${");
+      expect(text).not.toContain("undefined");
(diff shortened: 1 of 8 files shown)


Viewofmind:
The slop flag was correct — one of my tests was a tautology. Fixed in `2fcf084`.

The offender:

```ts
expect(depthProfileSystemPrompt(profile)).toBe(depthProfileSystemPrompt(profile));
```

`f(x) === f(x)` on a pure function returning a module constant. It cannot fail, so it tested nothing — and the comment above it claimed it was guarding prefix-cache stability, which made it worse than no test at all.

Replaced with one that asserts the property I actually meant. The way cache stability breaks in practice is someone interpolating a timestamp, a bot name, or a count into the guidance text:

```ts
for (const profile of DEPTH_PROFILES) {
  const text = depthProfileSystemPrompt(profile);
  expect(text).not.toMatch(/\d/);
  expect(text).not.toContain("${");
  expect(text).not.toContain("undefined");
}
```

Verified it fails for the right reason — injecting `${new Date().toISOString()}` into the `deep` text turns it red, and removing it turns it green again.

Also tightened a second weak test in the same file. `expect(quick).not.toBe(deep)` said almost nothing; it now asserts what `quick` has to contain, and that it does **not** carry the report contract that belongs to `deep`.

`pnpm typecheck` clean, all 5 tests in `server/depth-profile.test.ts` pass.

Viewofmind:
Note on the red macOS check — I don't think it's this PR.

The failure is `electron/browser-closed-shadow.electron.test.mjs`: *"compact nested scroll moved 0px instead of 600px"*, preceded in stderr by

```
SharedImageManager::ProduceOverlay: Trying to Produce a Overlay representation from a non-existent mailbox
skia_output_device_buffer_queue: Invalid mailbox
```

That looks like the runner's headless GPU compositor rather than a scroll regression, and this PR touches no Electron code — it adds `server/depth-profile.ts`, one field on `BotRecord`, one enum check on the PATCH route, and one call at the prompt-assembly site.

Ubuntu and Windows both pass here. On my other PR (#717) the mirror happens: macOS and Ubuntu pass, Windows fails on `server/room-chat-wait.e2e.test.ts` with `Matcher did not succeed in time` after 18s. Different platforms failing on unrelated changes, with `main` green, reads like flake rather than regression.

Happy to be wrong — if you re-run it and macOS stays red, I'll dig in properly.

Viewofmind:
> Is the persona exposed in the UI to the user?

Answered in code — the setting now has a control, rebased onto current `main`.

It sits directly under **Effort** in the agent profile and copies that card deliberately, because the two settings mean the same kind of thing: a default that sends nothing, plus explicit levels.

```
Answer depth
How much work a reply shows (Standard: nothing is added)
[ Quick | Standard | Deep report ]
```

Same `rounded-xl bg-card p-4` card, same segmented row, same `aria-pressed` and `bg-control` active state. The copy follows the reasoning in your Effort comment — *"Sending nothing is true on every engine"* — which is exactly why `standard` emits an empty string rather than a "be balanced" instruction.

To the other half of your question: the bot's persona is exposed, as the **Instructions** box. That is where this came from — people already hand-write output contracts there. One of my bots has one telling it to score 0-10 and keep foundation readiness separate from traffic achieved, and it returns structured, sourced audits from plain questions. This makes that a setting rather than something each person reinvents in a free-text field.

**Still not exposed: the contract text itself.** A viewer sees three labels, not the wording `deep` injects. I left it that way for now because making it editable per bot gives every bot a unique system prompt, which drops it out of the shared cached prefix. Showing it read-only under the control is the middle option if you want the transparency without that cost — say which you prefer and I will add it.

**One gap I cannot close from here:** the checklist wants before/after screenshots and I have no way to capture the panel rendering. The markup mirrors the Effort block line for line, `pnpm typecheck` is clean, and all 578 UI tests plus the server tests pass — but I have not seen it on screen. Happy to drop the UI back out and keep this server-only if you would rather add the control yourself.

Viewofmind:
**SOUL.md landed while this was open — should this live in the bot folder?**

#833 introduced the bot folder and moved prompt assembly onto a bot record mirrored to `SOUL.md`. This PR adds a separate `server/depth-profile.ts` that layers report-depth instructions into the same prompt, and it was written before that existed.

Two ways forward, and I would rather ask than guess:

1. **Rebase as-is** — depth stays its own module, layered alongside the persona.
2. **Fold it into the bot folder** — depth becomes a field on the bot record beside `soul`, readable and editable in the same place, and inheriting the same protection against a bot rewriting its own contract from untrusted content.

(2) reads as more coherent to me now that the folder exists, but it is a bigger change and it touches your new code, so it is your call. Happy to do either.

PR #727, opened 2026-09-02
Add /research

Viewofmind:
## What changed

A `/research` slash command, built the same way as `/learn` next door: `server/research-command.ts` recognises the command and expands it into a prompt, and `startTurn` runs the result as an ordinary turn.

No driver changes, no capability gate. Unlike `/learn` it is not wrapped in the `skillAuthoring` check, because it needs nothing beyond the tools a bot already has.

## Why

`/learn` turns work into a reusable skill. `/research` turns a question into an answered one with its evidence attached, rather than a confident paragraph.

Left to itself a model answers from what it already believes, checks nothing, and presents the result at uniform confidence. The prompt forces the three things that separate research from recall:

- **look** — go to primary sources, and check what you already have (this thread, your own notes, your skills) before reaching outward
- **corroborate** — every claim the answer rests on gets a second independent source, and disagreements get reconciled rather than resolved by preference
- **say what you could not stand up** — verified kept separate from inferred, gaps named rather than filled. *"NEVER fill a gap with a plausible number."*

It also tells the agent to treat fetched pages as data rather than instructions — a research turn opens attacker-controlled pages by design, so that belongs in the prompt before it reads one.

## Relationship to #724

They compose but neither needs the other: this sets the **procedure**, `depth: "deep"` sets the **shape of the report**. A bot with both gets researched work in report form. Fine to merge in either order, or to take only one.

## How it was verified

macOS, Node 24.20.0. `pnpm typecheck` clean, `pnpm lint` clean on the new file.

Eight tests in `server/research-command.test.ts`, aimed at the two ways this goes wrong:

- **swallowing a message that was never the command** — `research this for me`, `/researching is fun`, `/research-notes`, `please run /research later` all pass through untouched
- **quietly dropping the standards** — the prompt is asserted to still carry corroboration, the verified/inferred split, and the never-invent-a-number rule

Also covered: multi-line requests survive intact, a bare `/research` falls back to the conversation, and the expander does not re-expand its own output (the turn text passes through both expanders, so the second must not treat the first's result as a fresh command).

## Screenshots (UI changes)

None — no UI. It is typed into the composer like `/learn`.

## Checklist

Happy to convert this to an issue first if you would rather agree the shape before code — it followed `/learn` closely enough that a PR seemed the clearer way to show it.

## Summary by CodeRabbit

- **New Features**
  - Added the `/research` command for structured, source-aware answers that include research standards and safety guidance.
  - Research requests now work in direct conversations and rooms, including multiline requests, while regular messages remain unchanged.
  - The command is available in the composer’s slash-command picker.

- **Tests**
  - Added coverage for command parsing, prompt generation, direct and room-based expansion, fallback behavior, and API integration.

### DIFF
--- a/server/index.test.ts
+++ b/server/index.test.ts
@@ -3696,6 +3696,43 @@ describe("harness HTTP API", () => {
     expect(patched.body.error).toContain("not recognized");
   });
 
+  it("expands /research in a room, not just a 1:1", async () => {
+    // The command reaches a room member through a different path than a
+    // direct turn. Expanding in only one of them leaves the other receiving
+    // the literal "/research …" text, which looks like it works until you
+    // read what the bot was actually asked.
+    const lead = (await api("POST", "/api/bots")).body.bot;
+    let room: { id: string } | undefined;
+    try {
+      expect((await api("PATCH", `/api/bots/${lead.id}`, {
+        modelSelection: { instanceId: "claude", model: "claude-sonnet-5" },
+      })).status).toBe(200);
+      room = (await api("POST", "/api/groups", {
+        name: "Research room",
+        memberIds: [lead.id],
+        setup: { bulletin: "", defaultResponder: { kind: "member", botId: lead.id } },
+      })).body.group;
+
+      rmSync(fakeClaudeDump, { force: true });
+      expect((await api("POST", `/api/groups/${room!.id}/messages`, {
+        text: "/research who owns investsights.in",
+      })).status).toBe(202);
+
+      // the CLI dumps the raw stream-json turn, so read the whole prompt
+      // rather than assuming where the text sits inside it
+      const dump = await readJsonFileWhenReady<{ prompt?: unknown }>(fakeClaudeDump);
+      const prompt = JSON.stringify(dump.prompt ?? "");
+      // the procedure arrived, not the raw command
+      expect(prompt).toContain("[/research]");
+      expect(prompt).toContain("who owns investsights.in");
+      expect(prompt).toMatch(/corroborate/i);
+    } finally {
+      await api("POST", `/api/bots/${lead.id}/interrupt`);
+      if (room) await api("DELETE", `/api/groups/${room.id}`);
+      await api("DELETE", `/api/bots/${lead.id}`);
+    }
+  });
+
   it("buzzes when a turn dies before it can start", async () => {
     // A dispatch failure already leaves an error row in the thread, but the
     // person who has to fix it is often not looking at the thread — the cause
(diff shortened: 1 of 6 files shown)


Viewofmind:
Good catch — that was a real hole, fixed in `1598ab2`.

`/research` was expanded in `startTurn` but not in the room path, so a room member received the literal `/research …` text instead of the procedure. It looked like it worked until you read what the bot was actually asked.

The room path now composes the two expanders exactly as `startTurn` does:

```ts
const commandTurn = latestUser?.text
  ? expandResearchTurnText(skillAuthoring ? expandLearnTurnText(latestUser.text) : latestUser.text)
  : "";
```

Added the regression you asked for — `"expands /research in a room, not just a 1:1"` — which drives a real room turn through the fake Claude CLI and asserts the prompt the member received carries the procedure, not the raw command. Confirmed it fails without the fix.

Worth noting `/learn` was already handled on both paths; only the new command was one-sided, so this is a gap in my change rather than a pre-existing one.

`pnpm typecheck` clean, 8 unit tests and the new room test pass.

Viewofmind:
Rebased onto current `main` and registered the command in the new composer menu (`f741a25`).

The slash-command menu landed after I opened this, so `/research` worked if you typed it but was invisible — which for a discoverability feature is most of the point missing. It now sits alongside `/goal` and `/learn`:

```ts
const RESEARCH_COMMAND: ComposerSlashCommand = {
  id: "research",
  label: "/research",
  description: "Answer with sources checked, and say what could not be",
};
```

Offered with no capability gate, unlike its neighbours — `/goal` needs a non-DM group and `/learn` needs the skill recorder plus `agentsMcp`, while `/research` only uses tools the bot already has. Say the word if you would rather it were gated.

**One thing I could not verify:** this touches the composer, so the checklist wants before/after screenshots, and I have no way to capture the menu rendering from here. The change is a list entry mirroring `LEARN_COMMAND` exactly, and `pnpm typecheck` plus `composer-commands.test.ts` pass, but I have not seen it on screen. Happy to drop the UI half back out if you would rather take the server side alone and add the menu entry yourself.

`pnpm typecheck` clean, lint clean on both touched files, 8 unit tests and the room regression pass.

Viewofmind:
Rebased onto current `main`. Six days and ~250 commits had landed underneath this, including the bot folder (#833) and the browser engine swap (#850), so it was worth redoing rather than leaving to rot.

No conflicts, and no changes to the PR's own code. Full suite on Node 24: 3,915 passed, 0 failures. Typecheck clean.

PR #756, opened 2026-09-04
feat(memory): typed, dated, cited entries; a labeled selection in the prompt; a visible over-budget notice

AdityaUmale:
Phase A of the memory work after #754 (`session_search`). Plan and reasoning: `docs/superpowers/plans/2026-09-04-08-memory-structure-and-safety.md` (v2, with the adversarial review log at the end); research: `docs/memory-comparison.md`.

## Problem

`MEMORY.md` was a flat file whose first 200 lines rode every prompt. The bot appends, so the newest note dropped first at line 201. Nothing said what kind of thing a line was or where it came from. The only sign that memory had been cut was one sentence in settings. And a bot that "corrected" a fact deleted the old one with no trace.

## Change

**Structure** — `server/memory-entries.ts` parses and renders six sections (`Preferences`, `Decisions`, `Facts`, `Procedures`, `Episodes`, `History`), one entry per bullet, with an optional leading marker `(YYYY-MM-DD, thread <id>, msg <id>, supersedes <date>, superseded <date>)`. Ids are stored whole, so a bullet's `thread`/`msg` are valid `session_read` arguments. Unknown headings, prose lines, fenced code, and `*`/`+` bullets are kept in place: `render(parse(x)) === x` for every file the parser accepts. A file with no headings is a list of facts. Every existing `MEMORY.md` keeps working unchanged.

**Selection, not prefix** — `loadMemory` returns a labeled selection: profile first (preferences + decisions, 100-line cap, decisions cut first), then facts, procedures, episodes newest-first, unknown sections last, History never. The prompt says it's a selection, tells the bot to `Read` before editing and never write the view back, and shows worked marker examples. The block is redacted on the way out: a key the bot copied into memory never rides a prompt, while the file stays the bot's to write. Under budget the whole file loads. The bot is told never to delete a line — a replaced fact moves to `## History`.

**A notice the user can see** — `memoryCapacity()` feeds a gauge in Settings → Memory (`142 / 200 lines · 9.3 / 24 KB`, what won't load). On the first attended turn while over budget, one `activity` chip with `ok: false` is posted, keyed by the file's size signature on the bot record: once per state per bot, not per turn or per task; saving memory re-arms it; unattended turns (routines, webhooks, delegations) post nothing. `ChatView` now renders `ok: false` chips regardless of the tool-calls setting — the rule `GroupView` already applied. Engines with no private workspace (Grok API, box) get "memory does not load here" instead of a gauge.

**Not in this PR, on purpose** — nothing rewrites a memory file the harness didn't receive from the user. That waits for checkpoints to cover the private workspace (Phase B). `PUT` redaction and a read-only post-turn secret detector are the next PR (Task 5 of the plan).

## Docs

`docs/memory.md` (user-facing), `docs/memory-comparison.md` (the research this implements), the plan with its review log, and a "superseded" note on P6 Task 1 so nobody builds the older parser sketch.

## Tests

- `memory-entries.test.ts` (13): round-trip across every shape, implicit facts, malformed markers, continuation rules, and selection — drop order, profile cap, procedures don't starve facts, History never, unknown sections cut last, byte cap.
- `workspace.test.ts`: selection wiring, redaction at load, capacity, legacy seeds and the stale legacy intro, the prompt's guidance and thread id.
- `index.test.ts`: the visibility eval — one notice per state, none on the second turn or the bot's other task, re-armed by `PUT`, none under budget; prompt carries the labeled selection and thread id; `GET`/`PUT` return `capacity`.
- `activity-visibility.test.ts`: the chip predicate.

`pnpm vitest run`: 3221 passed. `pnpm typecheck`, `pnpm lint`, `pnpm check:contrast` clean.

## Verified on a real engine

Asked to remember a preference and a fact, the bot wrote both under the right headings with dated markers and the full thread id. Asked to correct the fact, it added the new bullet with `supersedes` inside the marker and moved the old one to `## History` with `superseded` — nothing deleted. (First attempt put `supersedes` at the end of the sentence; the worked examples in the prompt fixed that.)

https://claude.ai/code/session_01QiHp5wXnV6xtVER7xdEcnz

## Summary by CodeRabbit

- **New Features**
  - Memory is now organized into sections and selectively loaded within defined line and size limits.
  - Settings shows memory capacity, loaded content, dropped content, and whether memory is available.
  - Users receive a one-time notice when memory exceeds its available budget.
  - Memory handling now masks detected secrets and preserves supported metadata.
  - Failed tool activity and communication updates remain visible even when tool-call activity is hidden.

- **Documentation**
  - Added guides describing memory storage, selection, safety, and planned improvements.


milind-soni:
Thank you for this — typed, dated, cited memory entries and a visible over-budget notice are the right instincts, and the write-up made the intent easy to follow. Re-checked today against `main`: the branch is 343 commits behind, `main` has since grown `memory_update` (the bot's transactional memory write, in `server/workspace.ts`) which this branch does not have, and a trial merge conflicts in six files (`server/index.ts`, `server/index.test.ts`, `server/workspace.ts`, `server/workspace.test.ts`, `src/components/ChatView.tsx`, and `src/components/SettingsPanel.tsx`, which `main` has deleted). `main` overtook it rather than the other way round.

One design point worth carrying forward: injecting a **re-rendered** block into the prompt — a labeled selection that is not the file's own text — breaks `memory_update`'s exact-text replaces, because the bot would be quoting text that does not exist in MEMORY.md. Whatever the prompt shows has to be the file, byte for byte, for the write path to stay honest.

The parts that are being taken, on `main`, as small changes: the over-budget notice is being done on the existing `truncated` flag, and dates and sources on entries are landing this week — see #1040, which does both. If there is a piece of this PR you'd like to bring over on top of that (the typed kinds, for instance), a small PR on current `main` would be very welcome.

Closing this one; thank you for the care in it.

PR #759, opened 2026-09-04
Own the model-facing context across every engine

aivsomkar:
Makes OpenMausBot's transcript, budgeting, and compaction canonical across all engines. Tasks 1–7 of [the hybrid context runtime plan](docs/superpowers/plans/2026-09-03-hybrid-context-runtime.md), which is that plan's own review gate. Tasks 8–11 (the OpenMaus-owned runtime) are deliberately out of scope.

## Six bugs fixed

None of these were written in the plan — they surfaced from checking it against the code.

| | Before |
|---|---|
| **Claude threads could brick** | A dead resume cursor is never cleared by anything, and `classifyError` correctly rules the failure terminal — so every later turn resumed the same missing session and failed identically. No way back except switching engines. |
| **Codex, Pi and ACP silently forgot conversations** | A rejected resume fell through to a fresh session carrying only the current message. The harness recorded a normal turn; the user saw a bot that had forgotten everything. |
| **Double charging** | `openai-compat` and `minimax` sent the whole conversation twice on every rewind, model switch, or delegated result — `driverKind === "grok"` decided a property that three driver kinds share. |
| **Messages vanished** | `insertMessageAfter` wrote the active leaf to SQLite but not memory, so a follow-up sent while a screenshot settled disappeared on restart. |
| **A flat 40-message cap** | Same history for an 8k local model and a 200k Claude. |
| **Tool activity never survived a handoff** | A switched engine saw the chat but no evidence any file had been read or edited. |

## What changed

- **`server/context/`** — a pure layer: budget, projector, replay renderer, sanitizer, compaction planner, resume classifier
- **`contextOwnership`** is a required driver capability, so no engine can silently default into inline replay
- **Compaction** folds older history into a summary instead of dropping it; the display path is never touched and nothing is deleted
- **Rooms** are sized for the responding bot's model, attributed per member, and read the active path
- **`context.prepared`** — a metadata-only event making all of this observable

## Deliberate decisions

- **Token and fold logic is the original pi.dev version** (`3374576`): 40% of window, floor 4k, 128k default, `chars/4`, newest 12 kept. A share scales where a flat subtraction leaves small models permanently over the line.
- **Compaction only runs on replay paths.** A cleanly-resumed Claude or Codex turn pays nothing — the vendor CLI owns its own context.
- **In a mixed room the smallest window wins.** The budget is per-responder but the record is per-thread. Failure mode is a large model seeing less than it could, never lost history.
- **`after-accept` never replays.** An unproven boundary is not a licence to repeat side effects.

## Verification

`vitest 3272 · broker 7 · electron 90/90 · packaged-server smoke · 0 lint errors` on Node 24.14.1.

Run tests on Node 24 — on the shell default of 22 the electron suite reports 10 cancelled subtests and `pnpm test` exits non-zero, which reads like a fresh regression.

Also driven against the isolated fixture: 51 turns carrying 151 context items where the old cap was 40, with a credential-shaped canary confirmed absent from the event log. Compaction **cannot** be driven through that fixture — [`docs/verification/context-runtime.md`](docs/verification/context-runtime.md) explains why and what would close it; the real coverage is `server/context/rebuild.test.ts` against a live `Store`.

## Worth a careful read

- **`server/drivers/claude.ts`** relaunches a turn from inside a `close` handler. The one-recovery-per-thread guard is what prevents a relaunch loop.
- **The compaction port** — I first wrote my own token logic and reverted it to the original; worth confirming it's faithful rather than partial.

## Summary by CodeRabbit

* **New Features**
  * Added model-aware conversation context budgeting and history compaction.
  * Improved conversation recovery when session resumption fails.
  * Added safer, bounded tool activity context with sensitive data redaction.
  * Added context ownership and preparation diagnostics for runtime inspection.
  * Added support for durable compaction records and context-aware message handling.

* **Documentation**
  * Added design, implementation, and verification guides for the context runtime.

* **Tests**
  * Added broad coverage for budgeting, compaction, recovery, sanitization, projections, and driver behavior.


milind-soni:
Thank you for this one — the resume-recovery module in particular is the right idea done the right way: classify the boundary from the driver's own protocol state and never from error text. Re-checked today: the branch is 363 commits behind `main`, the merge state can no longer be computed, and CI never ran on it, so the maintainer's decision is **close and split** rather than try to land 4,368 lines in one go.

The first piece is up: #1038 lifts `server/context/resume-recovery.ts` and its `drivers/claude.ts` hunk onto current `main`, credited to you in the body and the commit. The only adaptation is that the rebuild a recovered turn carries comes from a small `recoveryText` the harness attaches alongside every `resumeCursor` (built by `turn-context.ts`, which `main` already has) instead of the context plan. Your tests are ported, plus one pinning the `init` boundary (a resumed session that emitted `init` and then died is never resent).

The rest — the windowed projection, the compaction pass, `contextOwnership` on the capabilities — is welcome as smaller PRs on current `main`, one concern each, where CI can actually run and reviews can be specific. If you'd like, open the projection first and I'll review it promptly.

Closing this one now; the branch stays available for cherry-picking.

PR #760, opened 2026-09-04
Make compaction verifiable through the control surface

aivsomkar:
**Stacked on #759** — base is `feat/hybrid-context-runtime`, since this depends on the `context.prepared` event. Merge that first.

Compaction could only be proven in unit tests. `docs/verification/context-runtime.md` said so outright. This closes both reasons why.

## Reason one: the override never arrived

`control-omb launch` builds its child env from scratch so the fixture cannot inherit the user's shell — correct for isolation, but it meant `OMB_CONTEXT_WINDOW` never reached the server and no model could be made small enough to overflow on a short thread. It now forwards that and `FAKE_CLAUDE_MODE` when explicitly set, because without them whole features cannot be exercised here.

## Reason two: nothing could force a rebuild

A cleanly resumed turn never compacts — the vendor CLI owns its own context, by design. So compaction needs a turn that *rebuilds*, and nothing in the mapped surface could produce one.

Adds **`edit_bot_message`** and an **`omb edit`** command: the rewind a person performs in the composer, which forks the conversation and answers again. It was unmapped despite being an ordinary user action, and it is the only route to a replay path.

## Proven end to end

On a 12k window: 30 long turns, then an edit →

```
mode: "replay-required"   compacted: true
tokensBefore 5358 against a 4,800-token budget, one record written
```

## Two gotchas found while driving it

Both now in the doc, because both cost me a run:

- **Editing an early message does not work.** The edit forks the branch at that message, discarding the history you just built, so the shorter branch fits again and nothing compacts. Rewind at the *newest* user turn.
- **`compacted:false` on a resumed turn proves nothing.** No rebuild was attempted, so it is not evidence either way.

## Tests

Five new, covering the parts I would otherwise have shipped bare: command mapping, `edit` being treated as mutating (it forks a live conversation and starts a turn, so it must never reach a silently discovered app), argument validation, the handler refusing while the bot is busy, and refusing an empty rewind.

`vitest 3277 · broker 7 · electron 90/90 · packaged-server smoke · 0 lint errors` on Node 24.14.1.

### DIFF
--- a/docs/verification/context-runtime.md
+++ b/docs/verification/context-runtime.md
@@ -47,31 +47,49 @@ that impossible. On a 200k window nothing clips:
 family floor for `claude-*`, not a figure the driver declared. Only
 `"catalog"` means the engine actually said so.
 
-## What this harness cannot prove
+## Driving compaction
 
-**Compaction does not fire here, and that is not a failure.** Two independent
-reasons, both worth knowing before you spend an afternoon on it:
+Compaction needs two things the default fixture does not give you: a window
+small enough to overflow on a short thread, and a turn that REBUILDS rather
+than resumes. A cleanly resumed turn never compacts — the vendor CLI owns its
+own context, and that is by design.
 
-- `control-omb launch` builds its child env from scratch — an allowlist of
-  platform keys plus fixed fixture values (see `childEnv` in
-  `scripts/control-omb.ts`). Arbitrary parent env is dropped on purpose, so
-  `OMB_CONTEXT_WINDOW=…` never reaches the server and every model keeps its
-  real window.
-- Compaction only runs on a turn that will REPLAY — a rewind, an engine
-  switch, an external update, or an `omb-replay` engine. The fixture's fake
-  engine resumes cleanly, so `willReplay` is false and no rebuild is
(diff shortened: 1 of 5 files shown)


milind-soni:
Thank you — this is a small, well-aimed PR, and the framing ("the composer rewind is the only mapped way to make the harness rebuild instead of resume") is exactly right. Re-checked today: the branch stacks on #759, which the maintainer has closed in favour of splitting it (see #1038 for the first piece), so this one cannot land as-is.

Its own delta is up as #1039, rebased onto current `main` and credited to you: `control-omb edit`, the `edit_bot_message` tool, and the `FAKE_CLAUDE_MODE` forwarding in `launchVerificationServer`. Two small adaptations: the tool takes the optional `task_id` its neighbours (`set_bot_model`, `interrupt`) take, with the same ownership and idle checks, and it returns what the route on `main` returns. Your tests are ported.

The `OMB_CONTEXT_WINDOW` forwarding is deliberately left out — it needs the context runtime from #759 and belongs with whichever of those smaller PRs brings the window in.

Closing this one; the branch stays available for cherry-picking.

PR #770, opened 2026-09-04
fix(local-vm): recreate an idle-removed Local VM instead of failing the turn

Viewofmind:
## The bug

`LocalVmIdleTimer` **removes** an unused Local VM rather than pausing it:

```ts
if (status.container === "running")
  await containerComputerAction("remove", undefined, undefined, target);
```

The next turn then fails with **"Create the Local VM (App Settings → Local VM)"** — which reads like a fault the person has to repair by hand, for a container the app itself deleted eight hours earlier. Step away overnight and your first message the next morning is an error.

I hit this on a live workspace. The container had run cleanly for `1d 20h`, peaked at 225 MB, and was removed by the idle timer overnight. Every bot with `computer: vm` then failed until someone opened Settings — including on a task that needed no desktop at all.

## Why this is inconsistent rather than just unfortunate

The cloud branch, ~80 lines below, already does the opposite:

```ts
if (!b && mountsCloudComputer && (wants === "cloud" || …)) {
  broadcast({ kind: "computer", botId: bot.id, state: "provisioning" });
  await box.provisionBox(cfg, bot.id, bot.name);   // absent box: made
}
if (b && !["idle","ready","running"].includes(b.state)) {
  broadcast({ kind: "computer", botId: bot.id, state: "waking" });
  b = await box.readyBox(cfg, bot.id);             // archived box: woken
}
```

Box provisions and wakes on first use. The Local VM did neither. One half of its lifecycle was automated — the removal — and the other half was left to the person.

This gives the Local VM the same treatment, behind the same `provisioning` broadcast.

## Deliberately narrow

`localVmRecreatableOnDemand` is true for exactly one state: the container is gone and a plain `run` brings it back. Everything else still surfaces as before, because it needs a decision rather than a retry:

- no container runtime installed
- daemon down
- image never prepared (a 1.4 GB download is not something to start unasked)
- `create_supported` false
- an existing container that is stale, unmanaged, or has unsafe network/security/persistence

**A stopped container is excluded on purpose.** `statusProblem` says *"This desktop image cannot safely resume; recreate the Local VM"* — so auto-`start` would contradict the file's own stated intent, and auto-recreating is destructive enough to stay a human call.

Per-bot mode keeps its instance cap; creating past it here would quietly do what the lifecycle route refuses with a 409.

The container is up before Cua Driver is, so the turn waits for readiness rather than failing on a desktop that is still booting — bounded at 90s, after which it reports the status it has rather than hanging.

## Where the predicate lives

In `container-computer.ts`, immediately above `statusProblem`, so the two stay in step: every branch the predicate declines maps to a message that function already returns. Being pure, it is covered by `container-computer.test.ts` with a fake runner and no container runtime.

## Verification — please read, it is partial

Run locally:

- `tsc -p tsconfig.server.json --noEmit` — clean
- `vitest run server/container-computer.test.ts server/local-vm-idle.test.ts server/local-vm-lease.test.ts` — **52 passed**, including 4 new
- `oxlint` on all three touched files — no new findings (one pre-existing `no-useless-spread` warning at `server/index.ts:1484`, present on `main` too; I verified by stashing)

**Not run locally:** the full `pnpm test` suite and the packaged/Electron smoke tests. CI is the first run of those. The recovery path itself needs a real podman/docker to exercise end to end, which is why the decision logic was extracted into a pure predicate that tests can reach.

The four new tests cover: idle-removed container → recreatable; stopped container → not; no runtime → not; image not prepared → not.

## Summary by CodeRabbit

* **New Features**
  * Local VM environments can now be recreated automatically when needed for a turn.
  * The system waits for the VM desktop to become ready before proceeding.
  * VM provisioning limits and unsupported states are handled safely.

* **Tests**
  * Added coverage for VM recreation eligibility, including missing, stopped, unavailable, and unprepared environments.

### DIFF
--- a/server/container-computer.test.ts
+++ b/server/container-computer.test.ts
@@ -24,9 +24,11 @@ import {
   containerComputerStatus,
   containerRuntimeStatus,
   containerRunArgs,
+  localVmRecreatableOnDemand,
   managedImageDockerfile,
   perBotLocalVmTarget,
   podmanSecurityIsHardened,
+  SHARED_LOCAL_VM_TARGET,
   setupCommands,
   type CommandRunner,
   type LocalVmTarget,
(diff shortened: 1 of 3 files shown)


Viewofmind:
Full-suite result, since the PR body only claimed the targeted suites.

`pnpm exec vitest run` locally: **193 failed | 2,677 passed** across 32 failing files — and **none of the failures are from this change**. Every one is under `server/drivers/*` (claude, codex, acp, pi, antigravity, agents-proxy), which spawn real agent CLIs this machine does not have installed:

```
claude       installed
grok         installed
codex        not installed
pi           not installed
antigravity  not installed
```

They fail with `ENOENT … dump.json` and `runtime.error` where a real turn was expected, and the long file durations (231s, 260s) are timeouts on processes that never start.

Verified rather than assumed, by running one failing suite on both revisions:

```
server/drivers/antigravity.test.ts  on this branch    12 failed | 18 passed
server/drivers/antigravity.test.ts  on pristine main  12 failed | 18 passed
```

Identical. Nothing under `server/container-computer*` or `server/local-vm-*` fails.

CI has the agent CLIs, so it should be green there — but flagging the local numbers so they are not a surprise if anyone reruns this on a workstation without those tools installed.

milind-soni:
Post-merge triage: both CodeRabbit findings remain in main at aabd0109. readyLocalVmForTurn counts instances before setting (and without checking) localVmProvisionBusy, so two per-bot turns can exceed the cap. It also has no dispatch-cancellation check before run or during the 90-second readiness loop. I had deferred this PR for those reasons; it merged concurrently during review. Recommend a targeted follow-up with concurrent-cap and Stop-during-setup regression tests before relying on this recovery path.

PR #777, opened 2026-09-04
feat(agents): session_search — let a bot recall past conversations

### LINKED ISSUE #720: session_search: let a bot recall its own past threads instead of re-deriving them

Viewofmind:
Closes #720 — and answers the question that issue was waiting on, below.

## The gap

A bot cannot read its own history. `searchMessages()` is wired to one HTTP route for the sidebar, so recall is something the person does by scrolling, never something the model can reach for. Ask a bot what it concluded last week and it asks you to repeat yourself.

This adds `session_search` to the agents tool surface with the two supporting pieces that make a recall tool actually get used: an index, so a wide search is cheap, and prompt guidance, so the model reaches for it before re-asking.

## The isolation question from #720

#720 asked whether recall should reach beyond a bot's own threads. **Reach decides recall**: a bot may read its own threads plus those of the bots `reachablePeers()` already lets it contact — the same set `list_bots` shows and `delegate_bot` accepts, so it honours section *and* the per-pair allow-list from #774.

Anything a bot cannot address, it cannot read. That keeps one rule instead of two: it would be odd for reading a teammate's transcript to be easier than messaging them. `scope: "self"` narrows to own threads.

If you would rather it were self-only by default, that is a one-line change to the `owners` array — say the word.

## FTS5 is an accelerator, not a requirement

Electron's Node 24 ships FTS5. The Node that runs `vitest` may not, and `CREATE VIRTUAL TABLE … USING fts5` throws `no such module: fts5` there. So the index is installed when available, and `sessionSearch` falls back to the same LIKE scan the sidebar has always used — correct, only slower, on a local transcript. The existing comment in `message-db.ts` already made that argument; this keeps it true.

Every test asserts behaviour, never which path answered, and passes on both.

## Two details that fail silently rather than loudly

**External content points at a VIEW, not at `messages`.** FTS5 resolves external content by column name, and a searchable "body" is not a column here — it is a text body OR an activity chip's tool name. `content='messages'` compiles and then fails every read with `no such column: T.body`, which a `catch` turns into "this runtime has no FTS5". The view supplies the column FTS5 looks for.

**Both paths match ANY term, not the literal phrase.** LIKE needed the exact substring, so `"Search Console indexed"` found nothing on one runtime and three rows on the other. A tool whose meaning depends on the runtime is worse than a slow one — `queryTerms()` is now shared by both.

I hit both while testing; neither surfaces as an error.

## Shape of the result

Dated excerpts attributed to the bot that said them, capped per thread so one busy conversation cannot crowd out the rest. An empty result reports how much was searched, because "nothing across 40 conversations" is evidence and "nothing across 1" usually means the scope was wrong.

## Verification

- `tsc -p tsconfig.server.json` and `tsc -b` — clean
- `vitest run` on `message-db`, `store`, `peer-roster`, `chief-of-staff` — **99 passed**, 8 new
- `oxlint` on all five touched files — no new findings (one pre-existing `no-useless-spread` at `server/index.ts:1499`, present on `main`; verified by stashing)
- FTS5 path exercised on Electron's Node 24 via a throwaway harness — same assertions, `index: FTS5`, all pass. The committed suite runs the fallback path on Node 22.

**Not run locally:** the full `pnpm test`. `server/drivers/*` needs agent CLIs this machine lacks (codex, pi, antigravity), and those suites fail identically on `main` — CI is the first full run.

The Chief's prompt is byte-pinned by a golden test, so the added guidance updates that pin deliberately; the same sentence goes to ordinary bots via `peerRosterSystemPrompt`.

## Summary by CodeRabbit

- **New Features**
  - Assistants can now search relevant past conversations for previously shared information and completed work.
  - Search results include dated excerpts and source details to help assistants recall context.
  - Searches can include the assistant’s own conversations and those of reachable teammates.

- **Improvements**
  - Assistants are now guided to check conversation history before asking users to repeat information or redoing work that may already be complete.


Viewofmind:
Closing — #754 landed the same feature and did it first. It was open a day before I started; I checked #720, saw no comments, and did not check for an existing PR. My mistake, and the duplicated review time is on me.

#754 is the better-scoped answer to #720 as well: own-threads-only is the conservative reading, and `session_read` for the full message behind a hit is something this PR did not have.

One thing worth carrying over, filed separately rather than as a competing PR: `ensureRecallIndex()` runs inside `open()` with no guard, and `CREATE VIRTUAL TABLE … USING fts5` is not available on every runtime. On Node 22 the whole message store fails to open, not just search:

```
$ node --experimental-strip-types -e "…appendMessage('t1', …)"
THROWS: no such module: fts5
```

Electron ships Node 24 with FTS5, so packaged users are unaffected — but `pnpm dev:server` and `vitest` break for a contributor on Node 22. Follow-up with a guard and a fallback is on its way.

PR #782, opened 2026-09-04
fix(companion): add Windows Tailscale CLI paths and use path.delimiter in searchPath

### LINKED ISSUE #755: Tailscale pairing always fails on Windows: no Windows path in tailscaleCandidates(), and searchPath() joins PATH with ":"
## Summary

On Windows, Tailscale pairing can never succeed, because `refreshTailnetNameOnce` in `companion/listener.js` cannot locate `tailscale.exe`. The Phone settings panel reports **"Tailscale found — MagicDNS is still needed"** to users who already have MagicDNS enabled tailnet-wide.

The code comment right above the function anticipates exactly this failure mode:

> a panel that says "turn on MagicDNS" to somebody who has MagicDNS on is worse than no message

That is the state Windows is in today.

**Version:** 0.1.50 (Windows 11, x64 installer)

### 1. `tailscaleCandidates()` has no Windows path

```js
return [
    "/Applications/Tailscale.app/Contents/MacOS/Tailscale",
    join(home, "Applications", "Tailscale.app", "Contents", "MacOS", "Tailscale"),
    "/opt/homebrew/bin/tailscale",
    "/usr/local/bin/tailscale",
    "/usr/bin/tailscale",
    "/run/current-system/sw/bin/tailscale",
    "tailscale",
];
```

The default Windows install location is `C:\Program Files\Tailscale\tailscale.exe`, which is absent. Every absolute candidate is a POSIX path, so on Windows the bare `tailscale` PATH lookup is the only one that can ever match — which defect 2 then breaks.

### 2. `searchPath()` joins with `:` instead of the platform delimiter

```js
const searchPath = () => [process.env.PATH ?? "", "/opt/homebrew/bin", "/usr/local/bin", "/usr/bin", "/bin"]
    .filter(Boolean)
    .join(":");
```

Windows delimits `PATH` with `;`. Joining with `:` welds `:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin` onto whatever the **last** existing PATH entry is, corrupting that one entry into a path that does not exist.

This makes the failure position-dependent and confusing to diagnose: whether the last PATH entry happens to be the one holding Tailscale decides whether pairing works.

## Reproduction

With Tailscale installed and MagicDNS enabled tailnet-wide, `tailscale status --json` on the host reports a valid name:

```
"DNSName": "hostname.tailnnnnn.ts.net.",
"MagicDNSSuffix": "tailnnnnn.ts.net",
```

Running the app's own probe logic under Node on Windows, all seven candidates fail:

```
/Applications/Tailscale.app/Contents/MacOS/Tailscale => ERR: spawn ... ENOENT
/opt/homebrew/bin/tailscale                          => ERR: spawn ... ENOENT
/usr/local/bin/tailscale                             => ERR: spawn ... ENOENT
/usr/bin/tailscale                                   => ERR: spawn ... ENOENT
/run/current-system/sw/bin/tailscale                 => ERR: spawn ... ENOENT
tailscale                                            => ERR: spawn tailscale ENOENT
```

Isolating defect 2 by controlling only where the Tailscale directory sits in `PATH`:

```
appended (becomes last entry)  => ERR: spawn tailscale ENOENT
prepended (first entry)        => OK  hostname.tailnnnnn.ts.net.
```

Same machine, same binary, same MagicDNS state — only the position in `PATH` differs.

## Suggested fix

Add the Windows install location to the candidate list, and use the platform's path delimiter:

```js
import { delimiter } from "node:path";

// in tailscaleCandidates()
"C:\Program Files\Tailscale\tailscale.exe",
"C:\Program Files (x86)\Tailscale\tailscale.exe",

// in searchPath()
.join(delimiter);
```

Reading `%ProgramFiles%` rather than hardcoding `C:\Program Files` would also cover relocated installs.

## Workaround

Launch the app from a shell with the directory **prepended** to `PATH` (appending hits defect 2):

```powershell
$env:PATH = "C:\Program Files\Tailscale;" + $env:PATH
Start-Process "$env:LOCALAPPDATA\Programs\openmausbot\OpenMausBot.exe"
```

Then click **Check again** — the panel resolves to `Ready on <host>.<tailnet>.ts.net`.

A reboot also fixes it on machines where the Tailscale entry is not last in the combined machine+user `PATH`, since Explorer then passes the updated environment down. That makes the bug look intermittent across machines, which is worth noting for triage.

santhiprakash:
## What changed

- `companion/src/listener.ts` now imports `delimiter` from `node:path` and uses it in `searchPath()` instead of the literal `":"`.
- `tailscaleCandidates()` now lists the two standard Windows install locations before the bare `tailscale` PATH lookup:
  - `C:\Program Files\Tailscale\tailscale.exe`
  - `C:\Program Files (x86)\Tailscale\tailscale.exe`
- Added a focused regression test in `companion/test/listener.test.ts` that proves the Windows candidates are present and that `searchPath()` joins with the platform path delimiter.

## Why

Fixes #755.

On Windows, `PATH` is semicolon-delimited. Joining with a literal colon welded the user's final `PATH` entry to the fallback `/opt/homebrew/bin`, producing a non-existent path and making the bare `tailscale` lookup fail. At the same time, no Windows absolute candidate existed, so there was no other path to try. The result is that Tailscale pairing silently fails on Windows depending on where `tailscale` appears in `PATH`.

## How it was verified

- `pnpm typecheck` passes (no output = clean).
- `pnpm exec vitest run companion/test/listener.test.ts` passes on Node v24.19.0.
- `pnpm exec vitest run companion/test/mdns.test.ts` still passes, confirming the existing listener imports are unaffected.
- `pnpm lint` reports 37 pre-existing warnings and 0 errors; none are in the changed files.

`companion/test/listener.test.ts` results:

```
 RUN  v4.1.10 /mnt/data/open-source/repos/OpenMausBot

 ✓ companion/test/listener.test.ts (3 tests) 10ms

 Test Files  1 passed (1)
      Tests  3 passed (3)
   Duration  7.05s
```

## Screenshots (UI changes)

None — this is a companion-side CLI discovery change with no renderer surface.

## Checklist

- [x] `pnpm typecheck` and the focused `vitest run` test suite pass locally
- [x] Server/companion behavior change comes with tests (`companion/test/listener.test.ts`)
- [x] No `dist-server/` edits

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Tailscale command discovery on Windows.
  - Corrected PATH handling across platforms, including Windows, macOS, and Linux.
  - Preserved the expected order when searching for Tailscale installations.
  - Improved reliability when locating Tailscale through common installation paths.

- **Tests**
  - Added coverage for platform-specific command discovery and PATH behavior.
  - Verified candidate search order and handling when PATH is unavailable.

### DIFF
--- a/companion/src/listener.ts
+++ b/companion/src/listener.ts
@@ -11,7 +11,7 @@
 // nobody runs is a thing that rots. index.ts owns the listeners now.
 import { execFile } from "node:child_process";
 import { homedir, networkInterfaces } from "node:os";
-import { join } from "node:path";
+import { delimiter, join } from "node:path";
 
 /** Interfaces that exist to tunnel, bridge or mesh traffic — utun (Tailscale
  * and every other VPN), vmnet/bridge (VMs, containers, internet sharing),
@@ -100,6 +100,8 @@ export function tailscaleCandidates(home = homedir()): string[] {
     "/usr/local/bin/tailscale",
     "/usr/bin/tailscale",
     "/run/current-system/sw/bin/tailscale",
+    "C:\\Program Files\\Tailscale\\tailscale.exe",
+    "C:\\Program Files (x86)\\Tailscale\\tailscale.exe",
     "tailscale",
   ];
 }
@@ -109,10 +111,10 @@ const TAILSCALE_BUDGET_MS = 5000;
 
 /** PATH with the usual package-manager locations added back, for the bare
  * `tailscale` attempt. Costs nothing when PATH was already complete. */
-const searchPath = (): string =>
+export const searchPath = (): string =>
   [process.env.PATH ?? "", "/opt/homebrew/bin", "/usr/local/bin", "/usr/bin", "/bin"]
     .filter(Boolean)
-    .join(":");
+    .join(delimiter);
 
 /** Ask the Tailscale CLI where it thinks we are.
  *
--- a/companion/test/listener.test.ts
+++ b/companion/test/listener.test.ts
@@ -1,4 +1,4 @@
-import { delimiter } from "node:path";
+import { delimiter, join } from "node:path";
 import { describe, expect, it } from "vitest";
 
 import { searchPath, tailscaleCandidates } from "../src/listener.ts";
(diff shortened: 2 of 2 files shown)


santhiprakash:
Pushed a follow-up commit to address the CodeRabbit review on the test file:

- `tailscaleCandidates` now asserts the full documented lookup order (macOS app → per-user macOS app → Homebrew → /usr/local → /usr → Nix → Windows Program Files → Windows Program Files (x86) → bare `tailscale`).
- `searchPath` cleanup now deletes `process.env.PATH` when it was originally undefined, instead of restoring the string `"undefined"`.

Both `pnpm typecheck` and the focused `vitest run companion/test/listener.test.ts` still pass.

PR #786, opened 2026-09-04
fix(antigravity): recover after upgrading from agy

milind-soni:
## Why
The 0.1.51 upgrade replaced the old agy CLI integration with the official Google ACP runtime. Two upgrade problems remain:
- Previously saved absolute agy paths and Windows shims keep taking precedence, even after the official runtime is installed.
- Model discovery has a five-second total deadline, too short for the packaged Google server on slower machines.

## Changes
- Recognize legacy agy paths and Windows cmd/bat/exe/ps1 shims; fall through to the verified managed runtime or official ACP on PATH.
- Preserve valid explicit official executables and fail closed for unrecognized custom paths.
- Give setup/model discovery a bounded 90-second cold-start allowance, matching the T3 Code approach. Keep caller-provided shorter test deadlines.
- Document the one-time migration sign-in and verify runtime version changes preserve the existing profile token.

## Evidence
- Seven legacy-path cases reproduced the old resolution failure (six initial cases were observed red before the fix).
- A 5.5-second startup fixture reproduced initialize timed out before the timeout fix.
- All 27 Antigravity tests pass, including account isolation, permission-mode enforcement, interactive questions, and real subprocess ACP turns against the fake engine.
- pnpm typecheck and git diff --check pass.
- Launched the isolated verification server and passed doctor. This is not proof of a real Google sign-in; no live user account, token, or app data was used.

T3 comparison: https://github.com/pingdotgg/t3code/blob/main/docs/user/providers-antigravity.md
T3 also requires separate Google sign-in for the official ACP provider. No account import, API-key fallback, or auth bypass is introduced.

The reporter has not supplied an error screenshot or OS yet, so these are confirmed code-level upgrade defects, not a claim that their exact incident has been reproduced.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved migration from the legacy Antigravity CLI to Google’s official runtime, including saved paths and Windows command shims.
  - Preserved existing Google sign-ins when the runtime updates.
  - Increased startup allowances for slower first-time setup and model discovery, especially on Windows.
  - Added clearer guidance when installation or sign-in is required.

- **Documentation**
  - Added upgrade guidance covering migration, sign-in behavior, startup delays, and expired or revoked sessions.

- **User Experience**
  - Added progress messaging while Google sign-in is starting.

### DIFF
--- a/apps/docs/content/docs/providers/antigravity.mdx
+++ b/apps/docs/content/docs/providers/antigravity.mdx
@@ -15,6 +15,14 @@ OpenMausBot connects to Google Antigravity through Google's official Agent Clien
 
 The runtime download is large because Google ships the ACP server and its local computer harness together. OpenMausBot stores it in its own data directory; it does not modify a system installation.
 
+## After upgrading from the older agy integration
+
+Version 0.1.51 switched from the community `agy` CLI to Google's official ACP runtime. An existing Antigravity app or `agy` login does not sign this separate runtime in. Complete **Install official Antigravity** and **Sign in with Google** in the model picker, then refresh and choose a model available to that account.
+
+Saved legacy `agy` paths, including Windows `.cmd` shims, now resolve to the official runtime. Genuine custom ACP executable paths are still respected. The migration does not copy credentials from another app, remove the old CLI, or change which Google account you authorize.
+
+The Google sign-in belongs to the engine instance, not the runtime version, so ordinary app updates do not require signing in again. Cold starts can be slow, particularly on Windows; setup and model discovery allow up to 90 seconds. If Google expires or revokes the saved session, sign in again when prompted.
+
 ## Multiple Google accounts
 
 Create another Antigravity engine instance for another account. Each instance has a separate private `GEMINI_HOME`, token, settings file, and model catalog. Signing one instance in never replaces another instance's account.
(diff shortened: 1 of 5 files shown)


milind-soni:
Computer-use verification found and fixed an additional server crash: after handing the browser a sign-in URL, an unobserved derived completion promise rejected when authentication expired or was cancelled. Retain the already-observed RPC promise instead. Added cancellation and provider-failure regression cases; 29 Antigravity + 3 setup tests pass, typecheck passes. Also show explicit Starting Google sign-in progress during cold startup.

Isolated real-renderer Electron exercise: saved legacy /old/install/agy setting, simulated official ACP installation, click Sign in with Google, authorize clearly labeled fake consent, send a message with the pre-upgrade gemini-3.7-flash-high selection, reload provider instances and renderer, send again without signing in. Both replies rendered and shared-control-surface transcript confirms two completed turns with unchanged model selection. No real Google account or user app data was touched. This is not verification of real Google OAuth or a physical Windows device.

The initial Ubuntu CI failure was an unrelated Box route polling timeout at server/index.test.ts:2131. All platforms passed on the subsequent sign-in lifecycle commit. Final UI-feedback commit awaits fresh CI.

PR #806, opened 2026-09-05
Bind the local-origin gate before the handlers that use it

milind-soni:
## In plain terms

Since this morning's #802, the packaged desktop app crashes at launch on Linux, and main's own "package + smoke" job has been red. Nothing in the unit legs catches it, so every open pull request now fails that job through no fault of its own (#754 among them).

## Cause

#802 turned the hoisted `function localOnly` into a `const { … localOnly … } = localOriginModule` destructure and left it below the `ipcMain.handle(..., localOnly(...))` registrations. Those run while the module evaluates, so the binding is still in its temporal dead zone when the first one calls it:

```
ReferenceError: Cannot access 'localOnly' before initialization
    at app.asar/electron/main.mjs:1491:37
    at ModuleJob.run
```

## Change

One move: the destructure now sits beside its import (the import is hoisted; the binding has to be too), with a comment saying why. A small node test pins the order statically so the next time someone moves it the failure is a red unit test rather than a red release.

## Test plan

- [x] `electron/local-only-hoist.node-test.mjs` passes; moving the binding back below the handlers turns it red with the line numbers in the message.
- [x] `pnpm check:electron` (112 modules), `pnpm test:electron` green, oxlint clean on both files.
- [ ] The "package + smoke (Ubuntu 24.04 x64)" job on this PR is the real proof; it is the job that has been failing on main.

### DIFF
--- a/electron/local-only-hoist.node-test.mjs
+++ b/electron/local-only-hoist.node-test.mjs
@@ -1,0 +1,25 @@
+// The local-origin gate is bound with a `const` destructure while ipcMain
+// handlers are registered at module top level. If that binding ever sits
+// below the first `localOnly(...)` again, the packaged app dies at boot with
+// "Cannot access 'localOnly' before initialization" — which only the Linux
+// package smoke sees, long after the unit legs are green. This pins the
+// order statically so the failure is a red test, not a red release.
+import { readFileSync } from "node:fs";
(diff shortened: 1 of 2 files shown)


milind-soni:
Superseded by #805, which landed the same hoist first. Re-opening only the static order guard as a separate PR.

PR #801, opened 2026-09-05
feat(approvals): bring native provider permission modes to parity

milind-soni:
## Summary
- Keep Ask / Auto / Full / Custom, with provider-specific support and explicit local desktop consent for elevated modes.
- Add Full access mappings for Claude, Antigravity, Cursor, Grok and OpenCode alongside Codex. Custom remains Codex-only.
- Use native Auto review for Codex, Claude and Cursor; unsupported providers fall back to Ask. Preserve native escalations, user questions and unattended safeguards.
- Require returning to Ask or Auto before changing providers with elevated permissions.
- Document behavior and add a durable isolated Electron smoke test for private grant transitions and resumed approval modes.

## Verification
- Full suite: 3,611 passed; 20 skipped; 1 todo (309 passing test files).
- Electron tests: 134 passed.
- Typecheck and production build passed.
- Isolated control fixture and Electron approval smoke passed; HTTP elevation rejected and Full → Auto → Ask transitions verified.
- Provider CLIs in integration smoke are fakes; live provider accounts were not exercised.

## Scope
Core permission handling parity, not an exact copy of every T3 permission option. Full access remains opt-in, Custom stays Codex-only, and remote elevation remains disabled. No version bump or release.

## Summary by CodeRabbit

- **New Features**
  - Approval modes now use each provider’s native review and permission controls.
  - Auto mode is available across more supported providers, with clear fallbacks to Ask where native automatic review is unavailable.
  - Full access is supported for additional providers, with provider-specific behavior.
  - Approval requests and native provider confirmations are surfaced more clearly.

- **Documentation**
  - Added provider mappings, contributor verification steps, and updated migration guidance for existing Auto-mode bots.
  - Clarified Full access limitations, including questions, confirmations, and operating-system permissions.

PR #809, opened 2026-09-05
fix(approvals): say why Auto stopped when nobody started the turn

AdityaUmale:
Follow-up to @cwill11's report that Auto mode "still asks for many commands" once a fleet is running ([Discord, 2026-09-03](https://discord.com/channels/1496552238640926740/1496552239077261322/1545187549960609862)).

## Not a missing mode

The "yolo" tier already shipped as **Full access** (#792, #801). This PR fixes the other half of that report: *why* Auto appeared broken.

## The actual cause

Auto is switched off entirely for an unattended turn. A webhook — or a bot delegating to another bot — marks the asker unattended (`server/index.ts:3744`), and `approvalModeForTurn` downgrades Auto to Ask before the provider spawns:

```js
// Native reviewers can approve before a permission reaches this process.
// Unattended Auto must therefore downgrade before spawning the provider.
if (mode === "auto" && isUnattended(bot.id)) return "ask";
```

That is deliberate and stays. The bug is what the resulting card *said*:

> This action needs you, so Approve for me stopped to ask.

That names the wrong cause and implies the next action might pass. It will not — with a fleet delegating between bots, every card looks like this until someone answers. So Auto reads as broken rather than paused, which is exactly the report.

## The fix

Say what actually happened, and name the mode that keeps running:

> A webhook or another bot started this turn, so Approve for me is paused and every action asks. Full access keeps working unattended.

The Full access hint is suppressed on providers where `supportsApprovalMode(driverKind, "full")` is false, so the card never points at a setting the user cannot pick.

The mode picker carries the same caveat, so the trade-off is visible before a fleet is running rather than only once it stalls:

> The provider reviews routine actions and asks about others; unattended turns always ask

## Refactor

The held text was a four-branch nested ternary inline in the `request.opened` fold. It moves to `approvalHeldReason` in `auto-approve.ts`, beside the verdict logic it describes — an if/else chain, unit tested, with the native and sandbox notes still ahead of any mode explanation. A card whose bot record is gone keeps the sandbox note it had before.

**No decision logic changes.** Only the explanatory copy and where it is computed.

## Verification

Five unit tests on `approvalHeldReason`. Confirmed they are genuine regression tests by restoring the old copy — 2 fail:

```
× says Auto is paused, not picky, when nobody started the turn
× does not offer Full access to a provider that cannot reach it
```

Drove the isolated fake-engine fixture per `docs/verification/README.md` — `doctor`, `new-bot`, `send`, `wait`, `messages` — and a live turn still settles with the change in place:

```
wait result= settled
user/text: hello
bot/text: hello from fake claude
```

**Limitation, stated plainly:** the shared control surface cannot set approval modes or fire an unattended webhook turn, so the new card copy itself is proven by unit tests and the wiring by typecheck — not end-to-end by that harness. Adding a map entry for it would need a new control command.

Full suite: **3666 passed**, 1 unrelated load flake (`group-goal-run.e2e.test.ts` timed out at 20s while taking 173s under full-suite load; passes isolated both with and without this change, and asserts nothing on `held`). `pnpm typecheck` clean; `oxlint` clean apart from one warning pre-existing on main.

## Summary by CodeRabbit

- **New Features**
  - Added clearer explanations when permission requests are paused, including approval mode, unattended activity, sandbox restrictions, provider approval, and Full access availability.
  - Approval messages now provide more relevant context for questions and actions that always require approval.
  - Peer-initiated activity no longer inherits a person’s Full or Custom approval permissions.

- **Documentation**
  - Updated the Auto approval mode description to clarify that unattended turns always require approval.

### DIFF
--- a/server/auto-approve.test.ts
+++ b/server/auto-approve.test.ts
@@ -5,6 +5,7 @@
 import { describe, expect, it } from "vitest";
 
 import {
+  approvalHeldReason,
   approvalKey,
   autoDecision,
   autoVerdict,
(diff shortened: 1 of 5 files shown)


milind-soni:
Reviewed and resolved the main-branch conflict. Adjusted the explanation for the newer peer-origin approval policy: a peer-started Full bot is explained using its origin-adjusted Auto mode, and the card no longer suggests Full access bypasses that policy. Added a regression test. 79 focused tests (including the isolated control-omb chat fixture) and server typecheck pass locally; waiting for refreshed CI before merging.

PR #814, opened 2026-09-05
fix(memory): keep the message store open when SQLite has no FTS5

Viewofmind:
Follow-up to #754 — small, and only the guard.

## The bug

`ensureRecallIndex()` runs inside `open()`, and `CREATE VIRTUAL TABLE … USING fts5` is not available on every runtime. Where it is missing the throw escapes `open()`, so **the whole message store fails — not just recall.**

On Node 22, a fresh checkout can neither run the dev server nor the tests:

```
$ node --experimental-strip-types -e "…appendMessage('t1', …)"
THROWS: no such module: fts5

$ pnpm vitest run server/message-db.test.ts
Tests  11 failed (11)
```

`node:sqlite` did not carry FTS5 before Node 24. Electron ships Node 24, so a packaged user never sees this — a contributor on Node 22 sees nothing else. `engines` asks for `>=24`, but the failure mode is a store that will not open rather than a version warning.

Verified on both runtimes here:

```
node22 recallIndexAvailable: false
node24 recallIndexAvailable: true      (Electron's node, ELECTRON_RUN_AS_NODE=1)
```

## The fix

Index creation is guarded, and `recallMessages` falls back to a LIKE scan over the same rows — which is what `searchMessages` has always done, and what the comment above it already argued was adequate for a local transcript.

The fallback deliberately **matches the indexed path** rather than merely returning something:

- **AND across terms**, because `ftsQuery` emits space-separated quoted terms and FTS5 reads those as "every term must appear". OR would answer a wider question on one runtime than the other — a tool whose meaning depends on the runtime is worse than a slow one.
- **the same stop-word list**, so `"archive reference on"` behaves identically.
- **matched terms bracketed** the way `snippet()` marks them, so a caller renders one shape of hit.

It gives up relevance ranking for recency. That is the honest trade.

## `recallIndexAvailable()`

Answers from an in-memory probe, not from whether a transcript has been opened. A caller can ask before any database exists — a test deciding whether an index-only case applies — and "not opened yet" is not the same answer as "this Node has no FTS5".

The one case that tests the index itself (backfilling rows that predate it) skips where there is no index to backfill. Everything else asserts behaviour rather than which path answered, and runs on both.

## Verification

```
before (main, Node 22)   Tests  11 failed (11)
after                    Tests  15 passed | 1 skipped (16)
message-db + store       Tests  78 passed | 1 skipped (79)
FTS5 path on Node 24     ranked recall intact, "[container]" snippet unchanged
tsc -p tsconfig.server.json   clean
oxlint on both files          no findings
```

Not run locally: the full `pnpm test`. `server/drivers/*` needs agent CLIs this machine lacks, and those suites fail identically on `main`.

---

Unrelated but worth saying: I had an overlapping `session_search` PR open (#777) and closed it — #754 was open a day before I started and I failed to check. Sorry for the duplicated review time. This is the one piece from that work that still applies, rebased onto yours.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with environments that do not support advanced full-text search.
  - Message storage now opens and saves messages even when full-text indexing is unavailable.
  - Message recall continues to support scoped searches, result limits, stop-word handling, and search terms containing special syntax.
  - Search results include relevant matching excerpts when full-text search is unavailable.
  - Recall results and message details now identify the original peer for relayed user messages.

- **Tests**
  - Added coverage to verify consistent recall behavior across supported search modes.

### DIFF
--- a/server/message-db.test.ts
+++ b/server/message-db.test.ts
@@ -12,6 +12,7 @@ import {
   insertMessage,
   readMessageText,
   readThread,
+  recallIndexAvailable,
   recallMessages,
   searchMessages,
   setActiveLeaf,
@@ -199,7 +200,11 @@ describe("message-db", () => {
     expect(readMessageText("dm", "m-bot")!.peer).toBeUndefined();
   });
 
-  it("recall indexes rows that predate the index", () => {
+  // Backfill is a property of the index itself, so this case only has meaning
+  // where FTS5 exists. Without it the DROP TABLE below has no table to drop and
+  // nothing to rebuild — recall still answers from the LIKE fallback, which the
+  // suite at the end of this file covers.
+  it.skipIf(!recallIndexAvailable())("recall indexes rows that predate the index", () => {
     // simulate a database written before messages_fts existed
     insertMessage("t-old", msg("m1", "legacy row about the quarterly forecast"));
     closeMessageDb();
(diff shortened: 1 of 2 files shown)


milind-soni:
Holding this one out of the next release for now. The first-recall initialization and fallback-only eight-term truncation findings are valid. The broad initialization catch also classifies every index failure as missing FTS5, and the no-FTS tests do not force that path on the supported CI runtime. Please add deterministic no-FTS coverage, preserve all query terms, initialize before selecting the path, and handle missing FTS5 separately from other database failures. LIKE substring matching also differs from the indexed token semantics and should be made explicit or reconciled.

Viewofmind:
Rebased onto current `main` — #817 touched the same file, so this needed redoing rather than merging.

Two things changed while rebasing:

**The fallback now returns `peer` too.** #817 added peer attribution to `RecallHit`, and without it the fallback credited a relayed `ask_bot` line to the person rather than the bot behind it. It now uses the same `peerAuthor()` resolution as the indexed path, so `recall names the bot behind a line another bot delivered with ask_bot` passes on both runtimes.

**The backfill case is guarded in place** rather than duplicated. It tests the index itself, so it skips where there is no index to backfill; everything else asserts behaviour rather than which path answered.

The bug is unchanged on current `main`:

```
$ node --experimental-strip-types -e "…appendMessage('t', …)"
STILL THROWS: no such module: fts5
```

Verification after the rebase:

```
Node 22 (no FTS5)   message-db + store   79 passed | 1 skipped
Node 24 (Electron)  index: true, "[container]" snippet unchanged
tsc -p tsconfig.server.json               clean
oxlint on both files                      no findings
```

milind-soni (changes requested):
Reviewed faca7b5988b1d7ce78b5800d26d564b92bb10ae5. Reproduced three issues in disposable databases:

- P1, message-db.ts:105–143: a database created with FTS5 and reopened on a runtime without it passes CREATE VIRTUAL TABLE IF NOT EXISTS because the table already exists. The readiness flag becomes true, and retained FTS triggers also make ordinary message inserts fail with `no such module: fts5`. Seed on Node24, reopen on Node22.12 with experimental SQLite: first recall succeeds, second recall and insert fail. This needs safe trigger/index handling and later rebuild, not just catching creation errors.
- P2, line 442: readiness is checked before db() initializes it. On an FTS5 runtime, the first recall uses LIKE/recency but the identical second query uses FTS exact matching/ranking.
- P2, line 495: slicing query terms to eight silently discards ninth-and-later AND conditions, returning messages that do not match the requested query.

Focused original tests pass, but fail to cover these cases. Small local regressions/fixes for the latter two pass on Node24 (82 tests) and Node22 (80, two index-only skips); the cross-runtime existing-database failure remains a blocker. Please also add the original author's Signed-off-by trailer per CONTRIBUTING.md.

milind-soni:
Thanks for investigating this carefully. We are retaining the documented Node.js 24+ requirement rather than adding a second recall implementation for Node 22. Closing this compatibility proposal without merging; a reproducible failure on a supported runtime would merit a focused report.

milind-soni:
Thank you for this — the guard is carefully built, the probe is memoised properly, and the fallback keeps the indexed path's contract, which is more than most fallbacks manage. Re-checked today against `main`:

- `package.json` engines is `node >=24`, and the app ships Electron 43 (Node 24 inside), so every supported runtime has FTS5. The "Node 22 checkout" the PR is written for is outside the supported range.
- That makes the LIKE scan a second search implementation that no supported build ever runs — and when it did run, it would change ranking (recency instead of bm25) without saying so. A search that silently answers differently on one machine is harder to support than one that refuses.

The part of this PR that is unarguably right is the diagnosis: a bare `no such module: fts5` from inside `open()` tells nobody what to do. So the fail-loud version is up as #1041 — the one `CREATE VIRTUAL TABLE` is wrapped and rethrown as an error naming the fix (install Node 24+) and the running Node version. It credits this PR.

Closing this one in favour of that; thanks for finding the edge and for the care in the write-up.

PR #858, opened 2026-09-06
fix: ignore Codex output after turn settlement

rahul-vanyar:
## Summary

- stop buffering Codex output once a turn starts settling
- stop processing remaining messages in the same stdout chunk after `turn/completed`
- add a regression where an approval request follows completion in the same write

## Why

Follow-up to #852. During the bounded shutdown wait, a late app-server request could still be processed after settlement began. In Full access that request could receive an automatic approval.

## Validation

- `pnpm typecheck`
- `pnpm vitest run server/drivers/codex.test.ts` — 54 passed
- regression fails without the guard and passes with it

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented approval requests received after a turn completes from being processed or surfaced.
  - Ensured completion events reliably stop further response handling.

- **Tests**
  - Added coverage verifying that late approval requests are ignored after turn completion.

### DIFF
--- a/server/drivers/codex.test.ts
+++ b/server/drivers/codex.test.ts
@@ -169,6 +169,13 @@ describe("CodexDriver turns (fake app-server)", () => {
     expect(threadStart.params).toMatchObject({ model: "gpt-5.6-sol", modelProvider: "openai" });
   });
 
+  it("ignores requests received after turn completion", async () => {
+    await create({ mode: "late-request" });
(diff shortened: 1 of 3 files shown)


rahul-vanyar:
All substantive CI is green on Windows, macOS, Linux, Android, iOS, packaging, and control-plane. CodeRabbit reports no actionable comments. The remaining Vercel status is the expected fork deployment authorization check. Ready to merge.

PR #859, opened 2026-09-06
Localize approval cards in Brazilian Portuguese

alisson-acioli:
## Summary

- complete `src/locales/pt-br.json`: the 75 approval-card strings extracted in #843 now ship translated, taking the pack from 7% to 100% of the English catalog
- record the reviewed English source hash for every value (`node scripts/generate-locale.mjs pt-br --accept`), so `pnpm i18n:check` flags this copy as stale if the English changes instead of letting it silently look current
- mirror the German block in `ApprovalCard.i18n.test.ts`, plus one case for the spoken skill prompt

No source changes — the extraction already landed in #843, so this is reviewed copy and its tests. `pt` still routes onto this pack, so a plain `pt` system language gets it too rather than falling back to English.

## Translation notes

Two places where a literal rendering would be wrong, worth a reviewer's eye:

- **`approval.voice.*` is spoken aloud**, so it reads as speech rather than as a transliteration. English "Should I {action} it?" cannot keep a bare infinitive in Portuguese: the object pronoun attaches to the verb and changes its ending, so "atualizar" + "it" is "atualizá-la", never "atualizar ela". `approval.voice.actionUpdate` / `actionEnable` therefore carry the pronoun and `approval.voice.skill` asks "Devo {action}?" — assembling *"Devo atualizá-la?"*. Both keys are read only by `approval.voice.skill` (`PendingApproval.tsx:75`), so nothing else is affected. The new test pins the assembled sentence, since the seam is invisible in the JSON alone.
- **`approval.voice.command`** interpolates the raw tool name, not a translated label. German fixed that English seam by adding a verb; pt-BR does the same, so `{requester} quer executar {tool}` reads correctly: *"Mochi quer executar Bash. pnpm test. Devo permitir?"*.

Register is informal ("você"), consistent with the `noEngines` strings already in the pack and with the German pack's "du".

## How it was verified

- `pnpm exec vitest run src/components/ApprovalCard.i18n.test.ts src/components/ApprovalCard.test.ts src/lib/i18n.test.ts scripts/generate-locale.test.mjs` — 45 passed
- `pnpm exec vitest run` (full suite) — 3971 passed, 20 skipped, 1 todo, 0 failed
- `pnpm i18n:check` — 8 catalogs valid, 81 English strings; `pt-br.json: 100% (81/81)`
- `pnpm exec tsc -b` — clean
- `pnpm exec oxlint src/locales src/components/ApprovalCard.i18n.test.ts` — clean

Rebased onto dd7e4f16.

## Screenshots (UI changes)

None: this is a string-only change and touches no component, so the approval card's layout is identical. The copy is reachable through **Settings → General → Português (Brasil)**, and the tests above assert the rendered markup for the card, the pending strip, the action buttons, and both spoken prompts.

## Checklist

- [x] Server behavior changes come with tests — n/a, no server change
- [x] No `dist-server/` edits
- [x] macOS-only code is platform-gated — n/a
- [x] Locale changes pass `pnpm i18n:check` and are reviewed by a native speaker (pt-BR)

## Summary by CodeRabbit

* **New Features**
  * Added Brazilian Portuguese translations for approval workflows, including tool actions, statuses, voice prompts, accessibility labels, and approval controls.
  * Improved Portuguese approval messaging for skill updates and command approvals.
* **Tests**
  * Added coverage for Brazilian Portuguese approval cards, pending and denied states, spoken prompts, approval strips, and action buttons.


alisson-acioli:
The `typecheck + test (windows-latest)` failure here is unrelated to this PR — could someone with write access re-run that leg?

The failing case is `server/room-chat-wait.e2e.test.ts:243` ("waits for a teammate summoned by a chained @mention who is busy elsewhere"), which timed out on its inner `expect.poll`. This PR changes only `src/locales/pt-br.json`, `src/locales/source-hashes.json`, and `src/components/ApprovalCard.i18n.test.ts`; nothing in that e2e reads the catalog or `t()`. Ubuntu and macOS both passed, and `vite.config.ts` sets `fileParallelism: false`, so the six added tests (plain `renderToStaticMarkup`, ~20 ms) never run alongside that file.

It looks like the same Windows timing flake #778 addressed in this file. That commit raised the `describe` ceiling to 45s, but the three inner polls still carry their own 15s budget (`:207`, `:243`, `:318`), so a slow Windows leg can still exhaust a poll well under the ceiling — which is what happened here. Happy to send that as a separate PR if it's worth closing, but I've kept it out of this one.

alisson-acioli:
Heads-up on an overlap with #874, which adds ten `approval.held.*` keys — the note a held approval card shows, which is currently composed server-side and printed verbatim, so it is the one line of the card that no pack can translate.

Both PRs write `src/locales/pt-br.json` and `src/locales/source-hashes.json`, so they conflict on `main`. The keys are disjoint, though: this PR brings the 81 card strings, #874 brings the ten `held` notes. Nothing here is overwritten or restated.

I have rebased #874 on top of this branch and resolved both files key-by-key rather than by taking a side, so the two compose to `pt-br.json: 100% (91/91)` — the card ends up fully translated including the held note. Full suite after the rebase: 3854 passed, 20 skipped, 0 failed; `i18n:check` valid at 91 English strings.

Nothing needed on this PR. It is the older one and stands on its own, so it should go in first — I will rebase #874 after it lands. Worth knowing that a squash merge here will turn this commit into a duplicate inside #874, which I will clear with that rebase.

One consistency note in this PR's favour: #874's `approval.held.localComputer` cites the Always allow button, and it uses "Sempre permitir" — the same wording as `approval.action.alwaysAllow` here — so the note and the button it names agree.

Issue #864, opened 2026-09-06
Localize the iOS app: a String Catalog, a Language setting, and pt-BR


alisson-acioli:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Not sure

_(The dropdown has no mobile option yet — this is `ios/`.)_

### Problem

The desktop app has had a language setting since #625, and eight packs since
#624 — including `pt-br`. The iOS app has none of it: it is English-only, with
no catalog, no `knownRegions`, and no way to change the language from Settings.
Someone running the desktop in Portuguese and the companion on their phone gets
two different languages for the same product.

This is not a gap anyone has filed yet. #453 asks for multi-language UI and
predates the mobile apps; #735 proposes i18n infrastructure for the React
renderer. Neither covers `ios/`, and no open PR touches it.

What makes it worth doing now is that most of the work is already done and
nobody noticed. SwiftUI's `Text`, `Button`, `Label`, `Section`,
`navigationTitle`, `TextField` and friends take a `LocalizedStringKey`, not a
`String` — so every literal already passing through them is *already* a
localization lookup against a table that does not exist yet. Counted on
`main`:

- **319 unique keys** across 375 call sites in `App/`, `ShareExtension/` and
  `Widgets/` already resolve through `LocalizedStringKey`. Adding a
  `Localizable.xcstrings` translates them without touching a line of Swift.
- **117 `Text(String)` call sites** do not, and need review one by one. At
  least 35 are user data — `Text(bot.name)`, `Text(connection.displayAddress)`,
  `Text(attachment.name)` — which must keep rendering verbatim, translated by
  nobody.

So the iOS app is roughly three quarters localizable today, and the missing
quarter is a small, reviewable list rather than a sweep.

### Proposal

Add a `Localizable.xcstrings` catalog to the iOS target and a **Language** row
in Settings, mirroring the desktop's *Settings → General* — pick a language or
follow the system, with English as the fallback for any key a pack omits, so a
partial pack is a usable pack exactly as `docs/localization.md` already
specifies for the renderer.

Two languages ship: **English and Brazilian Portuguese.** English is the source
catalog and stays the default — every key's English value is the literal
already in the Swift, so an English user sees byte-identical copy before and
after, and any key a translation omits falls back to it. Nothing that reads
correctly today can start reading wrong. pt-BR is the second column, which I
write and maintain; the other seven desktop languages then become one column
each, addable independently by whoever speaks them.

### Shape of the work

[...]

PR #865, opened 2026-09-06
Localize the iOS app: the seam, a Language setting, and Settings in pt-BR

alisson-acioli:
Part 1 of #864. Adds the iOS string catalog, a **Language** setting, and
Brazilian Portuguese — with Settings as the first translated screen, since that
is where the picker lives and so the screen proves itself.

## Why this is small

SwiftUI's `Text`, `Button`, `Section`, `navigationTitle`, `Label` and
`TextField` take a `LocalizedStringKey`, not a `String`. Every literal already
passing through them was *already* a localization lookup — against a table that
did not exist. Counted on `main`, that is **319 unique keys across 375 call
sites**.

So the seam is one modifier at the root:

```swift
.environment(\.locale, AppLanguage.resolved(language).locale ?? .autoupdatingCurrent)
```

Verified on the simulator that this carries plain `Text`, `Section` headers,
`Button` titles, `NavigationLink` labels, `navigationTitle`, **and sheets** —
no call site changes, no bundle swizzling, no relaunch.

## English cannot regress

English is the source catalog: every key *is* the English literal already in the
Swift. An English reader sees the same copy as before, and any key a translation
omits falls back to it — the same contract `docs/localization.md` sets for the
renderer. Adding a language is one case in `AppLanguage` and one column in the
catalog.

## The part that was not free

Settings also carried copy typed as `String`, which SwiftUI cannot localize.
Those move to `Text` / `LocalizedStringKey`. The values that are the *user's*
— a computer's name, the offline reason the computer itself sent — become
`Text(verbatim:)`, so the type now records which strings must never be
translated. That distinction is the thing worth reviewing here.

Two things I tried first and rejected, both caught by actually running it:

- **`String(localized:locale:)`** looks right and is not: its `locale` formats
  the interpolations, while the *translation* is chosen by the bundle. Strings
  built that way silently followed the phone's language.
- **Resolving an `.lproj` bundle by hand** then fails the other way. There is no
  `en.lproj` — English is the key itself — so English fell through to
  `Bundle.main`, which means "whatever the system prefers", which on a
  Portuguese phone is Portuguese. An app forced to English rendered half in
  Portuguese.

Letting the type system carry it removes that whole class of bug.

## Deliberately not here

- **Other screens.** Catalog-only follow-ups, no Swift, per #864.
- **`CompanionCore`'s user-facing enums.** `ActivityDetail` and `IslandIntro`
  produce their labels as `String` inside the package. The Settings pickers show
  them translated by looking their English text up as a key at the point of
  drawing; localizing the package properly needs `defaultLocalization` and its
  own catalog, which is its own change.
- **`session.actionError`.** Model-level error copy, still English; it belongs
  with the rest of the `Text(String)` work in part 3.

## Testing

- `swift test` in `ios/`: **337 tests, 0 failures.**
- Built and run on the iPhone 17 Pro simulator, both directions:
  app in pt-BR on an English system, and app in English on a pt-BR system.

## Summary by CodeRabbit

* **New Features**
  * Added language selection in Settings, including System, English, and Brazilian Portuguese options.
  * Added Brazilian Portuguese translations across connection, notification, activity, pairing, settings, and troubleshooting screens.
  * The selected language is saved and applied throughout the app.
* **Improvements**
  * Updated settings labels, statuses, accessibility text, and confirmation dialogs to support localization.

### DIFF
--- a/ios/App/AppLanguage.swift
+++ b/ios/App/AppLanguage.swift
@@ -1,0 +1,43 @@
+// The language the app draws itself in.
+//
+// English is the source catalog: every key in `Localizable.xcstrings` *is* the
+// English literal from the Swift, so a key a translation omits falls back to
+// copy that reads correctly rather than to a key name. A partial pack is a
+// usable pack, which is the same contract `docs/localization.md` sets for the
+// renderer.
+//
+// Adding a language is one case here and one column in the catalog. Nothing
+// else in the app has to know.
+import Foundation
+import SwiftUI
+
+enum AppLanguage: String, CaseIterable, Identifiable {
+    /// Whatever the phone is set to, which is also what an app with no
+    /// language setting at all would do.
+    case system
+    case english = "en"
+    case portugueseBrazil = "pt-BR"
+
+    var id: String { rawValue }
+
+    var label: LocalizedStringKey {
+        switch self {
+        case .system: "Follow the system"
+        // A language names itself. Someone who cannot read the language the app
+        // is currently in still has to be able to find their own in this list,
+        // so these two are marked `shouldTranslate: false` in the catalog.
+        case .english: "English"
+        case .portugueseBrazil: "Português (Brasil)"
+        }
+    }
+
+    /// `nil` follows the system: the environment keeps the locale SwiftUI
+    /// already handed it, so no choice here means no behaviour change.
+    var locale: Locale? {
+        self == .system ? nil : Locale(identifier: rawValue)
+    }
+
+    static func resolved(_ stored: String) -> AppLanguage {
+        AppLanguage(rawValue: stored) ?? .system
+    }
+}
--- a/ios/App/CompanionApp.swift
+++ b/ios/App/CompanionApp.swift
@@ -14,11 +14,21 @@ struct CompanionApp: App {
     @StateObject private var session = Session()
     @Environment(\.scenePhase) private var scenePhase
     @State private var liveActivities = LiveActivityCoordinator()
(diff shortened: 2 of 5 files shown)


alisson-acioli:
**Before main english**
1-before-main-english
**After english on a portuguese phone**
2-after-english-on-a-portuguese-phone
**After portuguese on an english phone**
3-after-portuguese-on-an-english-phone

alisson-acioli:
Pushed a fix for a real bug in this PR, found by driving the picker in a UI test instead of setting the preference from outside the app.

Picking a language redrew the body but left the navigation title in the previous one — "Settings" sitting above a screen that had already become "Ajustes". A navigation title is drawn by UIKit, which reads its string once and does not re-read it when the SwiftUI environment changes underneath it. Everything else — rows, section headers, button titles, footers — switched immediately.

The fix re-identifies the tree on the chosen language so that chrome is rebuilt. It costs the navigation stack and any view state below it, which seems the right trade for something changed deliberately and almost never; `session` lives on the `App` rather than in `RootView`, so the connection is not dropped, and the `onAppear` `session.connect()` it re-fires is already idempotent because `scenePhase` calls it on every activation.

Worth naming why it slipped through the first round: the seam was checked by launching with the preference already set, so the body was correct in every screenshot and the title was correct too — it had been built in that language from the start. Only tapping the control in a running app shows the difference. I have that UI test locally (it taps the row, picks Português, and asserts the title, the body and the row all switch); the repo has no app-level test target today, so I left it out rather than add one unasked — happy to include it if you want it.

PR #870, opened 2026-09-06
Localize the iOS copy that was still typed as String

alisson-acioli:
Part 3 of #864, and the last of the localization work. The catalog reaches 456
keys.

> **On top of #868, which is on top of #865.** Until those merge this diff shows
> their files too. On its own this PR is the catalog plus eleven Swift files.

## What the catalog could not reach

`Text(someString)` renders its string verbatim in every language — the
`String` overload is not a localization lookup. So a key can be present,
translated, and still never used. That is what this PR fixes, and the types now
carry the distinction that matters: **our copy versus everyone else's words.**

- Copy becomes `LocalizedStringKey` where the value is always ours: the
  composer's `+` menu, the secret card's status line, the scanner's validation
  message, local-discovery and dictation failures, the avatar shapes.
- A holder that can carry either becomes `Text`, and each assignment says which.
  In the attachment preview, the "couldn't be previewed" line is `Text("…")` and
  the `localizedDescription` beside it is `Text(verbatim:)`.

Interpolation needed the same care, because a `String` interpolated into a key
is invisible to it. A routine's row composes `Text` now instead of interpolating
a fallback, so *Deleted agent* is translated while the agent's own name is not.

`Continue with N bots` had English pluralization compiled into the format
string (`bot\(n == 1 ? "" : "s")`). It is a plural entry in the catalog now, so
each language brings its own rule — including a `zero` case for pt-BR, since
CLDR treats 0 as singular there and *0 bot* is not how anyone writes it.

## Deliberately still English

- **Anything assigned to `session.actionError`,** and the `.offline` reason.
  That channel carries the computer's own words; translating only our fallback
  would make a single alert bilingual.
- **Slash-command and predictive-chip prompts** — the labels are translated, the
  prompt text sent to the bot is not.
- **The default quick replies**, which become the user's own editable data on
  first launch. Translating them at seed time is a change to what gets *stored*,
  not to what is displayed, so it did not belong in a translation PR. Happy to
  do it separately if you'd rather a fresh install in Portuguese start with
  Portuguese chips.
- **A routine's schedule summary and its weekday names.** Those want
  `Calendar`'s localized symbols rather than seven catalog entries — a change to
  how the string is built, and its own piece of work.

## Testing

- `swift test` in `ios/`: **337 tests, 0 failures.**
- Walked every screen reachable without a paired computer, in both directions
  (app pt-BR on an English system, app English on a pt-BR system): onboarding,
  pairing, the scanner, Settings, Computers, Connection & Security, Quick
  Replies, Tasks & Routines, Connected Apps, New section.
- All 456 keys have a matching literal in the source; no dead entries.

The screens behind a pairing — chat, agent profile, the cards — are translated
in the catalog but I could not drive them without a real paired computer, so
they are verified by reading rather than by running.

## Summary by CodeRabbit

- **New Features**
  - Added a language preference with support for system default, English, and Brazilian Portuguese.
  - Added a language picker in Settings; the app updates its interface when the selection changes.

- **Improvements**
  - Expanded localization across settings, onboarding, errors, status messages, accessibility labels, and common interface text.
  - Improved handling of dynamic and server-provided text so it displays accurately without unintended translation.
  - Updated the bot continuation label for consistent wording.

### DIFF
--- a/ios/App/AgentProfileView.swift
+++ b/ios/App/AgentProfileView.swift
@@ -546,7 +546,7 @@ private struct ProfileFormSnapshot {
 }
 
 private extension AvatarCrop {
-    var label: String {
+    var label: LocalizedStringKey {
         switch self {
         case .mascot: "Mascot"
         case .circle: "Circle"
--- a/ios/App/AppLanguage.swift
+++ b/ios/App/AppLanguage.swift
@@ -1,0 +1,43 @@
+// The language the app draws itself in.
+//
+// English is the source catalog: every key in `Localizable.xcstrings` *is* the
+// English literal from the Swift, so a key a translation omits falls back to
+// copy that reads correctly rather than to a key name. A partial pack is a
+// usable pack, which is the same contract `docs/localization.md` sets for the
+// renderer.
+//
+// Adding a language is one case here and one column in the catalog. Nothing
+// else in the app has to know.
+import Foundation
+import SwiftUI
+
+enum AppLanguage: String, CaseIterable, Identifiable {
+    /// Whatever the phone is set to, which is also what an app with no
+    /// language setting at all would do.
+    case system
+    case english = "en"
+    case portugueseBrazil = "pt-BR"
+
+    var id: String { rawValue }
+
+    var label: LocalizedStringKey {
+        switch self {
+        case .system: "Follow the system"
+        // A language names itself. Someone who cannot read the language the app
+        // is currently in still has to be able to find their own in this list,
+        // so these two are marked `shouldTranslate: false` in the catalog.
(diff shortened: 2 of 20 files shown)


milind-soni:
Merge review: combined Swift package tests passed (337), and the combined iOS simulator build passed. Verified the interpolation warning with the installed SwiftUI runtime: LocalizedStringKey("Connected to \(name)") produces the catalog key "Connected to %@" and a separate argument, not a fully expanded lookup key. The saved-count singular case is currently unreachable because the label is shown only when count > 1. Duration-unit localization remains a small follow-up, not a regression introduced by this PR.

PR #874, opened 2026-09-06
fix(i18n): translate the note on a held approval card

alisson-acioli:
**Stacked on #859 and #873.** This branch carries their commits, so review those first —
the last commit here is the only one to review as new.

#859 and this PR both write `src/locales/pt-br.json` and `src/locales/source-hashes.json`,
so they conflicted on `main`. They are rebased together here and the keys turn out to be
disjoint — #859 brings the 81 card strings, this brings the 10 `approval.held.*` notes —
so `pt-br.json` lands at **100% (91/91)**, not at one PR overwriting the other.

One thing to know before merging: #859's commit appears here with a new hash from the
rebase. If #859 is squash-merged, that commit becomes a duplicate and this branch needs one
more rebase before it will merge cleanly. Happy to do that whenever #859 lands.

## The gap

Every other line of an approval card goes through the catalog: the title, the tool
label (`approval.tool.runCommand`), the status. The one line explaining *why* the bot
stopped does not — the server composes `held` and the client prints it verbatim.

In German, the only fully translated pack, that was the single English sentence left in
the card. And it is the sentence that matters most: it is the last thing between a bot
and someone's filesystem.

It could not be fixed by moving the text to the client, because `held` is not always
copy. Routine and profile requests put their apply error in the same field
(`redactSecretsInText(detail)`), which is free text with no catalog key, and cards are
persisted in the message DB, so text written before any of this existed has to keep
rendering.

## The shape

The note travels as both. `heldCode` names the catalog key, `held` keeps the English:

```ts
held: approvalHeldReason(heldContext),
heldCode: approvalHeldNote(heldContext),
```

The client translates by key and falls back to the text, which covers three cases at
once: a card saved before `heldCode` existed, an apply error that has no key, and a
client meeting a server that knows a key it does not.

The branching that picks the note stays in **one** place — `approvalHeldNote` returns
the key, `approvalHeldReason` is that plus a lookup on `HELD_NOTE`. The two cannot
disagree about which note a card gets, and the existing `approvalHeldReason` tests
needed no edits at all.

Ten keys, including the two delivery-failure notes that were inline string literals in
`index.ts`.

## Translations, and one deliberate non-translation

All eight packs carry the ten notes. German stays at 100%; Brazilian Portuguese reaches
100% once #859's strings are in (this PR supplies the ten `held` keys that #859 predates);
the other five gain their first translated approval copy (7% → 18%).

**The mode labels stay in English inside them, on purpose.** `ApprovalModeSelector.tsx`
renders `"Approve for me"` and `"Full access"` as hardcoded labels and never calls `t()`,
so a translated note would name a button the reader cannot find. `"Always allow"` *is* in
the catalog (`approval.action.alwaysAllow`, `de` = "Immer erlauben"), so that one is
translated and matches its button.

A test pins both halves of that rule, so a later `--force` regeneration cannot quietly
break it:

```
AssertionError: pt → approval.held.destructive:
  expected 'Isto parece destrutivo, então Aprovar…' to contain 'Approve for me'
```

Source hashes are recorded exactly as `generate-locale.mjs <code> --accept` writes them.

## Verified

- `vitest run` — **3854 passed, 20 skipped, 1 todo, 0 failed** (331 files), run after the
  rebase onto #859. On `main` + #873 alone the count is 3840; the delta is the 8 tests added
  here plus the 6 that come with #859.
- `node scripts/generate-locale.mjs --check` — `locale catalogs valid (8 languages, 91
  English strings)`, with `de.json: 100% (91/91)` and `pt-br.json: 100% (91/91)`
- the merge conflict in both locale files was resolved key-by-key, not by taking a side:
  #859's values are untouched and the ten `approval.held.*` keys are added beside them
- `scripts/generate-locale.test.mjs` — 11 passed
- `tsc -b` and `tsc -p tsconfig.server.json` — exit 0
- `oxlint .` — exit 0, no new warnings

**Each new test was checked against a mutation**, not just run green:

| mutation | caught by |
| --- | --- |
| server note text drifts from the catalog | `ships each note in the English catalog under the same key` + `names the guard that stopped the action` |
| `tFromServer` ignores the catalog | `translates a key the catalog knows` + `prefers the catalog over stale text saved with an older card` |
| a pack translates "Approve for me" | `keeps the untranslated mode labels verbatim in every pack` |

One test walks every combination of source × mode × unattended × permission ×
sandbox × full-access and asserts `approvalHeldReason` is always exactly
`HELD_NOTE[approvalHeldNote(...)]`, so the key and the English cannot diverge for any
input. Another asserts every key in `HELD_NOTE` is actually reachable, so the catalog
carries no dead copy.

## Not in here

The five still-partial packs stay partial. `generate-locale.mjs <code> --accept` cannot refresh
only the new keys in an incomplete pack — it drafts everything missing — so completing
them would add ~510 strings unrelated to this change. Partial packs are the documented
design ("a half-translated language is a usable language, not a broken one"), and
finishing them is its own PR.

## Summary by CodeRabbit

* **New Features**
  * Approval cards now explain why an action is being held, including destructive, sensitive, unattended, sandbox, and local-computer cases.
  * Held approval notices are translated into the viewer’s language across supported locales.
  * Existing cards remain compatible through fallback text when a localized message is unavailable.

* **Bug Fixes**
  * Approval notices now consistently reflect the reason an approval could not be delivered or completed.

* **Tests**
  * Added coverage for approval reason selection, localization fallbacks, and translated approval notices.


alisson-acioli:
Force-pushed: this branch is now rebased onto #859, which was already open and localizes the approval cards in Brazilian Portuguese. The two overlapped on `src/locales/pt-br.json` and `src/locales/source-hashes.json` and conflicted on `main`.

The keys are disjoint — #859 has the 81 card strings, this has the ten `approval.held.*` notes — so I resolved both files key-by-key rather than by taking a side. #859's values are untouched, and the result is `pt-br.json: 100% (91/91)`: with both in, the approval card is fully translated including the held note, which was the point of this PR.

Re-verified after the rebase, not carried over from before: full suite 3854 passed / 20 skipped / 0 failed, `i18n:check` valid (8 languages, 91 English strings, `de` and `pt-br` both 100%), `tsc -b` and `tsc -p tsconfig.server.json` clean, `oxlint .` with no new warnings.

The PR body has the details. Two things worth flagging for merge order: #859 is the older PR and stands alone, so it should land first; and because its commit appears here with a rebased hash, a squash merge there will leave a duplicate here that I will clear with one more rebase.

milind-soni:
Updated this branch with main and the original #859 ancestry, resolving the stacked conflicts without dropping the keyed held notes or Portuguese approval strings. Also changed the server-note catalog lookup to Object.hasOwn and added regression cases for constructor/toString/__proto__; inherited object properties must fall back to the supplied note, not become React children. The 101 focused approval/i18n tests and locale validation passed before integration; combined server verification and fresh CI are running before merge.

milind-soni:
Final local integration verification passed on the submitted tree: 3,888 Vitest tests passed (20 skipped, 1 existing TODO), typecheck passed, 132 focused approval/i18n/isolated-server tests passed, and all eight locale catalogs validate. The final main-sync commit is tree-identical to the tested parent; it only removes the criss-cross merge-base ambiguity after #859 merged. Auto-merge is enabled behind the required CI checks; no checks are bypassed.

PR #875, opened 2026-09-06
fix(chat): keep thinking row above composer

FsDevNinja:
## Summary

- observe transcript size changes while bottom-follow remains armed
- scroll after delayed rows such as TurnPresence mount, in both direct chats and rooms
- preserve detached scrollback instead of snapping readers back to the tail
- add regression coverage for active and detached bottom-follow behavior

## Verification

- focused bottom-follow/composer tests: 11 passed
- pnpm lint: passed with pre-existing warnings only
- pnpm typecheck: passed
- pnpm build: passed
- full suite: 3,830 tests passed; two environment-dependent fixture failures were isolated to a random occupied port and inherited ELECTRON_RUN_AS_NODE, then both files passed on clean focused rerun (26 tests)
- ProofShot against the repository's isolated fake-agent fixture: thinking row fully visible with 15.11px clearance above the composer; zero browser errors and zero server errors

No live conversation data was used during verification.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transcript scrolling so conversations remain pinned to the latest content while users are following the bottom.
  - Updated scrolling to respond when transcript content changes size or new content is rendered.
  - Preserved the user’s current position when they have intentionally scrolled away from the bottom.

### DIFF
--- a/src/components/ChatView.tsx
+++ b/src/components/ChatView.tsx
@@ -67,7 +67,7 @@ import { ActivityRun } from "./ActivityRun";
 import { TurnNarrationRun } from "./TurnNarrationRun";
 import { webhookMessageView } from "@/lib/webhook-message";
 import { splitTranscriptAttachments } from "@/lib/composer-attachments";
-import { BOTTOM_FOLLOW_THRESHOLD, shouldResumeBottomFollow } from "@/lib/bottom-follow";
+import { BOTTOM_FOLLOW_THRESHOLD, shouldResumeBottomFollow, useBottomFollowResize } from "@/lib/bottom-follow";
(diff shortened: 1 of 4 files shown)


FsDevNinja:
## Isolated before / after

Both captures use OpenMausBot's disposable fake-agent fixture; no live conversation data is present.

### Before — parent commit b3c2f74b

Measured clearance: **-32.89px**. The thinking row is behind the composer.

Before: thinking row hidden behind composer

### After — commit 8a802427

Measured clearance: **+15.11px**. The thinking row is fully visible above the composer.

After: thinking row visible above composer

Issue #632, opened 2026-08-31
feat(ios): select model reasoning effort from the companion app


milind-soni:
## Request

Let people view and change a bot model reasoning effort from the iOS companion app, matching the effort selector already available in the desktop app.

This complements #631, which adds model selection on iOS.

## Proposed experience

- Show the current reasoning effort beside the selected model in the bot composer or bot settings.
- Offer only the effort levels supported by that model, using the capabilities reported by the paired computer.
- Save the choice per bot and reflect it on desktop immediately.
- When the model changes, keep the effort if it is supported; otherwise choose the desktop default and explain the adjustment.
- Prevent effort changes while the bot has an active turn.

## Acceptance criteria

- iOS displays the active effort for each bot.
- A user can choose a supported effort while the bot is idle.
- Unsupported model and effort combinations cannot be selected or submitted.
- The selection persists and remains synchronized with the desktop app.
- Changing model or effort cannot alter an already-running turn.
- Loading, offline, and server-error states are handled clearly.

alisson-acioli:
Mostly implemented already, by the same #699 (`d3dfd3bc`) that covers #631, which did not reference this issue so it stayed open.

In current `main`, `ios/App/AgentProfileView.swift`:

- The effort picker is built from `selectedInstance?.capabilities?.effortLevels` (`:88-93`), so only levels that model reports are offered.
- Effort changes are blocked while the bot has an active turn (`canApplyModel` requires `current.busy != true`, `:98`).
- An unsupported model/effort combination cannot be submitted — `canApplyModel` checks `effortIsOffered` (`:103`).

One acceptance criterion behaves differently from the request, so this may be worth keeping open for that alone:

> When the model changes, keep the effort if it is supported; otherwise choose the desktop default and explain the adjustment.

Today an unsupported effort makes `canApplyModel` false, so Apply is simply blocked — the effort is not adjusted to the default, and nothing explains why the action is unavailable.

Issue #631, opened 2026-08-31
feat(ios): choose the AI model from the companion app


milind-soni:
## Request

Let people view and change the AI/model used by a bot directly from the iOS companion app, without returning to the desktop app.

## Proposed experience

- Show the current AI/model in the bot composer or bot settings.
- Offer a simple picker containing the models already configured and available on the paired computer.
- Save the choice per bot and reflect the change on desktop immediately.
- Prevent changes while that bot has an active turn, matching desktop behavior.
- Show a clear unavailable or offline state when the paired computer cannot provide the model list.

## Scope and safety

The phone should select from models the desktop already exposes. Provider setup, API keys, and credentials remain on the computer and are never sent to or managed by the iOS app.

## Acceptance criteria

- The iOS app displays the active model for each bot.
- A user can select another available configured model while the bot is idle.
- The selection persists and stays synchronized with the desktop app.
- Active turns cannot silently change models midway through a run.
- Loading, offline, unavailable-model, and server-error states are handled clearly.

alisson-acioli:
This looks already implemented, by #699 (`d3dfd3bc`, "feat(ios): add bot model settings"), which did not reference this issue so it stayed open.

In current `main`, `ios/App/AgentProfileView.swift`:

- Provider and model pickers built from the instances the paired computer exposes, with an unavailable provider shown as a disabled "Current provider (unavailable)" row and unavailable instances disabled (`:127-146`).
- Changes are blocked while the bot has an active turn — `canApplyModel` requires `current.busy != true` (`:98`), and the UI says so when it is busy (`:157`).
- `canApplyModel` also refuses a model the selected instance does not offer (`:101-105`), so an unavailable selection cannot be submitted.
- Loading and no-provider states are handled (`"Loading models"`, "No model providers are available").

Covered by `ios/Tests/CompanionCoreTests/BotModelClientTests.swift`, added in the same PR.

PR #880, opened 2026-09-06
ux(desktop): put reasoning effort beside the model picker

### LINKED ISSUE #633: ux(desktop): make reasoning effort easier to find beside the model picker
## Problem

Reasoning effort is currently buried in the per-bot settings panel, while model selection is available directly in the conversation header. People naturally expect to choose the model and its effort together, so the current placement makes the effort control easy to miss.

## Proposed experience

Move or surface the reasoning-effort selector inside or directly beside the existing model picker in the desktop conversation header.

A compact presentation could show the active choice together, for example: `GPT-5.6 · High`. Opening it should let the user choose a model and then one of that model supported effort levels without navigating away from the conversation.

## Acceptance criteria

- The current model and effort are discoverable from the conversation header.
- Effort choices update when a different model or provider is selected.
- Only effort levels supported by the selected model are shown.
- Models without configurable effort do not show an empty or disabled control.
- The layout remains usable at narrow desktop widths and does not overcrowd the header.
- Keyboard navigation, focus states, labels, and screen-reader behavior remain accessible.
- The existing per-bot model selection and persistence behavior is preserved.
- There is one shared selection flow so the header and settings cannot drift out of sync.

alisson-acioli:
Closes #633

## Problem

Reasoning effort lived only in the per-bot settings dialog, while the model it qualifies is chosen from the conversation header. Model and effort are one decision to the person making it, so the control was easy to miss where it sat.

## Solution

The levels moved into an `EffortRow` that both surfaces render, writing through the same `setModel` dispatch:

- **Chat header** — the picker's trigger now reads `Claude Sonnet 5 · High`, and the popover carries an Effort row between the model list and *Use a local model*.
- **Settings dialog** — the Model section's hand-rolled segmented control is gone; it renders the same `EffortRow` in its card.

Sharing one component is what keeps the two from drifting. Before this, the dialog patched the bot record through `updateBot` while the picker dispatched `setModel` — two write paths to one field. Both already funnelled into the same `botPatchQueue.enqueue(botId, { modelSelection }, before)`, so consolidating on `setModel` changes no persistence behavior.

The row reads levels off the engine the bot actually runs on, so a level the server would reject (`server/index.ts:1020`, 400; `server/index.ts:3909`, 409) is never offered in the first place.

### Two calls worth flagging

**Default is not the `none` level.** Default sends no level and lets the engine decide; `none` is a level the engine is told to use. Only `pi` declares both, so on that engine they appear side by side. Collapsing them would change what already-configured bots send, so both stay and the tooltips say which is which.

**Pills that wrap, not a segmented control.** `pi` offers Default plus six levels, which a fixed-width segmented control would squeeze in a 380px popover. Measured in the running app with the full union rendered: two rows, 60px, no horizontal overflow. Horizontal scroll was the alternative and would hide options behind an affordance — the opposite of what this issue asks for.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| Current model and effort discoverable from the header | Trigger reads `Model · Level`; the tooltip carries `engine · model · Level effort` |
| Effort choices update when model or provider changes | The row reads `capabilities.effortLevels` off the bot's active engine; effort is per-engine, and `pick()` already drops it when the instance changes |
| Only supported levels shown | Same source the server validates against; nothing else is rendered |
| No empty or disabled control without effort | `EffortRow` returns `null` — no heading, no card, no placeholder |
| Usable at narrow widths | The effort suffix sits inside the existing `@max-4xl/chathead:hidden` span, so the chip folds to the provider mark exactly as before; the tooltip keeps the level |
| Keyboard, focus, labels, screen reader | Real `<button>`s in DOM order, `aria-pressed` on each, `role="group"` + `aria-label="Reasoning effort"`, per-level `title`, and the repo's `focus-visible:ring-2 ring-accent/70` |
| Existing selection and persistence preserved | Same `{ ...modelSelection, effort }` shape and the same queue as before |
| One shared selection flow | One `EffortRow`, one `setModel` dispatch, rendered by both surfaces |

## Verification

Driven against a disposable fake-engine instance (`scripts/control-omb.ts launch`), never a live install.

- Picking **High** in the header wrote `modelSelection: { instanceId: "claude", model: "claude-sonnet-5", effort: "high" }` to the server, left the other bot untouched, and survived a reload.
- Focusing a level and pressing **Enter** flipped `aria-pressed`, updated the header, persisted, and left focus in place.
- The engine-without-effort case was checked against a second fixture whose driver declares no `effortLevels`: the popover goes straight from the model list to *Use a local model*, and the settings section shows only the Model card.
- The settings dialog's `contained` in-flow popover still opens inside the scroller. While checking it I found the popover and the settings card rendering the effort control twice in one view, which reads as a bug — the popover now skips its own row when `contained`, since that caller renders one.

## Tests

`src/components/ModelPicker.test.ts` is new — there was no test for this component. Seven cases: no levels renders nothing (declared-empty and unknown-engine included), only the engine's levels appear, Default and None stay distinct, the active level is marked, `xhigh` renders as `X-High`, and the trigger shows the effort only when the bot carries one.

Each assertion was checked against a mutated implementation; all six mutations (dropping the empty guard, rendering the full union, removing `aria-pressed`, removing the trigger suffix, removing the tooltip suffix) break at least one test. One of them exposed a weak assertion — `toContain("· High")` passed via the tooltip — which is why the trigger's effort span carries a `data-model-effort` hook.

- `vitest run` — 3884 passed, 20 skipped, 1 todo, 0 failed
- `tsc -b` and `tsc -p tsconfig.server.json` — clean
- `oxlint .` — byte-identical to `main` (exit 0, same 31 pre-existing warnings)

No server or iOS files are touched; #632 and #631 cover the iOS side separately.

## Summary by CodeRabbit

* **New Features**
  * Added reasoning-effort selection alongside model selection.
  * Added clear labels for available effort levels, including an “xhigh” option and a “Default” setting.
  * Model picker displays the selected effort level alongside the model name, including in compact views.
  * Bot settings now use the same effort-selection controls for a consistent experience.

### DIFF
--- a/src/components/BotSettingsDialog.tsx
+++ b/src/components/BotSettingsDialog.tsx
@@ -349,7 +349,7 @@ export function BotSettingsDialog({ bot }: { bot: Bot }) {
 
             {section === "access" && <AccessSection bot={bot} derived={derived} />}
 
-            {section === "model" && <ModelSection bot={bot} derived={derived} />}
+            {section === "model" && <ModelSection bot={bot} />}
 
             {section === "permissions" && <PermissionsSection bot={bot} derived={derived} />}
 
--- a/src/components/ModelPicker.test.ts
+++ b/src/components/ModelPicker.test.ts
@@ -1,0 +1,132 @@
+import { createElement } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
+import { afterAll, describe, expect, it, vi } from "vitest";
+
+import type { Bot, InstanceInfo } from "@/state/store";
+import type { EffortLevel } from "../../server/contracts.ts";
+
+// The picker reads the engine catalog off the store, and the store module
+// touches window/localStorage at import time — the same shape
+// ComputerPanel.browser.test.ts uses to render a store-backed component
+// under vitest's "node" environment.
+const fixture = vi.hoisted(() => {
+  vi.stubGlobal("window", {});
+  vi.stubGlobal("localStorage", { getItem: () => null, setItem: () => {} });
+  return { instances: [] as InstanceInfo[] };
+});
+vi.mock("@/state/store", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@/state/store")>()),
+  useStore: () => ({
+    state: { instances: fixture.instances },
+    dispatch: vi.fn(),
+    refreshInstances: vi.fn(),
+    refreshModels: vi.fn(),
+  }),
+}));
+
+const { EffortRow, ModelPicker } = await import("./ModelPicker");
+
+afterAll(() => vi.unstubAllGlobals());
+
+function engine(effortLevels?: readonly EffortLevel[]): InstanceInfo {
+  return {
+    instanceId: "codex",
+    driverKind: "codex",
+    displayName: "Codex",
+    snapshot: { state: "available", version: "1.0.0" },
+    models: { default: "gpt-5.6", options: [{ id: "gpt-5.6", label: "GPT-5.6" }] },
+    ...(effortLevels ? { capabilities: { effortLevels } } : {}),
+  };
+}
+
+function bot(effort?: EffortLevel): Bot {
+  return {
+    id: "atlas",
+    threadId: "thread-atlas",
+    name: "Atlas",
+    title: "",
+    description: "",
+    notifications: true,
+    color: "green",
+    unread: false,
+    modelSelection: { instanceId: "codex", model: "gpt-5.6", ...(effort ? { effort } : {}) },
+    messages: [],
+  };
+}
+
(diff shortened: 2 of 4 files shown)


alisson-acioli:
02-popover-high
07-settings-contained-fixed

milind-soni:
Reviewed against current main alongside the sidebar/settings/portrait changes. Typechecking and focused tests pass. In an isolated real-browser fixture I selected High from the chat-header picker and used the keyboard to restore Default; the pressed state and selected model/effort updated correctly. Reusing the same effort row in settings keeps this small and consistent. GitHub test/build checks are green.

PR #884, opened 2026-09-07
fix(claude): route a signed-out turn to the sign-in card

### LINKED ISSUE #674: Noodle default bot, asks for /login but /login doesn't work
### Before filing

### Area

Chat UI (src/)

### What happened

Image

### Version and install

0.1.46

### Logs

```text

```

alisson-acioli:
Closes #674

## The dead end

On an unauthenticated Claude Code CLI, a turn ended with nowhere to go:

1. bot: `Not logged in · Please run /login`
2. user: `/login`
3. bot: `/login isn't available in this environment.`

The app asks the user to run a command, and then tells them that command does
not exist here. There is no way out from inside the product.

Reproduced on the current `main` by pointing `CLAUDE_CONFIG_DIR` at an empty
directory, against the real `claude` CLI (2.1.263) — all three messages, in
order.

## The cause

Neither phrase is ours; both are CLI output relayed verbatim into the
transcript.

The CLI does report the failure properly. Its assistant frame carries the
machine-readable verdict:

```json
{"type":"assistant",
 "message":{"model":"<synthetic>",
            "content":[{"type":"text","text":"Not logged in · Please run /login"}]},
 "error":"authentication_failed","is_api_error_message":true}
```

`server/drivers/claude.ts` read neither field. The `assistant` case took only
`message.content` and emitted it as `assistant_text`, so the CLI's instruction
became an ordinary bot bubble. The turn did settle as failed, but with
`stop_reason: "stop_sequence"` winning over `terminal_reason: "api_error"`, so
the recorded reason said nothing either.

The path to the sign-in card already exists and already works:

`runtime.error { setup: true }` → `server/index.ts:3155` → an `activity`
message with `tool.setup` → `ChatView.tsx:781` → `ErrorRow` → `<EngineSetup>`.

`codex.ts:1127` and `acp/core.ts:437,459,990` all emit it. The Claude driver
was the only engine with no auth classification on turn output.

Note that `authenticated` from the driver snapshot is produced but never
consumed server-side, so nothing gates a turn on it — a fresh snapshot would
not have prevented this. The gap is in classifying the turn's own output.

## The change

- `claudeAuthFailure()` — reads the CLI's error flag first, then confirms with
  the existing `classifyError` from `retry.ts`. The flag gate is what keeps a
  model answer that merely discusses being logged out from being read as a
  failure; a real reply never carries `error` or `is_api_error_message`.
- The `assistant` case emits `runtime.error { setup: true }` instead of the
  text bubble, so the chat offers the sign-in card that is already built.
- `result` settles `auth_required` — the same vocabulary `codex.ts:1129` and
  `acp/core.ts:992` already use. Every `stopReason` consumer treats it as an
  opaque display string, so a routine that fails this way now names the real
  reason instead of `stop_sequence`.

25 lines of production code in one driver.

## Scope

**This is presentation, not an authentication fix.** It changes nothing about
how sign-in works; it carries the user to the path the app already has.

Deliberately out of scope: the second balloon
(`/login isn't available in this environment.`) arrives with `is_error: false`
and no error field at all — indistinguishable from a normal reply at the
protocol level, so it cannot be classified. It also stops mattering: once the
first message becomes the sign-in card, nobody types `/login` into the chat.

## Tests

`server/testing/fake-claude-cli.ts` gains a `not-logged-in` mode replaying the
frames captured from the real CLI.

- `claude.test.ts` — the turn produces `runtime.error { setup: true }`, emits
  no `assistant_text` and no `content.delta`, and settles
  `ok: false, stopReason: "auth_required"`.
- `claude-auth.test.ts` — the classifier, including the false-positive gate.

Verified in both directions: reverting the two behavioral hunks fails the
end-to-end test on the missing `runtime.error`; removing the flag gate fails
the false-positive test.

Full suite: **3896 passed, 20 skipped, 0 failed** (333 files).
`tsc -b` and `tsc -p tsconfig.server.json` clean. `oxlint .` exits 0 with the
same 31 pre-existing warnings as `main` — no delta.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Claude authentication failures.
  * Signed-out sessions now show a setup error instead of displaying authentication messages as assistant responses.
  * Completed turns now clearly indicate when authentication is required.
  * Other API errors, such as service overloads, continue through normal retry handling.

### DIFF
--- a/server/drivers/claude-auth.test.ts
+++ b/server/drivers/claude-auth.test.ts
@@ -2,7 +2,7 @@
 // runner, so they never read or mutate the developer's real credentials.
 import { describe, expect, it } from "vitest";
 
-import { claudeSignedIn } from "./claude.ts";
+import { claudeAuthFailure, claudeSignedIn } from "./claude.ts";
 
 describe("claudeSignedIn", () => {
   it("uses the CLI's machine-readable auth status", async () => {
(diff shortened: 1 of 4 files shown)


milind-soni:
Reviewed and verified against current main in an isolated fake-engine server. A signed-out turn now produces an activity error with tool.setup=true (and no fabricated assistant reply), which routes to the existing sign-in UI. Focused driver/auth tests and typechecking pass; all GitHub test/build jobs are green. The separate Vercel status is a fork-deployment authorization requirement, not a build failure.

Issue #558, opened 2026-08-29
Proposal: a minimal locale seam, landed one screen at a time


ecogetaway:
### Problem

#453 asks for multi-language UI support and is still open. Two attempts have
already been made and neither landed:

- #311 was closed because it patched compiled JavaScript rather than the
  application sources.
- #357 took the right approach — a locale provider with English fallback,
  persisted selection, interpolation, and a full zh-CN catalog — passed several
  review rounds, and was then closed by its author because the branch could not
  stay merge-able against a fast-moving `main`.

The second one is the interesting failure. The approach was sound; the *shape*
of the change was the problem. A branch touching 47 files cannot survive this
repo's commit rate, no matter how good it is.

### Proposal

Land the seam first, then convert screens one PR at a time.

**PR 1 — the seam only** (small, no UI conversion):

- `src/lib/i18n.ts`: a `LocaleProvider` + `useT()` returning a `t(key, vars)`
  with English fallback and `{placeholder}` interpolation. Plain React context,
  **no new runtime dependency** — this is well under the "thirty lines of code"
  bar in CONTRIBUTING, and `i18next`/`react-intl` would be the wrong altitude
  for this codebase.
- Locale selection persisted through the existing store.
- One catalog file, `en`, covering a single screen (App Settings → General
  seems the natural first, since it also hosts the language selector).
- `pnpm typecheck && pnpm test` green; before/after screenshots.

**Subsequent PRs — one screen each.** Each is independently reviewable,
independently revertible, and small enough to merge before it conflicts.
Untouched screens keep rendering English via the fallback, so `main` is never
half-broken.

### On languages

I'd maintain **Hindi** as the first non-English catalog, and keep it current as
the UI evolves rather than dropping it after one PR. The point of the seam
though is that it unlocks every language, not one — and #311/#357 between them
show there are already people willing to bring zh-CN and others once there is
something to plug into.

### Before I write anything

Two questions, since CONTRIBUTING asks for agreement on approach first:

1. Is a locale seam something you want in the codebase at this stage, or is it
   better parked until the UI settles?
2. If yes — does the incremental shape above look right, and is App Settings the
   screen you'd want converted first?

Happy to defer entirely if @08820048 wants to pick this back up; they said in
#357 they planned to return with smaller PRs, and they've done the most work
here already. I'd rather coordinate than duplicate.

Issue #453, opened 2026-08-25
Request: Add multi-language UI support


qiyuechen0929:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Chat UI (src/)

### Problem

Hi, thanks for creating OpenMausBot!

I would like to suggest adding multi-language support for the UI.

Currently, the interface is mainly in English, which may be difficult for some users. It would be great if users could switch between different languages (for example, English, Chinese, etc.).

A simple language selection option in the settings would make OpenMausBot more accessible to users worldwide.

Thanks for your great work!

### Proposal

Hi, thanks for creating OpenMausBot!

I would like to suggest adding multi-language support for the UI.

Currently, the interface is mainly in English, which may be difficult for some users. It would be great if users could switch between different languages (for example, English, Chinese, etc.).

A simple language selection option in the settings would make OpenMausBot more accessible to users worldwide.

Thanks for your great work!

PR #885, opened 2026-09-07
Translate Settings → General, the screen that hosts the language picker

alisson-acioli:
Closes nothing on its own — it is the next screen of #558, and moves #453 forward.

## The bug

Pick **Português (Brasil)** in Settings → General and nothing changes. Same for every
other language in the list. The report reads like a broken setting, so I checked the
seam first: `PATCH /api/config {language}` persists, `config.language` comes back,
`App.tsx` calls `setLocale` and bumps the epoch. All of it works.

The catalog is what does not reach the screen. 91 strings, spent on approval cards and
the no-engines screen — two surfaces a person may never see on the day they change the
setting. The Settings dialog they are *looking at while they change it* was entirely
hardcoded English, so the honest UX of the feature was: choose a language, watch
nothing happen, conclude it is broken.

#558 said App Settings → General was the natural first screen to convert, "since it
also hosts the language selector". It was the one screen that never got converted.

## The change

The modal chrome (title, search box, section nav, close) plus the whole General
section: profile, skin, channel turns, language, tool calls, updates, diagnostics,
analytics. 55 keys under `settings.*`, translated in all eight packs.

| pack | before | after |
| --- | --- | --- |
| de | 91/91 | 146/146 |
| pt-br | 91/91 | 146/146 |
| es · fr · hi · ja · zh | 16/91 | 71/146 |

Two things worth a reviewer's eye:

**`SECTIONS` carries a `labelKey`, not a `label`.** `t()` reads the active pack at call
time, so a label resolved in that module-scope array would be frozen to whatever
language the app booted in — and this is precisely the screen where the language changes
under the reader's cursor. The nav, the header and the search filter all resolve the key
during render instead. The keyword index stays English on purpose: it is a search aid,
and translating it would drop matches for people who type the English word.

**The channel-turn range message moved to the call site.** `parseRoomTurnTimeoutMinutes`
stays pure and keeps its English constant (and its unit test); both of its rejection
branches mean "out of range", so the component prints the catalog sentence instead of
the parser's.

Out of scope, still English by fallback: the other six sections, the skin names and
their descriptions (`lib/skins.ts` is data, not copy in this component), and onboarding.
`settings.language.subtitle` still says only part of the interface is translated,
because that is still true.

## Verification

Real app, isolated data dir and home, `server/index.ts` + `vite` on their own ports, no
touching of my own instance:

- Before: pt-BR selected, the dialog stays English end to end — the reported bug,
  reproduced.
- After: the app boots in pt-BR (Configurações · Geral · Perfil · Tema · Mecanismos ·
  Acesso remoto · VM local · Uso) and the General cards read in Portuguese.
- Switching pt-BR → 日本語 with the dialog open reflows it live, no reload: 設定 · 一般 ·
  エンジン · 言語 · ツールの呼び出し. That is the module-scope freeze the `labelKey` change
  prevents.
- Search still filters on the translated label (`エンジン` matches the Engines row) and
  the empty state interpolates the query: 「zzzz」に一致するものはありません.

(146 strings, 8 languages, valid), full `vitest run` — 333 files, 3891 passed. The new
`SettingsModal.i18n.test.ts` renders the screen twice through `setLocale` and fails if a
label is ever frozen at module scope (verified by mutating one and watching it go red).

Brazilian Portuguese is mine and reviewed line by line. The other seven are drafts in the
sense `docs/localization.md` means it — reviewed for terminology and placeholders against
the existing packs, but a native speaker's correction on any of them is welcome and easy
to land, since every value is one line in one file.

Base: `9c681f44` (rebased onto 0.1.61). Head: `alisson-acioli:feat/settings-general-i18n`.

## Summary by CodeRabbit

- **New Features**
  - Localized the Settings interface, including navigation, labels, profile options, themes, room-turn limits, language selection, tool calls, updates, diagnostics, analytics, and validation messages.
  - Added Settings translations for German, Spanish, French, Hindi, Japanese, Portuguese (Brazil), Simplified Chinese, and English.
  - Settings labels and messages now update dynamically when the language changes, with English fallback support.
  - Improved update status messaging for preparation, installation, retry, and restart states.

- **Tests**
  - Added coverage for dynamic language switching and fallback behavior in the Settings modal.


alisson-acioli:
Rebased onto `9c681f44` (0.1.61). #896 landed in the two files this stack touches most, so the
merge was not mechanical — the new updater copy is translated rather than left behind:

- **`UpdatesRow` (#885)** picked up `preparing`, `handed-off`, the `installMode: "handoff"` split
  and `retryable === false`, so the catalog gained `settings.updates.startingDownload`,
  `.preparing`, `.readyInstall`, `.openingTerminal`, `.restarting`, `.handedOff`, `.quitReopen`,
  `.install`, `.preparingShort`, `.opening` and `.restartingShort`. Two strings #896 reworded —
  `Check failed:` → `Update failed:`, and the diagnostics subtitle now naming the desktop and
  updater logs — were updated in English and in all seven packs rather than silently kept at the
  old wording.
- **`updateLabel` (#886)** gained `preparing`, the handoff variants of `downloaded` and
  `installing`, and the updater's own `state.message` passthrough: `sidebar.update.preparing`,
  `.readyInstall` and `.openingTerminal`.

#896's new tests pass unchanged against the translated code, because English is still what the
catalog returns by default: `"Preparing update…"`, `"Version 0.2.0 ready — install"` and
`"Opening a terminal…"` come out of `updateLabel` exactly as `SidebarProfileMenu.test.ts` expects.

All eight packs are still complete (557/557 after the new keys), and a full `vitest run` on the
rebased tip is green: **338 files / 3917 tests, exit 0**, including `electron/*` and the
updater-coordinator tests that arrived with #896.

milind-soni:
Reviewed the General/settings-chrome translations and pushed a small cleanup for the review finding: selection now follows the rendered search results, so a language change cannot leave the panel on a section absent from its filtered rail. Verified in an isolated real browser: picker switches English→Portuguese→English live; search for mecanismos in English, then change to Portuguese without editing the query, selects Engines. The same check fails on the previous implementation and passes with this fix. Typecheck, catalog validation (8 languages), and 30 targeted tests pass. Waiting for the updated platform CI before merging.

milind-soni:
Updated head 71cee9ab now passes every GitHub platform/build check, including Windows. The combined current-main integration also passed typechecking, production renderer build, catalog validation, and 4,216 tests. Live English/Portuguese switching and the unchanged-query regression were verified in an isolated browser. Merging the translation foundation; later stacked translation PRs still need their incremental review.

PR #886, opened 2026-09-07
Translate the sidebar

alisson-acioli:
**Stacked on #885, and targeted at `main`.** That branch's commit rides along here — review
only the last commit (`aec8e85`). They share `en.json` and `source-hashes.json`, so they
cannot land independently without a conflict; if #885 is squash-merged, say the word and I
rebase this one on top of the squashed commit.

## Why the sidebar next

#885 answered the report ("I pick Português and nothing changes") on the screen where the
choice is made. This is the screen the person then looks at all day: the rail is on display
in every view, in every session, so it is where an English string is most visible.

## What is in it

139 keys covering the rail and everything that lives in it:

- bot and channel rows, including the previews (`Working…`, `You: …`, `The team is working…`)
- both context menus, the section picker and Move to context
- New Channel, Archived bots, the density menu, the "+" menu, search and its empty state
- the section headers (their titles and the reorder announcement)
- the Tools menu, the profile menu with every updater state, and the phone button's device status

All eight packs carry them. de and pt-BR remain at 100%; es · fr · hi · ja · zh go from 71
to 210 of 285 strings.

## Three decisions a reviewer should check

**Built-in section names go through a helper, not the lib.** `sidebarSectionLabel` in
`lib/sidebar-layout.ts` stays pure and English — its test still pins `"Channels"`. The
component maps the four built-in ids to catalog keys and falls through to that helper for a
section someone named themselves, because a user's own section name is their text and must
never be translated.

**`botConfirmCopy` and `deriveSidebarPhoneStatus` keep their signatures.** Both are exported,
pure and tested against exact English. They compose catalog strings now, and every existing
assertion still passes: the default locale is English, so `botConfirmCopy("delete", "Willow")`
is still `"Delete Willow?"` — the name comes from the argument, never from a catalog value.

**Plural counts are two keys, not a plural engine.** `1 bot` / `{count} bots` reads as
`sidebar.newChannel.createOne` and `…createMany`. `t()` has no plural rules and does not need
them for two-form languages; adding an engine for six strings would be the wrong altitude for
this catalog.

## Verification

Isolated instance again (own data dir, own home, own ports; my real instance untouched),
`language: "pt-br"` on disk:

- The rail boots in Portuguese: Buscar · BOTS · Ferramentas · Você, and the "+" menu reads
  Novo bot · Novo canal · Exportar backup · Equipes.
- Right-clicking a bot gives the whole menu translated: Fixar · Tornar chefe de gabinete ·
  Mover para seção · Marcar como não lida · Editar perfil · Duplicar · Copiar ID da conversa ·
  Arquivar · Excluir.
- Switching to 日本語 **with both menus open** reflows every one of them live, no reload:
  検索 · ボット · ツール · あなた · ピン留め · セクションへ移動 · 会話 ID をコピー · 削除.

That last check caught a real bug before this branch existed as a commit:
`SidebarMoreMenu` renders its `label` as visible text *and* as the aria-label, and my first
pass resolved it into a separate variable that only reached the aria-label — the Tools row went
blank. A default parameter is evaluated per call, so `label = t("sidebar.tools")` is both
correct and smaller. The unit tests were green with the row blank; the running app was not.

(285 strings, 8 languages, valid), full `vitest run` — **335 files, 3895 passed**, including the
existing `SidebarBotListItem`, `SidebarPhoneButton`, `SidebarProfileMenu` and `sidebar-layout`
suites that assert the English copy. New `Sidebar.i18n.test.ts` renders a row in pt-BR and ja and
checks the confirmation copy in three languages.

Brazilian Portuguese is mine and reviewed line by line; the other seven are model-assisted drafts
in the sense `docs/localization.md` means it, checked for placeholders and against the terminology
the existing packs already use. A native speaker's correction is one line in one file.

Base: `9c681f44` (rebased onto 0.1.61). Head: `alisson-acioli:feat/sidebar-i18n`.

## Summary by CodeRabbit

- **Localization**
  - Added translations for settings, sidebar navigation, bot management, backups, updates, and remote-access statuses across supported languages.
  - Localized settings modal, sidebar, profile menus, status messages, tooltips, confirmations, and room timeout controls.
  - Settings search now matches translated section names.
  - Added localized updater status messages and improved update button availability for in-progress or non-retryable states.

- **Tests**
  - Added coverage confirming settings and sidebar content update correctly when switching languages.


alisson-acioli:
Rebased onto `9c681f44` (0.1.61). #896 landed in the two files this stack touches most, so the
merge was not mechanical — the new updater copy is translated rather than left behind:

- **`UpdatesRow` (#885)** picked up `preparing`, `handed-off`, the `installMode: "handoff"` split
  and `retryable === false`, so the catalog gained `settings.updates.startingDownload`,
  `.preparing`, `.readyInstall`, `.openingTerminal`, `.restarting`, `.handedOff`, `.quitReopen`,
  `.install`, `.preparingShort`, `.opening` and `.restartingShort`. Two strings #896 reworded —
  `Check failed:` → `Update failed:`, and the diagnostics subtitle now naming the desktop and
  updater logs — were updated in English and in all seven packs rather than silently kept at the
  old wording.
- **`updateLabel` (#886)** gained `preparing`, the handoff variants of `downloaded` and
  `installing`, and the updater's own `state.message` passthrough: `sidebar.update.preparing`,
  `.readyInstall` and `.openingTerminal`.

#896's new tests pass unchanged against the translated code, because English is still what the
catalog returns by default: `"Preparing update…"`, `"Version 0.2.0 ready — install"` and
`"Opening a terminal…"` come out of `updateLabel` exactly as `SidebarProfileMenu.test.ts` expects.

All eight packs are still complete (557/557 after the new keys), and a full `vitest run` on the
rebased tip is green: **338 files / 3917 tests, exit 0**, including `electron/*` and the
updater-coordinator tests that arrived with #896.

PR #887, opened 2026-09-07
Translate the chat column

alisson-acioli:
**Stacked on #885 and #886, and targeted at `main`.** Those two commits ride along here —
the only new one is `8dac97b`. All three touch `en.json` and `source-hashes.json`, so they
land in order; say the word after a squash-merge and I rebase.

## Why this screen

#885 fixed the report on the screen where the language is chosen; #886 did the rail that is
always on display. This is the column people actually read.

## What is in it

98 keys:

- bubbles and their actions — copy, edit, reply, pin/unpin, regenerate, the `‹ 2/3 ›` version
  switcher, the collapse/expand of a long message
- day separators, the webhook task card, the execution timeline, the pinned banner, the
  empty state, "show earlier/later messages", the jump-to-latest pill
- the header: profile, find (⌘F), stop, bot's computer, inspector, and the usage and
  working-folder chips with their tooltips
- the find bar, the reply quote, and the read-aloud button

de and pt-BR stay complete; es · fr · hi · ja · zh go from 210 to 308 of 383 strings.

## Two things worth a reviewer's attention

**`FALLBACK_LABELS` holds keys now.** `live-activity.ts` writes the one line under the
mascot ("Running a command", "Searching the web"). Its table is built once at import time,
so an English label resolved there would be frozen to the language the app booted in — the
same trap the sidebar's section list had. The table carries `LocaleKey`s and resolves them
per call. The server's own narration (`tool.spoken`) is deliberately left as it arrives: it
is a sentence the server wrote, not a key this catalog can answer.

**The transcript is memoized, so it needed a memo key.** `MessagesList` is `memo(...)`, and
none of its props change when the language does. Switching language with a conversation open
left every settled bubble — day separators included — in the old language while the header
around them switched. `activeLocale()` is new in `lib/i18n` and is passed as `locale`; it is
not read inside the component, and the prop's comment says exactly that. Its test pins it to
the locale that survives fallback (`de-AT` → `de`, unknown → `en`).

Four pure helpers keep their signatures, their tests, and their English defaults:
`liveActivityLabel`, `usageDetail`/`costCaption`, `describeRun`, `replyAuthor`/`replySnippet`.

## Verification

Isolated instance (own data dir, home and ports), `language: "pt-br"` on disk:

- header and transcript read in Portuguese: Abrir o perfil do agente · Buscar na conversa (⌘F) ·
  Computador do bot · Inspetor — eventos de execução e protocolo bruto desta conversa ·
  Copiar mensagem · Gerar a resposta de novo · Responder · Fixar esta mensagem no topo da
  conversa · "Conversa com Ivy" · **Hoje 23:33**
- the find bar: "Buscar nesta conversa", Resultado anterior · Próximo resultado · Fechar a busca
- switching language live with the conversation open now reflows the transcript too:
  今日 → Hoje → Heute on the day separator, no reload. That is the memo-key fix; before it,
  the separator stayed in the previous language while the header switched — found by driving
  the real app, not by a unit test.

(383 strings, 8 languages, valid), and `vitest run`: **335 files / 3899 tests pass**. One
server test (`rejects a sidebar section Chief collision…`) hit its 20s timeout in a full run on
a loaded machine and passes on its own — it does not touch the renderer, and the same file
passes 197/197 in isolation. Reporting it rather than hiding it.

Brazilian Portuguese is mine and reviewed line by line; the other seven are model-assisted
drafts as `docs/localization.md` describes, checked against the terminology already in the packs.

Base: `9c681f44` (rebased onto 0.1.61). Head: `alisson-acioli:feat/chat-i18n`.

## Summary by CodeRabbit

- **New Features**
  - Added localized labels, tooltips, accessibility text, status messages, errors, and confirmations across chat, settings, sidebar, replies, activity, usage, and speech controls.
  - Added expanded translations for English, German, Spanish, French, Hindi, Japanese, Brazilian Portuguese, and Simplified Chinese.
  - Language changes now update visible chat content immediately.

- **Tests**
  - Added coverage for translated settings, sidebar, chat content, locale switching, and default English behavior.


alisson-acioli:
The red job here is a flaky Box test in `server/index.test.ts`, not this change — the evidence,
in case it is useful beyond this PR:

- **This run:** `harness HTTP API > keeps a bot when its hidden Box exists or the provider cannot
  prove it absent` failed on `ubuntu-latest` with `Matcher did not succeed in time` on the
  `expect.poll` at `server/index.test.ts:2314`. **macOS and Windows passed the same commit.**
- **#886 failed the same test**, also only on ubuntu.
- It is not specific to my branches: run `34049450343` (`feat/research-command`) failed on
  *windows* with three Box tests in the same file, and `main` itself (`34060554319`) failed a
  different timing test (`room-chat-wait.e2e`) on windows the same day.
- The file cannot see this PR: `server/index.test.ts` imports nothing from `src/` (no `@/`
  imports at all), and `fileParallelism: false` means no renderer test is running beside it.
  Every string in this PR lives in the renderer.
- Locally the test passes on this branch (3/3 runs, ~8s each), and the full suite is
  335 files / 3899 tests green apart from a *different* load-sensitive test on my own machine,
  which I reported in the PR body.

The poll takes vitest's default 1s budget while the surrounding turns are real HTTP round trips
against a spawned harness, so a slow shared runner loses the race. Happy to send a separate
one-line PR giving that poll an explicit timeout if you want it fixed rather than re-run — I'd
rather not smuggle a server-test change into a localization PR.

A re-run of the ubuntu job should be green.

Issue #776, opened 2026-09-04
Rooms should show when a bot answers from a private thread


milind-soni:
Follow-up to #754 (`session_search`), which lets a bot recall its own past threads and rides in room turns too. That is the feature people asked for: a bot mentioned in a team channel should be able to answer about work discussed in its DM.

The gap is disclosure, not capability. When a bot answers in a room using something it recalled from a one-to-one thread, nobody in the room can see that it did.

## Why this is worth building

Recipient misalignment — an assistant answering in a shared space using what it was told privately — is a measured failure mode, not a hypothetical. AgentCIBench reports mean leakage around 68% on exactly this shape. A 2026 study of private-plus-shared agent memory (arXiv:2607.05318) found that adding a shared surface to private memory cut agent-to-agent leaks but pushed visibility violations to 63-90%, and concluded that a vague policy is worse than no rule at all. Both ChatGPT projects and Claude projects solved the same problem by hard partitioning; we are deliberately not doing that, because the partition is the thing users are complaining about. So the honest alternative is to make the crossing visible.

The worst detail of the 2024 Slack AI incident was not the leak, it was that the citation pointed at the victim's own private channel: the interface said the wrong thing about where information came from.

## Proposal

1. When `session_search` or `session_read` returns a hit from a thread that is not the current one, and the current thread is a room, record it on the turn.
2. Render an activity chip in the room: "<Bot> recalled 2 messages from its chat with you", clickable through to the source thread for the user (and only the user).
3. Attach provenance to each recalled item in the model's context — surface and timestamp — so the bot can say where something came from when asked.
4. First-disclosure confirmation: the first time a given bot recalls private material into a given room, ask the user once, and remember the answer per room.

## Not proposed

Blocking recall in rooms. That removes the feature. The point is that the user sees the crossing happen and can stop it, not that it cannot happen.

PR #901, opened 2026-09-07
feat(rooms): show when a bot answers from a private thread

Viewofmind:
Closes the visible half of #776.

## The gap

#754 gave bots recall, and it rides in room turns — a bot mentioned in a channel can answer about work discussed in its DM. That is the feature people asked for, and this PR does not touch it. The gap the issue names is **disclosure**: when a bot answers in a room using something it recalled from a private thread, nobody in the room can see that it did.

## What this does

When `session_search` or `session_read` returns a hit from a thread that is not the current one, **and the current thread is a room**, an activity chip lands in that room:

> Cia recalled 2 messages from its private chat with you

It reuses the chip mechanism room turns already use for "finishing another conversation", so there is no new surface to render.

**Once per (room, source thread).** A bot that searches three times in one turn announces each private thread once, not once per search. The ledger lives in `server/recall-disclosure.ts`.

**A 1:1 stays silent.** There is nothing to disclose when the user already owns every thread the bot can reach — a room is the only place a recall can be a disclosure.

## The model is told too

Crossed hits are marked in the tool output — `task "Audit", private to this user` — with a closing line saying the room has been shown the recall and the bot should say where something came from if asked.

That is point 3 of the issue, and it matters on its own: without it a bot cannot answer "where did you get that?" honestly, which is the failure the issue's Slack AI example is about — the interface saying the wrong thing about where information came from.

## One design call

The ledger is in memory, so a restart re-announces. That is deliberate: a duplicate chip is noise, a missing one is the entire bug. If per-room consent lands later (point 4), it will want persistence anyway and can carry this with it.

## Not included

Point 4, the first-disclosure confirmation. It needs persisted per-room state and a card flow; the visible half is worth shipping before the asking half. Also not included: click-through from the chip to the source thread, which wants UI work this PR deliberately avoids.

## Tests

Unit tests for the ledger — first crossing claimed, repeat silent, a new source announced alongside a known one, rooms kept apart, re-announce after forgetting.

An integration test drives the real routes: something said privately, recalled from a room, chip appears once, a second search adds nothing, `session_read` of the same message is already announced, and the same recall from the bot's own 1:1 stays silent.

Full suite on Node 24: 3,914 passed, 0 failures. Typecheck and oxlint clean.

## Summary by CodeRabbit

- **New Features**
  - Room conversations now clearly identify when recalled messages originated from a private conversation.
  - A disclosure notice is shown the first time private content is recalled from each source conversation.
  - Search and read results include clearer labels and explanatory notes for cross-conversation recalls.
  - Repeated recalls avoid duplicating the same disclosure notice.
  - Direct searches within private conversations remain marked as current and undisclosed.

### DIFF
--- a/server/drivers/agents-proxy.ts
+++ b/server/drivers/agents-proxy.ts
@@ -864,13 +864,18 @@ async function callTool(name: string, args: Json): Promise<{ text: string; isErr
     }
     const lines = hits.map((hit) => {
       const when = typeof hit.at === "number" ? new Date(hit.at).toISOString().slice(0, 10) : "";
-      const where = hit.current ? "this conversation" : typeof hit.task === "string" && hit.task ? `task "${hit.task}"` : "an earlier task";
+      const task = typeof hit.task === "string" && hit.task ? `task "${hit.task}"` : "an earlier task";
+      const where = hit.current ? "this conversation" : hit.crossed ? `${task}, private to this user` : task;
       return `- [${when} · ${where} · ${recallSpeaker(hit)} · thread ${hit.threadId} · message ${hit.messageId}] ${hit.snippet}`;
     });
+    const crossed = hits.some((hit) => hit.crossed === true);
     return {
       text:
         `${hits.length} matching message${hits.length === 1 ? "" : "s"} from your earlier conversations (best match first):\n${lines.join("\n")}\n\n` +
-        "These are your own past notes. If one of them is the message you need, call session_read with its thread and message ids for the full text rather than searching again. Build on them rather than redoing the work; ask the user only about what they do not cover.",
+        "These are your own past notes. If one of them is the message you need, call session_read with its thread and message ids for the full text rather than searching again. Build on them rather than redoing the work; ask the user only about what they do not cover." +
+        (crossed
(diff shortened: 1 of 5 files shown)


milind-soni:
Useful visibility improvement, but there is a lifecycle gap before merging: forgetRecallCrossings is only called by tests, despite its comment saying it is used when a room is cleared. Once a source is announced, clearing that room transcript can remove the visible disclosure while the process-wide ledger suppresses every later disclosure from that source. Please wire clearing/deletion into the ledger and add a real-route regression proving recall after clearing announces again. This is disclosure, not a consent/access-control boundary, which should remain explicit.

milind-soni:
Closing as already shipped: the complete #901 patch is present on main through #907 (5b547b3b). Verified by checking that the entire PR diff reverse-applies cleanly to main; no second merge is needed. Thank you for the disclosure work.

Clarification to my earlier lifecycle note: forgetRecallCrossings is not wired into production cleanup, but current task deletion removes the thread and replacement tasks get a new ID. I have not reproduced the same-thread transcript-clear scenario through a current public route, so that should not be treated as a confirmed present-day disclosure bypass. Ledger cleanup remains a follow-up concern rather than a reason to merge this patch twice.

PR #904, opened 2026-09-07
fix(settings): Close settings X works after editing SOUL.md

### LINKED ISSUE #845: Bot Settings dialog: "Close settings" (X) button doesn't close after editing SOUL.md
**Version:** 0.1.57
**OS:** macOS

## Steps to reproduce

1. Open a bot's new Settings dialog (the redesigned one from #833 — Overview/Identity/Soul/Skills/etc. section rail).
2. Go to the **Soul** section and edit the SOUL.md text in the textarea.
3. Click the dialog's own "Close settings" X button (top-right of the inner panel, `aria-label="Close settings"` in `BotSettingsDialog.tsx`).

## Expected

The settings dialog closes.

## Actual

Nothing happens — the button click has no effect and the dialog stays open. The only way to dismiss it is via a different, outer close control (top-left), not the dialog's own visible close button.

## Possible cause

`BotSettingsDialog.tsx`'s keydown handler disables Escape handling whenever it detects a nested `[role="dialog"], [role="alertdialog"]` element still mounted inside the settings dialog subtree (there's a drift-warning dialog and a rollback `ConfirmDialog` that can appear in that same tree). If such a nested dialog element is left mounted/invisible after an edit, it's plausible it's also intercepting or blocking the X button's click, not just Escape — worth checking whether a stray dialog node persists in the tree after typing in the Soul textarea.

Editing SOUL.md itself seems necessary to trigger it — haven't confirmed whether other sections (e.g. Identity) hit the same issue.

XiaoZ-0218:
## Summary
- Fixes the Bot Settings dialog Close settings (X) button becoming unclickable after editing SOUL.md (#845).
- Root cause: the right-hand header lacked `shrink-0` and the section scroller lacked `min-h-0`, so Soul's tall textarea could flex-shrink the header hit box while the X icon still painted — clicks missed the button; the outer backdrop still dismissed.
- Also only yield Escape/Tab to *visible* nested dialogs, matching the existing focusable filter.

## Test plan
- [ ] Open a bot's Settings dialog → Soul
- [ ] Edit the SOUL.md textarea (add several lines)
- [ ] Click the dialog's top-right Close settings (X) — dialog should close
- [ ] Re-open, edit again, press Escape — dialog should close
- [ ] With a nested dialog open (e.g. skill review), Escape should still close only the nested layer

Fixes https://github.com/milind-soni/OpenMausBot/issues/845

## Summary by CodeRabbit

* **Bug Fixes**
  * Escape and Tab interactions now work correctly when nested dialogs are hidden or have no visible area.
  * Prevented tall text areas and notification banners from shrinking the settings header.
  * Improved close-button behavior within the settings dialog.

### DIFF
--- a/src/components/BotSettingsDialog.tsx
+++ b/src/components/BotSettingsDialog.tsx
@@ -190,7 +190,11 @@ export function BotSettingsDialog({ bot }: { bot: Bot }) {
       // review, the model picker's popover, a computer warning) owns Escape
       // and Tab while it is up: Escape closes only that layer, and the focus
       // trap below must not pull focus back out of it.
-      if (dialog?.querySelector('[role="dialog"], [role="alertdialog"]')) return;
+      // Only a *visible* nested dialog owns Escape/Tab. A hidden or
+      // zero-size leftover (display:none, empty hit box) must not trap
+      // the settings dialog's own dismiss paths.
+      const nested = dialog?.querySelector<HTMLElement>('[role="dialog"], [role="alertdialog"]');
+      if (nested && nested.getClientRects().length > 0) return;
       // BotInstructionsDialog portals to document.body, so it is not in this
       // subtree: a key pressed with focus outside this dialog belongs to
       // whatever holds focus, never to us.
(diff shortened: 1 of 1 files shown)


milind-soni:
Reviewed against current main and exercised the real BotSettingsDialog/StoreProvider against a disposable fake-engine server. Editing a long SOUL and clicking the center of the X closes the dialog at both 1000×640 and 1000×480; the SOUL edit remains persisted. The header keeps its 26×26 button hitbox. This fixes the flex sizing cause without adding event-order workarounds or a DOM dependency. All GitHub test/build jobs pass; Vercel only asks for fork deployment authorization.

PR #905, opened 2026-09-07
fix(settings): close Bot Settings X after SOUL textarea edit

### LINKED ISSUE #845: Bot Settings dialog: "Close settings" (X) button doesn't close after editing SOUL.md

santhiprakash:
## What changed

- `BotSettingsDialog`: the close X now handles `onMouseDown` with `stopPropagation` and is explicitly `type="button"`. The dialog panel also stops `mousedown` propagation so clicks inside (including the Soul textarea) do not bubble to the backdrop.
- Section nav buttons are explicitly `type="button"`.
- Added a `jsdom` Vitest regression test (`src/components/BotSettingsDialog.test.ts`) that opens settings, selects Soul, edits the SOUL textarea, and clicks the X, asserting the dialog unmounts and `settingsOpen` becomes false.

## Why

Issue #845 reported that the Bot Settings X did not close the dialog after the user focused or edited the SOUL.md textarea. The close relied solely on `onClick`; adding a `mousedown` handler and stopping propagation makes the close more robust and prevents the backdrop `onMouseDown` from swallowing or racing the close action.

## How it was verified

- `pnpm typecheck` passes.
- `pnpm vitest run src/components/BotSettingsDialog.test.ts` passes (with `jsdom` environment, Node 24).
- Manual Playwright fixture `src/testing/bot-settings.tsx` confirms `dialog count after close: 0` after editing the Soul textarea.

## Screenshots (UI changes)

No visual change.

## Checklist

- [x] `pnpm typecheck` and the focused `pnpm vitest run src/components/BotSettingsDialog.test.ts` pass locally  (the full `pnpm test` suite requires Node >=24 and was not run in the agent environment)
- [x] UI behavior change comes with a regression test
- [x] No `dist-server/` edits

Fixes #845

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where the Bot Settings dialog could remain open after editing the SOUL text area and closing it.
  - Improved dialog interactions to prevent unintended event propagation and ensure navigation and close controls behave correctly.

- **Tests**
  - Added regression coverage for closing the Bot Settings dialog after editing settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "electron-builder": "^26.15.3",
     "electron-updater": "^6.8.9",
     "esbuild": "^0.28.2",
+    "jsdom": "25.0.1",
     "oxlint": "1.80.0",
     "tailwindcss": "^4.1.11",
     "typebox": "1.3.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -87,6 +87,9 @@ importers:
       esbuild:
         specifier: ^0.28.2
         version: 0.28.2
+      jsdom:
+        specifier: 25.0.1
+        version: 25.0.1
       oxlint:
         specifier: 1.80.0
         version: 1.80.0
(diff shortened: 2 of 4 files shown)


milind-soni:
Thanks for working on this. We have merged #904 for the same issue: it fixes the flex layout so the visible X retains its actual hitbox. Verified with real browser clicks after editing SOUL at both normal and short window sizes. We should avoid adding the extra mousedown dismissal and jsdom dependency here; the programmatic .click() test does not exercise browser hit-testing. Keeping one layout fix avoids duplicate event paths.

PR #907, opened 2026-09-07
fix(browser): picture the page a bot browsed, not nothing and not its desktop

### LINKED ISSUE #906: Browser turns settle with no picture since the engine swap: the screen gate still lists the Electron surface's tool names, and nothing supplies browser frames
## What happens

On 0.1.61 (macOS), a bot with the built-in browser opens a page successfully — `agent_browser_open`, `agent_browser_screenshot` and `agent_browser_snapshot` all run and are approved, and the bot describes the page correctly — but **nothing reaches the transcript**. The chat shows the screenshot as "Unavailable".

Verified rather than assumed: no image file is written anywhere under `~/.openmausbot`, and every message from that turn has a JSON payload between 199 B and 1 KB — orders of magnitude too small to carry a PNG.

The Browser panel's own copy says *"Pages the bot looks at appear in the chat as screenshots. A live view you can watch and take over is coming next."* The first sentence is not true on 0.1.61, so there is currently no way at all to see what a bot's browser sees.

## Two causes, both from the engine swap

**1. `server/screen-frame-gate.ts` still lists the removed surface's tool names.**

`SCREEN_TOUCHING_TOOLS` has the Electron browser's names — `browser_navigate`, `browser_type`, `browser_screenshot` and so on. agent-browser's tools are `agent_browser_open`, `agent_browser_click`, `agent_browser_screenshot`. None match, so `entry.touched` never becomes true and `finalScreenFrame` returns `null` for every browser turn.

**2. Nothing supplies a browser frame.**

`startScreenPoller` only runs when `previewCapture` is set, and that is assigned solely by the computer branches (Local VM, VPS, Box). The browser's capture lived in `server/drivers/browser-proxy.ts`, which was deleted with the Electron surface (#850) and never replaced. A bot whose only surface is the browser therefore has no frame source at all.

The two combine: a browser-only bot has no camera, and a bot that also holds a computer has one pointed at the wrong surface.

## The second half matters on its own

agent-browser runs its own headless Chrome on the host, not inside the bot's desktop. So for a bot holding **both** a computer and the browser, fixing only (1) would settle a web-browsing turn with a picture of an idle Local VM. That is arguably worse than no picture — the interface would be confidently showing the wrong thing.

## Why it is worth more than a missing thumbnail

The browser system prompt tells every bot:

> "At a sign-in, password, MFA, CAPTCHA, payment-detail, or other protected-input step, stop and ask the user in chat to complete it."

The bot is instructed to hand sign-in to the human, but with no visible surface there is nowhere for the human to take over. That is #894's problem approached from a different side: that thread argues the browser is *detectable*; this is that the handoff has nowhere to happen even where detection is not the obstacle.

#834 (picture-in-picture) does not cover this. It is frontend-only, opened 5 Sep — before the swap — and renders the native view that was removed the next day.

## Reproduce

1. Give a bot the built-in browser on 0.1.61.
2. Ask it to open any page.
3. The page loads and the bot describes it; the transcript settles with no image.

## Environment

OpenMausBot 0.1.61, macOS (Intel), agent-browser 0.36.0.

A PR follows.

Viewofmind:
Closes #906.

## The symptom

A bot opens a page on 0.1.61. `agent_browser_open`, `_screenshot` and `_snapshot` all run and are approved, the bot describes the page correctly — and the transcript settles with **no picture**. No image file is written, and the turn's messages carry payloads of 199 B to 1 KB, far too small for a PNG.

Meanwhile the Browser panel says *"Pages the bot looks at appear in the chat as screenshots."*

## Two causes

**The screen gate still lists the Electron surface's tool names.** `SCREEN_TOUCHING_TOOLS` has `browser_navigate`, `browser_type`, `browser_screenshot`. agent-browser's are `agent_browser_open`, `agent_browser_click`, `agent_browser_screenshot`. Nothing matches, so `entry.touched` never flips and `finalScreenFrame` returns `null` for every browser turn.

**Nothing supplies a browser frame.** `startScreenPoller` runs only when `previewCapture` is set, and that is assigned solely by the computer branches (Local VM, VPS, Box). The browser's capture lived in `server/drivers/browser-proxy.ts`, deleted with the Electron surface and never replaced.

## Why fixing only the first would be worse than the bug

agent-browser runs its own headless Chrome on the host, not inside the bot's desktop. For a bot holding **both** a computer and the browser, adding the tool names alone would settle a web-browsing turn with a picture of an idle Local VM — the interface confidently showing the wrong room. I wrote that version first; it is why this PR is shaped the way it is.

So the poller now holds both captures and pictures **whichever surface the touching tool acted on**. A browser tool gets the page, a computer tool gets the desktop, and a status read redirects nothing.

## The capture

`agentBrowserFrame()` runs the pinned binary with the **same session env as the MCP mount** (`AGENT_BROWSER_SESSION`, `AGENT_BROWSER_RESTORE`, the encryption key), so it photographs the daemon the bot is already driving rather than launching a second browser beside it.

A 10-second cap: the poller runs next to a live turn, and a hung browser must not hold the fold open. A missing picture beats a stuck turn.

Read-only browser tools stay out of the gate by the rule already applied throughout that list — `_snapshot`, `_read` and `_get_text` return text, and the waits change nothing the action before them did not already count.

## Overlap with #899

#899 touches `server/browser-engine.ts`, its test and `server/index.ts` as well, but for packaging — bundling the pinned binary and Chromium into desktop targets. Different intent, same files. Happy to rebase onto it whenever it lands; nothing here depends on how the binary arrives, only on running the one already resolved.

This does not overlap #834, which is frontend-only and predates the swap.

## Tests

`screenSurfaceForTool` — browser tools to the browser under every spelling, everything else to the computer, and the case that motivated it: a browsed page must not be illustrated with an idle desktop.

The gate — agent-browser's acting tools now count; its read-only tools and waits still do not.

`agentBrowserFrame`, driven by a fake binary — returns the picture the browser wrote, carries the mount's session env, fails with the browser's own reason, refuses to invent a picture the browser never wrote, and gives up on a hung browser rather than holding the turn open.

Full suite on Node 24: 3,924 passed, 0 failures. Typecheck and oxlint clean.

## Summary by CodeRabbit

- **New Features**
  - Added browser-specific screenshots for web-based tasks, including timeout and failure handling.
  - Screen updates now reflect whether the bot interacted with the browser or computer.
  - Room conversations now clearly announce when a bot recalls messages from a private conversation.
  - Repeated recalls from the same source are announced only once, while recalls within the private conversation remain undisclosed.
  - Saved model selections are now respected when creating bots.
  - Login sessions now remain active during continued use, and pairing details include a public URL.

### DIFF
--- a/server/browser-engine.test.ts
+++ b/server/browser-engine.test.ts
@@ -1,8 +1,9 @@
 import { createHash } from "node:crypto";
-import { chmodSync, mkdirSync, mkdtempSync, readFileSync, statSync, writeFileSync } from "node:fs";
+import { spawn } from "node:child_process";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, statSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { delimiter, join } from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 import {
   agentBrowserFrame,
@@ -20,7 +21,12 @@ import { removeTempDir } from "./testing/cleanup.ts";
 
 const posix = process.platform !== "win32";
 const scratch: string[] = [];
+vi.mock("node:child_process", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:child_process")>();
+  return { ...actual, spawn: vi.fn(actual.spawn) };
+});
 afterEach(async () => {
+  vi.mocked(spawn).mockReset();
   for (const dir of scratch.splice(0)) await removeTempDir(dir);
 });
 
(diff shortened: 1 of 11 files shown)


Viewofmind:
Heads-up: this is incomplete, and I am the one who left the hole. Testing on 0.1.62 the picture still does not appear.

`screenTouchingTool` strips only `^mcp__.+?__`. That covers Claude's `mcp__browser__agent_browser_open`, but the desktop's own approval cards — and the drivers behind them — spell the same tool `browser__agent_browser_open`, the server segment without the `mcp`. That name falls through both branches, so `touched` never flips and a browser turn still settles with no frame.

Measured on 0.1.62:

```
browser__agent_browser_open        false   <- the spelling the app actually sends
browser__agent_browser_screenshot  false
mcp__browser__agent_browser_click  true
agent_browser_open                 true
computer__screenshot               false   <- not only the browser
computer__computer_status          false   (correct)
```

The last two matter beyond this PR: the computer server's tools carry the same spelling, so `computer__screenshot` was being missed the same way, before any of the browser work.

Fix is one line — strip any leading `<segment>__` rather than only `mcp__` — plus tests using the spellings the app really emits, which is what I should have checked here instead of the two already in the test file. PR follows once the full suite finishes.

Sorry for the extra round trip on a release you had already cut.

PR #920, opened 2026-09-07
fix(screen): match the server__tool spelling the desktop actually sends

Viewofmind:
Follow-up to #907, which I left incomplete. Testing on 0.1.62, a browser turn still settles with no picture.

## What is wrong

`screenTouchingTool` strips `^mcp__.+?__` — Claude's spelling for a mounted tool. But the desktop's own approval cards, and the drivers behind them, spell the same tool `browser__agent_browser_open`: the server segment without the `mcp`. That name matches neither branch, so `touched` never flips and `finalScreenFrame` returns `null`.

Measured on 0.1.62, before this change:

```
browser__agent_browser_open        false   <- the spelling the app sends
browser__agent_browser_screenshot  false
mcp__browser__agent_browser_click  true
agent_browser_open                 true
computer__screenshot               false   <- not only the browser
computer__computer_status          false   (correct)
```

**The fifth line is the one worth noticing.** The computer server's tools carry the same spelling, so a Box or Local VM turn reported as `computer__screenshot` has been missed the same way — before any of the browser work, and unrelated to it.

## The change

Strip every leading `<segment>__` instead of only `mcp__`. A bare tool name never contains a double underscore, so nothing else moves: read-only tools and waits stay excluded by the same rule whatever prefix they arrive with, and `screenSurfaceForTool` keeps the computer/browser disambiguation you added on merge.

## On how this happened

The user's approval card in front of me read `browser__agent_browser_open`, and I wrote #907's tests against the two spellings already in the test file instead. Both the code and its tests then shared the same wrong assumption, so a green suite proved only that the mistake was consistent. The names here come from the running app.

Full suite on Node 24: 4,276 passed, 0 failures. Typecheck and oxlint clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved screen activity detection for browser and computer tools, including alternate tool-name formats.
  - Browser interactions now correctly recognize when a screenshot is required.
  - Computer screenshots are correctly associated with the computer screen.
  - Read-only tools remain excluded from screen activity detection.
  - Improved handling of both legacy and namespaced tool names while rejecting invalid desktop tool formats.


milind-soni (commented):
Code review complete at 1d4605ebe02ed9617f5f445eb1e01d7da9763bcf: no actionable correctness/security findings. This is a focused fix for the mounted server__tool names; it merges cleanly into main 7f0c9304, and all 71 screen-gate tests plus server typecheck pass on the integrated tree. An isolated create/send/wait/messages smoke also passes (not a claim of native screenshot rendering).

The remaining merge prerequisite is the original author's Signed-off-by trailer, required by CONTRIBUTING.md. Please sign off your commit and push the updated branch; I cannot add that certification on your behalf.

milind-soni:
Updated against current main in e735065c. The namespace matcher now recognizes desktop server__tool names while preserving older MCP spellings, including local_vm and computer-use. Touch detection and surface routing share that matcher; computer/browser identity stays explicit so similarly named tools cannot attach the wrong screen.

Verification: 78 focused tests pass, including the isolated real-server screenshot-settlement fixture now emitting browser__agent_browser_screenshot. That checks frame persistence before a queued follow-up and rejection of a late frame after thread deletion. Server typecheck, lint, and diff checks pass. Added namespace length/characters, legacy compatibility, read-only exclusions, and surface-collision regressions. No live user data, native browser, or VM was used. Waiting for this head’s CI before merge.

milind-soni (approved):
Reviewed final head e735065c3e7ae5e1176740e53897d753273abe24. The shared matcher handles desktop server__tool names without regressing legacy mcp__ namespaces or computer/browser collision routing. 78 tests passed, including the isolated real-server final-screen settlement fixture now emitting the exact desktop alias; server typecheck and lint passed. This verifies name routing and frame persistence, not a new browser/VM feature. Ready once fresh CI and review checks pass.

PR #924, opened 2026-09-07
Translate the bot settings dialog

alisson-acioli:
Stacked on #923 — the diff below the last commit is that PR. Base is `main`.

## The gap

Every one of the bot settings dialog's twelve sections was still English in
every language: the section rail and its search, Overview, Identity and its
avatar card, Soul, Skills, Memory, Routines, Access, Model, Permissions, Voice
& alerts, History and Usage.

## What changed

215 keys across all 8 packs.

**The section rail carries keys.** `BOT_SECTIONS` is a module-scope array — a
rendered label there would freeze the language the module was imported in. It
now holds `labelKey`s and the dialog resolves them while it renders, the same
shape the app settings rail already uses.

Search matches the translated label *plus* the English keywords, and the
keywords stay English deliberately: they are what someone types, and the docs
and the CLI use the English term. Driving the dialog in pt-BR, "permiss" finds
**Permissões** and "workspace" still finds **Acesso**.

**`runStatusLabel` moved into `schedule-label`.** The calendar (#923) and the
Routines section both needed it; one definition beats two that drift.

**`BotProfileAvatarCard` came along.** It is only rendered by Identity, and
leaving it behind meant an entirely English card ("Reset mascot", "Upload
image", "SHAPE") sitting inside a translated section.

## What stays English, and why

Overview's body sentences — *does* / *can reach* / *won't* — the prompt
preview's section labels, and the history summaries are built by
`server/bot-overview.ts` and `server/system-prompt.ts`. Same boundary the
earlier PRs drew, but worth calling out: this is the largest block of
server-side copy visible on one screen so far. In pt-BR the Overview reads:

> Visão geral · Atlas · **Faz** · Nada agendado ou aprendido ainda. · **Pode
> alcançar** · Works in its private workspace. · **Não fará** · Command
> approvals use Ask mode…

Translating those means moving the sentences into the catalog on the server
side and having the renderer resolve keys — a real change with its own
migration, so it belongs in its own PR rather than smuggled into this one. Happy
to take it next if you want it.

## Testing

- `tsc -b`, `oxlint`, `generate-locale.mjs --check` — clean; all 8 packs at
  100% (2085 strings).
- Full suite: **358 files, 4289 tests passed**, 1 file / 39 tests skipped.
- `sections.i18n.test.ts` asserts every rail entry is a real catalog key, that
  the rail follows a language change, that the English keywords survive, and
  that History renders translated.
- Driven against a running instance. pt-BR rail and sections:

[...]

PR #923, opened 2026-09-07
Translate Automations, the last English screen

alisson-acioli:
Stacked on #921 — the diff below the last commit is that PR. Base is `main`.

## The gap

Automations was the last screen still entirely English in every language: the
calendar and its grid, the event editor, the event details, the paused-routines
list, the calendar sidebar and mini month, and the routine run card that also
appears in chat.

## What changed

210 keys across all 8 packs, and three helpers that could not survive a
translation the way they were written.

**`scheduleSentence()` rewrote its own output with English regexes.** It called
`intervalLabel()` for the short form and then reshaped the result — `"Every 5
min"` → `every 5 minutes`, `1 hr` → `1 hour`, `N hr` → `N hours`. In any other
language those patterns match nothing and the sentence comes back as the short
label. Each shape is now its own key, with separate singular and plural hour
keys so "1 hour 30 minutes" doesn't become "1 hours".

**A run status was read as logic.** The calendar rendered
`status.replace("waiting", "needs you")` — the same labels-as-logic shape as
`teamMapStatus()` in #921. It's now `runStatusLabel()`, keyed off the stable
status string.

**The run card's `COPY` and `GOAL_COPY` are module-scope maps**, evaluated once
at import. They now hold `labelKey`s and the card resolves them while it
renders.

## Dates follow the app language

`niceTime`, `niceDate`, `whenLabel`, the hour gutter and `calendarRangeLabel`
passed `[]` to `toLocaleString`, so they followed the OS rather than the app —
the same fix main made in the computer panel. They use `activeLocale()` now.

The same-month range label also concatenated in English word order. Driving the
app in Portuguese showed "setembro 7 – 13, 2026"; it's a key now, so each
language builds its own:

| | |
|---|---|
| pt-BR | 7 – 13 de setembro de 2026 |
| de | 7. – 13. September 2026 |
| ja | 2026年9月7日 – 13日 |

Weekday initials in the mini month come from the translated day name instead of
a hardcoded `["M", "T", "W", …]`, so German reads M D M D F S S.

## Testing

- `tsc -b`, `oxlint`, `generate-locale.mjs --check` — clean; all 8 packs at
  100% (1870 strings).
- Full suite: **357 files, 4285 tests passed**, 1 file / 39 tests skipped.
- `schedule-label.i18n.test.ts` covers the regex-rewrite trap directly: it
  asserts the pt-BR interval sentence is assembled from keys, that 60 minutes
[...]

PR #921, opened 2026-09-07
Translate the team map, webhooks, teams and the skill recorder

alisson-acioli:
Stacked on #918 — the diff below the last commit is that PR. Base is `main`.

## The gap

Six surfaces were still entirely English in every language: the team map and
its shared-context editor, the bot-instructions dialog, the webhooks panel, the
team library, the skill recorder, the credential-request card in chat, and the
remote agent settings panel.

## What changed

299 keys across all 8 packs:

- **TeamMapPage** — the header, section cards, agent handoffs, and the shared
  context dialog (including its byte counter and unsaved-changes confirmation).
- **BotInstructionsDialog** — the whole dialog.
- **WebhooksPanel** — the editor, the setup and activity tabs, the delivery
  outcomes, the relative timestamps and the rotate/delete confirmations.
- **TeamLibraryPanel** — Explore, Import and the folder scout, the import
  preview with its counts, and the three "what gets loaded" notices.
- **SkillRecorderPage** — the intro, the recording and review screens, and the
  saved-skill confirmation.
- **SecretRequestCard** — the four description states and the footer status.
- **RemoteAgentSettingsPanel** — the profile fields and the notifications row.

## Two helpers that carried copy

`teamMapStatus()` returned a rendered label that the map also **compared**
(`status.label === "Working"` picked the pulsing dot). It now returns
`labelKey`, so the comparison is against a stable key and the component
translates at render — the same fix the Local VM section got in an earlier PR.
`src/lib/team-map.test.ts` asserts the new shape.

`eventLabel()` in the skill recorder builds a sentence per recorded step; it
resolves through the catalog on every call, with singular and plural keys for
the typed-character count. macOS keycap names (`Return`, `Tab`, `Space`) stay
English — they are what is printed on the key.

`TEAM_BACKUP_EXCLUSIONS` lives in `shared/`, which the server imports and which
cannot reach the renderer catalog. The panel no longer imports it and renders
the catalog copy for the same fact instead; the constant stays where it is.

Catalog entries from the community teams repository, webhook payload previews
and adb device states come from outside the renderer and stay untranslated, as
in the earlier PRs.

## Following main

#917 merged while this was in flight, and it brought two conventions worth
matching here, both in the second commit:

- dates format with `activeLocale()` rather than the OS default — the team map's
  "last saved" line and the webhook activity timestamps;
- an error parked in React state carries `{ key }` instead of a resolved
  sentence, so it survives a language switch. `SkillRecorderPage` already had
[...]

PR #606, opened 2026-08-31
Carry explicitly selected skills in team packages

LukichevDsgn:
## Summary

Carry explicitly selected imported skills in setup packages, disabled by default. This preserves the original skill-package contribution while narrowing the older proposal to the part supported by the current app.

- Package export accepts `{"format":"package","skillIds":["source-check"]}` at `POST /api/teams/export`. Omitted or empty `skillIds` exports no skills; regular **Export backup** is unchanged. No new export picker or scoped bot/room export is added.
- Packages carry validated `SKILL.md` definitions and per-agent references. Shared validation checks names/frontmatter, size limits and references; conflicting content or portable metadata is rejected. Supporting files, enabled state and local attachment paths are not carried.
- Teams → Import previews the included names and explains that they remain disabled. Imports reuse the existing protected skill installer and additive rollback, without enabling instructions or changing existing bots.
- The old manual-routine proposal is removed from this PR. Existing once/daily/interval schedules and paused routine imports remain unchanged. No new scheduler or attachment-handling behavior is introduced.

## Verification

- 300 tests passed across `server/index.test.ts`, `server/bot-package.test.ts`, `server/package-export.test.ts`, `server/team-manifest.test.ts`, `server/skills.test.ts`, `server/team-backup.test.ts`, `server/team-backup-workflow.test.ts`, `server/team-package-skills.e2e.test.ts`, and `src/lib/team-import.test.ts`.
- Both renderer and server type checks, whole-repository strict lint, and the scoped diff check passed.
- New isolated fake-engine workflow proves explicit-only export, exclusion of unselected private instructions, disabled imports even from an enabled source skill, unchanged original bots/backups, and cleanup after an induced skill-storage failure. A copied bot's real fixture turn receives no disabled skill instructions.
- Actual Safari App → Teams → Import → native file picker showed both fixture skill names and the disabled-on-import/profile-review notice before **Add team**. No live app data was used. The scripted E2E is not presented as visual coverage.

See `docs/verification/team-backups.md` for the API example and repeatable fixture test. No new dependencies.

PR #927, opened 2026-09-07
Translate the last English surfaces

alisson-acioli:
Stacked on #921 → #923 → #924, all of which target `main`; the file count here shrinks to this PR's own 53 files once they land. It also carries a merge of `main`: #919 reworded the Full access warning after this branch's base, and that one file needed a hand resolution — see below.

Follows #917 and #918. With this the language selector changes the whole app: what the screen-by-screen passes left behind was the long tail — chat satellites, calls, the VM workspace and remote desktop, the command palette, pairing, the browser panel, skins, team import, the speaker, and the errors the store and the lib helpers throw for someone to read.

332 keys across the eight packs (2417 strings, every pack at 100%), 53 files.

### Two shapes needed more than a `t()` wrap

**Module-scope constants and maps resolved once, at import**, so the first language a module was loaded in was the only one it ever spoke:

- `GOAL_RUN_PREVIEW_LABEL` and `GoalRunCard`'s status copy now hold `LocaleKey`s and resolve at render; a skin carries `taglineKey`, not a resolved `tagline`.
- `FULL_ACCESS_WARNING` and `LOCAL_COMPUTER_AUTO_WARNING` are functions now. Both are what someone reads before letting a bot act unattended, so they have to be in that person's language.
- The pairing-failure, remote-access-failure and turn-timeout messages move into their callers the same way.

**A composer chip's `label` doubled as its kind**, picking the icon and the remove label. The kind stays `"PASTED" | "IMAGE" | "FILE"`; only the badge a person reads is translated.

`src/lib/remaining-surfaces.i18n.test.ts` guards both shapes.

### On #919

That PR reworded the Full access warning after this branch's base. The catalog carries the new wording, so the assertion it added still holds — I ran main's `ApprovalModeSelector.test.ts` against this branch and all 11 tests pass.

### Left in English on purpose

- `composer-attachments.ts`'s last-resort `"Attached file"` — `server/index.ts` imports that module by relative path, so it can never reach the renderer catalog. Commented in place.
- The `AbortError` message the store throws to cancel a turn, and the server sentence `phone-setup.ts` compares against with `startsWith`. Both are control flow, not copy; translating either breaks a check.
- The raw protocol summaries in `lib/inspector.ts` — a developer event trace, better left greppable.
- Sentences the server generates (bot overview, system prompt, connector blurbs, adb device states). Those want a server-side catalog of their own, which is its own PR.
- `src/testing/*` and `mascot-preview.tsx`, harnesses rather than app.

### Checks

- `tsc -b`, `oxlint`, `node scripts/generate-locale.mjs --check` clean
- full `vitest run` green
- drove the running app in pt-BR and switched live to 日本語: pairing page, bot settings → voice, sidebar, chat header and composer all re-render without a reload

## Summary by CodeRabbit

* **New Features**
  * Expanded localization across settings, dialogs, calls, attachments, scheduling, routines, search, updates, voice, desktop controls, and accessibility labels.
  * Added translated calendar, timeline, status, error, and pluralized messages.
  * Expanded translations for German, English, Spanish, French, Japanese, Portuguese, and Chinese.
  * Dates, times, numbers, and token counts now follow the selected locale.

* **Tests**
  * Added coverage for localized content across supported locales and major interface areas.


alisson-acioli:
Review findings applied in b7618214, plus the same class of miss elsewhere.

- **GoalRunCard**: `turn`/`turns` reuses `chat.usage.turnsOne`/`turnsMany`; the `coordinating` line is `goal.coordinating`.
- **ManageMembersPanel**: `common.save` plus `teams.count.botOne`/`botMany`.
- **PromptPreview**: all three `toLocaleString()` calls take `activeLocale()` — and so do the seven others I then found in ChatView, GroupView, TaskPicker, TeamMapPage, IdentitySection and SoulField.

My first sweep required 12+ characters and two English function words, so short strings slipped through it. A stricter one — singular/plural ternaries, bare capitalised JSX words, and `toLocaleString()` with no argument — found the rest: `Cloud`/`Local`/`Effort` in the model picker, the three VM states, the inspector's entry count, the task timeline's event labels, the SOUL.md byte counter and its hint, the companion device count, the token count in the task picker, and the channel-member fallback name. 26 more keys across the eight packs; the catalog is at 2443.

`tsc -b`, `oxlint` and `generate-locale --check` clean; full `vitest run` green (359 files, 4296 tests). Both sweeps now come back with nothing but selectors, regex fragments and the exclusions listed in the PR description.

alisson-acioli:
CI on b7618214 was mine — fixed in 9df4fedc.

`pnpm typecheck` runs two projects and the second one caught it:

```
src/lib/taskTimeline.ts(3,19): error TS2307: Cannot find module '@/lib/i18n'
```

`server/task-timeline.test.ts` imports that module by relative path, so it lands in `tsconfig.server.json`, where the `@/` alias does not exist. Vitest resolved it through the Vite alias, which is why the tests were green locally while `tsc -p tsconfig.server.json` was not.

Rather than leave the four timeline labels in English, the event now carries a `labelKey` — a literal union, no import — and ChatView resolves it; `label` stays the English fallback `server/task-timeline.test.ts` asserts on.

I also checked the whole boundary: `composer-attachments`, `drafts` and `taskTimeline` are the only `src/lib` modules the server reaches by relative path. `drafts` has no copy in it, and `composer-attachments` is the one documented in the PR description as staying English.

Verified with both halves of the script this time — `tsc -b --force` and `tsc -p tsconfig.server.json`, both clean — plus `oxlint`, `generate-locale --check`, and the renderer suite together with that server test (842 tests).

alisson-acioli:
All applied in 7076efc2 — one declined, with a reason.

**Major — server-bound denial message.** Right, and it was in four places, not one. I traced it: `/api/bots/:id/respond` hands `message` to `adapter.respondToRequest`, which hands it to the ask broker, which writes it into the answer the engine reads — right next to the driver's own English notes (`"OpenMausBot: the turn ended"`). Nothing renders it back. So `store.tsx`, `PendingApproval`, `CallView` and `GroupCallView` all send English again, and the four keys are gone from the eight packs.

**teams footer composition.** Confirmed, and it was structural rather than per-language. Gluing a possessive to a counted noun phrase works in English ("Your 3 existing bots and their conversations…") and breaks in every language with agreement — pt-BR rendered "Suas 1 bot existente e as conversas…". Each pack now carries the whole noun phrase in the variant. Checked all three counts (0, 1, 3) in all eight; zh needed the `的` moved to `teams.existing.none` as well, which the suggestion did not cover.

**skill-recorder keyCount.** Real bug: absent count fell through to the plural key and rendered "1 keys". Normalized before the check.

**ChatView sr-only state.** Right — the visible label was translated and the screen-reader prefix still spoke `running`/`complete`/`failed`. Mapped to keys.

**UpdateBanner / VoiceSettings / TeamMapPage.** Right, and they exposed a gap in my sweep: JSX text sitting after an icon on the same line, which my filters read as markup. Closing that pattern found twelve — the updater's five button labels, the image preview's loading and retry text, two in the Local VM pane, the inspector title, the search branch hint, and the voice preview button. For the locale argument I fixed the root instead: `formatTime` in `state/store.tsx` used `toLocaleTimeString([])`, so all seven of its callers followed the browser; that plus the three remote-desktop clocks now pass `activeLocale()`.

**Declined: zh `teams.subtitle`.** The English is elliptical too ("or bring your own."), and ending on `的` is a natural elliptical possessive in Chinese. Left as is.

+14 keys, −4, catalog at 2453. Typecheck on both projects, `oxlint`, `generate-locale --check` clean; full `vitest run` green (359 files, 4296 tests).

alisson-acioli:
Applied in d298ceb6 — and the sentence already existed.

`schedule.intervalStarting` ("{cadence} · starting {date}, {time}") is what the routines page uses, with the word order each language needs. The panel was reimplementing it and keeping the joining word in English. It now passes its own compact date into that key instead of duplicating the sentence, so no new key:

```
en     Every 2 hr · starting Sep 8, 3:00 PM
pt-br  A cada 2 h · a partir de 8 de set, 15:00
ja     2 時間ごと · 8 de set 15:00 から
```

Looking at that function I found one more of my own: the next-run clock right below it still called `toLocaleTimeString([])`, so its date followed the app while its time followed the browser. Fixed in the same commit.

Typecheck on both projects, `oxlint` and `generate-locale --check` clean; full `vitest run` green (359 files, 4296 tests).

alisson-acioli:
Applied in 328d45f3, and it was in two places.

`schedule.onceAt` is already exactly this "{date}, {time}" pairing, so both panels call it rather than pasting a literal comma. The join was the visible half of the bug — Japanese does not use one there:

```
en     Today, 15:00
pt-br  Hoje, 15:00
ja     今日 15:00        (was "今日, 15:00")
de     Heute, 15:00
```

Sweeping the pattern instead of the reported line turned up `ComputerPanel.tsx`, which carries the same copy of `nextRunLabel`. Fixed there too. No new key.

Typecheck on both projects, `oxlint` and `generate-locale --check` clean; full `vitest run` green (359 files, 4296 tests).

alisson-acioli:
@milind-soni rebased onto `57eba9a7`, on top of #924. `mergeable: true` again.

This one had drifted furthest, so it is a re-derivation rather than a rebase: every conflict was resolved toward main's code and the catalog lookups re-applied on top.

**Files main rewrote under it**

- **`BrowserPanel`** (#1044). The live view is main's; the branch had only ever translated the install card. The whole panel is keyed now — tabs, address bar, take-control/return-to-bot with its three hover states, the restart confirm, the typing dialog, and the admin/profile gates main added. `browserPanel.own`, `ownBody`, `lastFailed` and `serverAnswered` had no place left and are deleted rather than left orphaned.
- **`ModelPicker`** keeps your extracted `ModelEngineRail`; the branch's inline rail is gone.
- **`ComputerPanel`** keeps `routineScheduleState()` / `routineNextLabel()` from `lib/routine-display.ts`; the branch's duplicate `routineScheduleLabel` / `runStatusLabel` / `nextRunLabel` are gone.
- **`PairPage`** gained the e-mail sign-in flow, which is keyed here too.
- **`ChatMarkdown`** gained the wrap / download / copy toolbar, also keyed.
- **`lib/skins.ts`** carries `taglineKey`s, with **Daylight** (#929) added to the list.
- `lib/assemblyai-transcription.ts` and the skill recorder went away with #1034 and their keys went with them.

**Terminology followed main** — channel → group, task → thread — across `call.*`, `goal.*` and `members.*`, in English and all 7 packs, hashes refreshed.

**Three upstream commits** that the old branch had picked up through a merge (#606, #919, #922) are dropped: they are in main already.

419 keys here, 1092 across the stack.

Green locally: `tsc -b` + `tsc -p tsconfig.server.json`, `oxlint --deny-warnings .`, `generate-locale.mjs --check` (9 packs valid, 2750 English strings), `vitest run src/` — 160 files, 1337 tests, all passing.

Two things I deliberately did not pull in, both new since the branch was cut, both easy follow-ups if you want them in the same series:

1. **The Memory panel** (#1042). You rewrote it while this was rebasing, so `bot-settings/MemorySection.tsx` and `lib/memory.ts` are main's English. Its journal builds sentences in English word order (`"Scout added 2 lines to MEMORY.md"`), which needs the same key-per-shape treatment `scheduleSentence()` got — a change worth reviewing on its own rather than buried at the end of this stack. The nine `botSettings.memory.*` keys the old panel used are deleted, so nothing is left dangling.
2. **`zh-tw`**, which postdates the whole stack: the new keys fall back to English there. `generate-locale.mjs zh-tw` in one pass across the four PRs would close it.

PR #930, opened 2026-09-08
Make the thread header a label strip, not a toolbar

alisson-acioli:
The header above a thread measured **65px**. The same header in the app this
layout follows measures ~38px (calibrated against the 12px macOS close button
in a screenshot). It is now **45px**, and the two controls that were making it
tall are the reason it could shrink at all.

## What was setting the height

Nothing was styled around these; they were simply the tallest things in the row,
while every control beside them is 30px:

- the rename pencil in `RenameTitle`, at `size-10` (40px)
- the call button in `CallView`, at `size-9` (36px)

Both come down to match their neighbours. Then `py-3` → `py-1.5`, the bot name
15px semibold → 13px medium, its avatar 28 → 18.

## Two details that are not size

- **The name and the avatar open the same bot**, so they became one hover target
  carrying the tint a sidebar row uses. The pointer cursor has to be set
  explicitly: Tailwind v4's preflight dropped it from buttons, and the fix
  already in `styles.css` is scoped to `[data-sidebar]`.
- **A hairline under the header** replaces empty space as the separation from
  the messages — in the 1:1 thread and in a channel alike.

## On the weight

`font-medium` is not a lighter name. Inter renders visibly heavier than the SF
this layout was drawn against, so 500 here reads the way 600 reads there. The
same reasoning applies to the sidebar rows, but those live in another PR.

## Checks

`pnpm typecheck`, `pnpm lint`, and the component suite (317 tests). Header height
measured in the browser before and after: 65 → 45.

## Summary by CodeRabbit

* **Style**
  * Refined call, chat, and room header controls for a more compact, consistent appearance.
  * Tightened header spacing and added subtle bottom borders in chat and room views.
  * Reduced title and room-name text sizing for improved visual hierarchy.
  * Updated profile, avatar, and rename controls with smaller dimensions and spacing.
  * Added pointer feedback to the chat profile control.

### DIFF
--- a/src/components/CallView.tsx
+++ b/src/components/CallView.tsx
@@ -149,7 +149,9 @@ export function CallTargetButton({
         aria-label={label}
         title={label}
         className={cn(
-          "relative flex size-9 items-center justify-center rounded-full transition-colors",
+          // sized with the other header controls: at size-9 this one button set the
+          // height of the whole strip
+          "relative flex size-8 items-center justify-center rounded-full transition-colors",
           active
             ? "bg-danger text-white hover:brightness-110"
             : unavailable
--- a/src/components/ChatView.tsx
+++ b/src/components/ChatView.tsx
@@ -1114,22 +1114,25 @@ export function ChatView({ bot }: { bot: Bot }) {
         className={cn(
           // @container so the chips on the right can fold to icon bubbles
           // when the column is narrow (side panel open, small window)
-          "@container/chathead flex items-center justify-between px-5 py-3",
+          "@container/chathead flex items-center justify-between px-4 py-1.5",
+          // The header is a label for the thread, not a title bar: a hairline
+          // separates it from the messages instead of empty space.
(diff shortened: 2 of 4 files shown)


milind-soni (commented):
Reviewed head 0d5eb6dbb2fc6241c6e0e6de54d4acdbd3f31200: no actionable product bug found. It integrates cleanly with main 7f0c9304; nine focused ChatView/GroupView/RenameTitle tests, typecheck, and targeted lint pass. The original author Signed-off-by trailer is missing under current CONTRIBUTING.md. Please add it and push; we can land this after the updated head's checks pass. Native/interactive visual behavior was not independently exercised in this pass.

milind-soni:
Thanks for exploring a more compact header. We are keeping the current header sizing and clearer bot identity for now, so the avatar/text reduction is not a change we want to ship. Closing without merging as a design decision; this is not superseded by the separate sidebar-avatar fix.

PR #937, opened 2026-09-08
fix(electron): prove a data-directory lease dead across a reboot

aivsomkar:
## The bug

A user could not launch OpenMausBot after a restart:

> **OpenMausBot could not start safely**
> OpenMausBot is already using this data directory (process 459). Close the other instance first.

There was no other instance to close, and no in-app recovery — the app was bricked on every launch.

## Root cause

The data-directory lease identified its owner by **pid alone**:

```js
const owner = { version: 1, pid: process.pid, host: hostname(), token, createdAt };
```

Three things combine into a permanent failure:

1. **The lease survives an unclean exit.** `release()` runs on `will-quit` / `process exit`. A macOS restart force-kills apps that miss the shutdown deadline, and the desktop does async CUA + utility-server teardown before quitting. The lease is left holding the old pid.
2. **Pids reset on reboot and get recycled.** A login-item launch records a low pid; after the next boot that pid belongs to a root-owned daemon.
3. **`EPERM` counts as alive** (`processIsAlive`, by design — the pid exists but this account cannot signal it). So a recycled pid reads as a live owner, the stale-lease reaper never fires, and startup fails forever.

Both pids in the user's report — **459** and **400** — sit squarely in the range macOS assigns during boot, consistent with the app being launched at login.

## The fix

Record which boot wrote the lease, and require a match before believing the owner is alive.

```js
function ownerIsAlive(owner) {
  return !ownerPredatesThisBoot(owner) && processIsAlive(owner.pid);
}
```

**The pid check is unchanged and still required.** Boot identity is an added *necessary* condition for declaring a record dead — it can never make a record look alive, and it never relaxes exclusion. A wrong "stale" verdict would let two instances share one data directory, which is the exact harm this module exists to prevent, so both signals are exact rather than heuristic, and neither is derived from the wall clock (NTP can move it):

- **macOS / Linux** — the kernel's boot session id (`kern.bootsessionuuid`, `/proc/sys/kernel/random/boot_id`), fixed for the life of one boot. Complete coverage.
- **Everywhere else** — a since-boot clock cannot run backwards within a boot, so a recorded uptime above the machine's current uptime can only have been written before a reboot.

Records written by older builds carry neither field and keep the previous pid-only behaviour exactly, so an in-place upgrade cannot fail on an "invalid lease".

The contention message now names the lease record, so anyone still stuck can recover without contacting support:

> OpenMausBot is already using this data directory (process 459). Close the other instance first. If no OpenMausBot is running, delete this lease record and start again: "/Users/…/.openmausbot/openmausbot-server.lease".

### Known limit

Windows has no cheap synchronous boot id, so it relies on the uptime-regression check. That is **sound** (it cannot produce a false "stale") and it catches the common launch-at-login case that caused this report, but it misses one shape: machine up a long time, app launched later in the session than the previous session's uptime. That case degrades to today's behaviour, now with a recoverable error message. Closing it fully would need either a ~300-500ms PowerShell spawn at startup or a wall-clock age comparison that an NTP correction could turn into a false "stale" — deliberately not taken.

## Test plan

Written test-first; every new test was watched fail against the pre-fix module.

- `node --test electron/data-dir-lease.node-test.mjs` — **17/17 pass** (11 pre-existing + 6 new)
  - `a stale lease whose pid was recycled by a root-owned daemon no longer blocks startup` — reproduces the report exactly using pid 1 (always live, always root-owned, so `kill(0)` answers `EPERM`). **Fails on `origin/main`, passes here.**
  - `a lease left behind by an earlier boot is retired even though its pid is live`
  - `a lease recorded above the machine's current uptime is retired` — the Windows path
  - `a live owner from the current boot still blocks a second instance` — guards against a false "stale"
  - `boot identity does not weaken exclusion: only one of many live racers wins` — 8 concurrent live claimants, exactly one wins, directory left usable
  - `the contention error names the lease record so a stuck owner is recoverable`
  - The 6 pre-existing tests that plant old-format records still pass unchanged, proving the upgrade path.
- `vitest run server/data-dir-lease.test.ts` — 5/5. The record-shape guard was updated to pin the two new fields rather than loosened.
- `vitest run server/index.test.ts` — 198/198.
- `node --test "electron/*.node-test.mjs"` — 140 pass, 0 fail. The 11 cancellations in `approval-trusted-mode` / `server-boot-probe` are present on pristine `origin/main` and unrelated.
- `tsc -b && tsc -p tsconfig.server.json` — clean. `oxlint` on changed files — clean.

Not yet exercised on a real reboot or on Windows hardware.

## Platforms
|---|---|---|
| macOS     | yes | in this PR — exact fix via `kern.bootsessionuuid`; reported bug reproduced pre-fix and fixed |
| Windows   | yes | in this PR — the non-darwin/linux branch *is* the Windows path, no separate win32 gate needed; sound but partial (see Known limit); NOT built on Windows |
| iOS       | no  | n/a: no data-directory lease on the phone clients |
| Android   | no  | n/a: no data-directory lease on the phone clients |

## For the affected user, before this ships

Quit OpenMausBot, then:

```bash
rm -f ~/.openmausbot/openmausbot-server.lease \
      ~/.openmausbot/.openmausbot-server-child/openmausbot-server.lease
```

Lock state only — no chats, bots, or settings are touched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection and recovery of stale data-directory leases after system reboots or process ID reuse.
  - Strengthened handling for lease contention and concurrent acquisition.
  - Updated already-in-use errors with guidance for removing stale lease records.
  - Preserved compatibility with older lease records while recording additional boot-session and uptime details.
  - Improved recovery after interrupted or competing lease-acquisition attempts.

### DIFF
--- a/electron/data-dir-lease.mjs
+++ b/electron/data-dir-lease.mjs
@@ -1,6 +1,7 @@
+import { execFileSync } from "node:child_process";
 import { createHash, randomUUID } from "node:crypto";
 import { existsSync, linkSync, mkdirSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "node:fs";
-import { hostname } from "node:os";
+import { hostname, uptime } from "node:os";
 import { join } from "node:path";
 
 const LEASE_NAME = "openmausbot-server.lease";
@@ -11,6 +12,7 @@ const DELEGATED_CHILD_DIR = ".openmausbot-server-child";
 const CHILD_LEASE_ENV = "OPENMAUSBOT_INTERNAL_DATA_DIR_LEASE";
 const UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
 const MAX_PID = 0x7fffffff;
+const MAX_BOOT_ID = 128;
 const MAX_REAPER_GENERATIONS = 128;
 
 export class DataDirLeaseError extends Error {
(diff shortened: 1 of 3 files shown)


milind-soni:
Reviewed the startup and delegated-server paths. The boot-aware lease fix is worth bringing in. I added test-only fixes for the two review findings (privileged PID-1 runners and synchronized contention), plus stale reaper/delegated-child coverage. Verified 19 lease tests, 250 server tests, typecheck, and an isolated recovery/chat fixture preserving existing bot data. Waiting for checks on the updated head before merging.

PR #942, opened 2026-09-08
feat(playbooks): propose playbook changes through a confirmation card

### LINKED ISSUE #935: Let a Chief of Staff propose playbook changes for a section peer
### Before filing

### Area

### Problem

Once a bot exists, its `playbooks` can never change. There are exactly two write paths, and both only ever write to a *freshly created* bot:

- `server/index.ts` ~9506 — package import
- `server/team-backup.ts` ~125 — backup restore

No PATCH route and no in-process tool can touch them. `BOT_PROFILE_PATCH_FIELDS` (`server/bot-profile.ts`) doesn't include `playbooks`, and neither does the general bot PATCH.

So changing one playbook on a bot you're already using means re-importing the whole package. Import is deliberately additive (#797), so that creates a fresh, numbered duplicate of *every* agent in the package — not just the one that changed — and each duplicate starts over on everything the package format can't express (`cwd`, `composio`, model and approval-mode overrides). That's a lot of collateral for a one-line playbook edit, and it repeats for every edit.

What makes this feel like a gap rather than a deliberate restriction: **a Chief of Staff may already replace the more powerful field.** `propose_profile` (`server/drivers/agents-proxy.ts`) lets a Chief set `soul` — up to 24000 bytes of standing instructions — for any peer in its section. Playbooks are the *weaker* instrument: they're only mounted when a declared trigger matches the job, capped at three per turn (`selectInstalledPlaybooks`), and rendered with an explicit disclaimer that they "do not grant tools, connected apps, permissions, or authority to override safety and user approval requirements." Yet they're the one instruction surface a Chief cannot touch.

### Proposal

Let a Chief of Staff propose a playbook change for a peer in its own section, using the confirmation-card flow that already exists for `propose_profile`, and the provenance discipline that already exists for `skill_manage`.

Shape:

- **Single-playbook operations, not wholesale replacement** — upsert/remove by `key`. Handing every playbook to a tool call is impractical when one playbook's `instructions` can be 24000 characters; in practice the caller wants to push the one playbook that changed.
- **Same authorization as today**: Chief-only, same-section-only, re-validated at confirm time — `ProfileRequestService`'s `validateTarget` (`server/index.ts` ~4858) already implements exactly this rule.
- **Nothing applies without the user confirming a card**, same as `propose_profile`.
- **Carry provenance.** `renderInstalledPlaybooks` currently tells the model these are "reviewed, package-authored playbooks." That stops being true the moment a Chief can write one, so either record the origin on the playbook (`source: "package" | "chief"`) and render the two differently, or drop the claim. `skill_manage` already treats this as a first-class concern — it requires a `source` label for exactly this reason — so there's a house pattern to follow.
- **Could ship opt-in.** Skill authoring is gated behind `OMB_SKILL_AUTHORING_ENABLED` (`server/drivers/agents-proxy.ts`); playbook authoring could start the same way if that's the preferred risk posture for a new authoring surface.

One implementation note: the existing profile-request payload won't carry this as-is. `ProfileRequestChanges` is `Partial<Record<ProfileRequestField, string>>` (`shared/profile-request.ts`) — all values are strings, while an `InstalledPlaybook` is `{ key, name, summary, triggers[], instructions }`. So this either widens that payload type (pulling in `profile-versions.ts`, `profile-revision.ts` and the card UI) or gets its own request kind alongside routine requests and skill staging. My instinct is the separate request kind is cleaner, but I'd rather follow the maintainers' preference than guess.

### Alternatives considered

- **In-place team import update (#897)** — solves this more generally, but runs straight into the additive-only boundary that import deliberately holds (untrusted package input must not reach existing records), and is a much larger change. This issue is the narrow version that never touches that boundary: the input comes from a bot inside the running system, not from a file, and it's already behind a user confirmation.
- **Put the process guidance in `soul` instead** — possible today, but it collapses standing rules and per-job process into one blob and loses the trigger-based mounting that makes playbooks cheap (only matching playbooks enter context, max three).
- **Re-import and clean up afterwards** — the status quo. Works, but the cleanup and re-configuration cost scales with the number of agents in the package and repeats on every edit.

Remownz:
Closes #935.

## Problem

A bot's playbooks are fixed at install time. `bot.playbooks` is only ever written while creating a *fresh* bot — once by the package import (`server/index.ts`) and once by a backup restore (`server/team-backup.ts`) — and no PATCH route or in-process tool can reach the field afterwards. `BOT_PROFILE_PATCH_FIELDS` doesn't include it, and neither does the general bot PATCH.

So changing one playbook on a bot you already use means re-importing the whole package, which is additive by design (#797) and therefore duplicates *every* agent in it, with each duplicate losing `cwd`, connected apps, and any model or approval override.

Meanwhile a Chief of Staff may already replace the **stronger** instruction on a section peer: `propose_profile` sets `soul`, up to 24000 bytes of always-mounted standing instructions. Playbooks are the weaker instrument — mounted only when a declared trigger matches the job, capped at three per turn (`selectInstalledPlaybooks`), rendered with an explicit non-authority disclaimer — yet they were the one instruction surface a bot could not touch.

## What this adds

`propose_playbook`, alongside `propose_profile`, with the same authority model and the same confirmation card.

- **Single-playbook upsert/remove by key**, never a wholesale replacement — one playbook's `instructions` can be 24000 characters, so handing over a whole set is not a realistic tool call.
- **Chief-only and same-section-only** for a peer, re-validated at confirm time, reusing the rule `ProfileRequestService` already enforces.
- **The proposal's limits are the package schema's own** (`shared/playbook-request.ts` mirrors the playbook block in `server/bot-package.ts`), so a card cannot produce a playbook an imported package could not have produced.
- **The pinned revision covers the target's whole playbook set** plus a private receipt, so a card written before another one landed fails closed rather than applying to a state the user never saw.
- Playbook proposals share the existing per-bot-per-thread proposal budget, and the card is a durable request, so it never offers "always allow".

### Provenance is recorded, not assumed

`renderInstalledPlaybooks` used to tell the model that every playbook was "reviewed, package-authored". An approved proposal carries one user approval, not a package review, so that claim would stop being true. Each playbook now states its own source in the rendered prompt:

```
<playbook name="Decision Brief" source="bot-authored, user-approved">
<playbook name="Ticket Quality" source="package-authored">
```

`source` is optional and absent means `package`, so records written before this change keep their meaning. It survives a backup roundtrip, and an export drops it again — once exported it *is* package content, reviewed by whoever imports it.

## Design note

This is a separate request kind rather than an extension of `profile-requests.ts`. `ProfileRequestChanges` is `Partial<Record<field, string>>`; a playbook is a structured record, so widening it would have touched redaction, diffing, the revision hash and both confirm paths inside a security-sensitive file. A third instance of the routine/profile request pattern seemed the smaller change. Happy to fold it in instead if you'd prefer that.

## Screenshots

A Chief proposing a playbook for a section peer — the header names the peer, because the card sits in the *Chief's* thread:

Chief proposes a playbook for a peer

A bot replacing one of its own playbooks, with the line diff and an already-confirmed card above it:

Bot updates its own playbook

## Tests

- `server/playbook-requests.test.ts` — 14 unit tests: validation against the package limits, the per-bot ceiling, secret redaction into the durable payload, "nothing would change", the Chief rule at propose *and* confirm, staleness, deny, and repeat-confirm never applying twice.
- `server/installed-playbooks.test.ts` — the rendered prompt labels each playbook's source.
- `server/index.test.ts` — through the real routes: the card stages without applying, confirming applies and reports the key, a repeat confirm settles, an ordinary bot is refused for a peer, a Chief is not, and the decisions audit records `card-shown:playbook` / `user-approved:user`.
- `server/team-backup.test.ts` — the existing roundtrip fixture now carries a bot-authored playbook, so it proves provenance survives a restore.

## Provenance and checks

- Upstream base commit: `600e315c` ("docs: update OpenMausBot support checkout link")
- Head branch: `Remownz:feat/chief-playbook-proposals`
- `pnpm typecheck` — clean
- `pnpm i18n:check` — clean, 8 languages at 100% (8 new keys)
- `npx vitest run server/` — 2851 passed, 1 pre-existing failure unrelated to this change (`control-omb` finds an extra locally installed engine; verified identical on unmodified `600e315c`)
- `npx vitest run src/ shared/` — 828 passed
- Screenshots were taken against an isolated instance with its own data directory; the `pr-assets` branch holding them is not part of this PR's diff.

https://claude.ai/code/session_01QGJ4YuGSoQP2weNtWvuR9a

## Summary by CodeRabbit

- **New Features**
  - Bots can propose adding, updating, or removing playbooks through reviewable approval cards.
  - Users can confirm or cancel playbook changes, with safeguards against duplicate or outdated approvals.
  - Authorized leaders can propose playbook changes for eligible peers.
  - Playbooks now show whether they were package-authored or user-approved.

- **Accessibility & Localization**
  - Added playbook approval labels, prompts, and accessible text across supported languages.

- **Bug Fixes**
  - Preserved playbook authorship information through backups and restores.


Remownz:
Thanks — two of the three were real misses on my side and are fixed in 9b6331bb. The third I'd like to push back on, with a reason.

**Fixed: `lastPlaybookRequestId` reached clients.** Correct, and it was an oversight in how I mirrored the profile pattern: `wireBot` and `wireTrustedApprovalBot` both already strip `lastProfileRequestId`, and I never added the playbook counterpart. It's destructured out of both now, right next to its sibling.

**Fixed: missing accessible label on the card's details block.** Also correct — `ApprovalCard`'s `<pre>` picks a label per request kind and playbook cards fell through to the generic `approval.aria.details`. I used a new `approval.aria.playbookChange` rather than the suggested `approval.aria.reviewPlaybook`: the latter is `PendingApproval`'s "Playbook change to review" phrasing, while this element's siblings are the plain nouns (`approval.aria.profileChange` → "Profile change"). New key, all 8 locales, `i18n:check` clean.

**Not changed: 403 vs 404 on a confirm-time refusal.** This one is deliberate consistency rather than an oversight. `profile-requests.ts` — the module this one is modelled on — does exactly the same thing:

```ts
// propose()
if (refusal) throw new ProfileRequestError(refusal, 403);   // line 248
// resolve()
if (refusal) throw new ProfileRequestError(refusal, 404);   // line 348
```

The split looks intentional in context: at confirm time the neighbouring failures also stop being about permission and start being about the card, with the stale check next to it returning 409 under the comment *"the card is no longer applicable as prepared."* A refusal there means the card can no longer be applied as written, not that the caller is being told off.

I'd rather not have the new module answer differently from the one it mirrors. If maintainers prefer 403 for both, I'm happy to make that change — but then in `profile-requests.ts` too, so the two stay aligned, and that felt like scope this PR shouldn't quietly take on.

On the docstring-coverage check: this codebase documents *why* in prose comments above the interesting code rather than per-function docblocks, and I followed that (see the header comments in `playbook-requests.ts`, `playbook-revision.ts`, and `shared/playbook-request.ts`). Happy to add formal docblocks if that threshold is something the project actually enforces.

Checks after this commit: `pnpm typecheck` clean, `pnpm i18n:check` clean (8 languages), `server/index.test.ts` + `server/store.test.ts` 270 passed, `src/components/` 331 passed.

milind-soni (changes requested):
Reviewed 9b6331bbd73b873aed3cbfe16ac35a3145552d0f. P2 at shared/playbook-request.ts:24: perBot permits 80 installed playbooks, but the package schema permits only 40 agents[].playbooks references. Approving the 41st playbook succeeds and then createBotPackageExport fails with `package.agents.0.playbooks Too big: expected array to have <=40 items`. Keep the limits aligned (the minimal fix is cap=40) and cover approval followed by export at the boundary.

Reproduced locally with the real service/export path. Cap=40 plus the regression passes 17/17. After resolving current-main conflicts, typecheck, targeted lint, i18n, 96 focused tests and an isolated real-server approval smoke pass. Please update the currently conflicting branch, preserve both playbook proposals and newer memory tools, and rerun CI. Both author commits also need Signed-off-by trailers per CONTRIBUTING.md.

PR #953, opened 2026-09-08
fix: answer AskUserQuestion with the model's own options, not Allow/Deny

aivsomkar:
> **Supersedes #344**, and carries its commit. @AdityaUmale found this bug independently and got the mechanism right where this branch got it wrong; his fix is the one that ships here. See the merge commit for what was taken and what was reconciled.

Claude's built-in `AskUserQuestion` reached OpenMausBot as a **permission** ask — the CLI routes it through `--permission-prompt-tool` exactly like a `Bash` or `Read` call. So a question like *"which model should this bot run on?"* arrived as a card offering **Deny / Always allow / Allow once**, over a truncated blob of JSON.

Worse than ugly: **Allow once** made the CLI run the tool against an interface that is not there, and the model was told

```
The user did not answer the questions.
```

The click was thrown away. In auto mode the question could be answered *for* you, so you never saw it at all.

## The fix

`server/permission-proxy.ts` recognises that one tool and forwards it as a **question** carrying the model's structured questions. The answer goes back through the tool's own `answers` field — what its schema calls "User answers collected by the permission component" — keyed by each question's text. Verified against claude 2.1.260:

```
proxy → {"behavior":"allow","updatedInput":{…,"answers":{"Do you prefer tea or coffee?":"Tea"}}}
model ← The user answered: "Do you prefer tea or coffee?"="Tea".
        Read the answers carefully — they may request clarification, changes,
        or that you not proceed — and follow what they actually say.
```

`src/components/QuestionCard.tsx` draws it the way Claude Code does: a tab per question, the model's own options with their descriptions, an "Other" free-text row, one **Submit answers**. One card for the whole set, so a person sees what they are committing to before answering any of it — `questionAnswersByQuestion` maps that single reply back per question.

## Also here, because the same gap made them wrong

- **No approval mode may auto-answer a question**, whatever channel it arrived on — typed as a question at the card, and a backstop in `auto-approve.ts` for one that arrives mis-typed.
- **A question asked inside a room renders at all.** It used to fall through to null while the asking bot waited out its 15-minute timeout.
- **A card's answer behavior comes from the CARD, not the button's label** (`src/lib/card-answer.ts`), so a question offering an option called "Allow" is no longer sent as an approval — which the broker refuses, closing the card as unreachable.
- **Closing a question answers it** instead of doing nothing.
- **A permission never takes its buttons or its summary from a nested `questions[]`** in its arguments. On the phone those labels become the decision buttons.
- **A timeout's own note is never filed as the person's answer** — the broker now puts `source` on the wire, and only a real person's words reach `answers`.
- **An answered question keeps its words** (`answeredText`), so a settled card still shows what was chosen after a reload instead of reading "answer".

The payload is bot-authored, so `shared/ask-question.ts` validates rather than casts: counts and lengths capped, duplicate labels dropped (a radio group cannot express them), question text scrubbed through the same secret-redaction boundary as a routine's instructions.

## Platforms

|---|---|---|
| macOS     | yes | in this PR, tested locally (OMB2) |
| Windows   | yes | in this PR (shared server + desktop UI, no platform gate); NOT run on Windows, but [`Package Windows` built green](https://github.com/milind-soni/OpenMausBot/actions/runs/34242380036) on a real Windows runner |
| iOS       | yes | single-question asks: in this PR (the flat `choices` fallback drives the buttons `ios/App/ChatView.swift` already draws). Multi-question card: #954, stacked on this branch |
| Android   | yes | same as iOS (`CompanionJson` has `ignoreUnknownKeys = true`). Multi-question card: #955, stacked on this branch |
| Companion | n/a | no new `/api/*` route — answers go through the existing `POST /api/threads/:id/respond` |

- [x] iOS: multi-question card in `ios/App` — #954
- [x] Android: multi-question card in `android/app/.../ui` — #955

## Test plan

- Full suite: **4523 passed**, 39 skipped. `typecheck`, `lint`, `i18n:check` clean.
- `server/permission-proxy.test.ts` (from #344, adapted to the one-card shape) — the ask leaves as a question, the `answers` contract comes back, a timeout's note is not filed as a choice, an unanswerable call is denied rather than carded.
- `shared/ask-question.test.ts` — parsing a real tool input, the caps, the malformed payloads that must stay ordinary approval cards, the answer text, and reading that answer back per question.
- `src/components/QuestionCard.test.ts` — the model's options render, "Always allow"/"Allow once"/"Deny" do not, radio vs checkbox per `multiSelect`, and a settled card shows what was answered.
- `src/lib/card-answer.test.ts` (from #344) — behavior from the card, not the label.
- `server/drivers/claude.test.ts` — a structured ask opens as a question with its questions and flat labels; a permission never borrows buttons or summary from a nested `questions[]`.
- Manual: packaged as OMB2 from this branch and run against a real bot on macOS.

## Summary by CodeRabbit

* **New Features**
  * Added structured question cards with selectable options, multi-select responses, custom text answers, and multi-question navigation.
  * Added support across web, Android, and iOS conversations and transcripts.
  * Submitted answers are sent back to the assistant and remain visible after completion.
* **Bug Fixes**
  * Questions are no longer presented as permission approvals or automatically approved.
  * Answered cards now display the user’s actual response.
  * Invalid or oversized question data is safely rejected and sensitive content is redacted.


aivsomkar:
⚠️ **Do not merge this as-is.** Two things surfaced after it was opened.

**1. #344 is the same bug, open since Sept 6** (@AdityaUmale). I did not find it before building this. It covers strictly more of the surrounding breakage than this PR does: questions asked inside a room, closing a question with its X, a card deciding allow/deny from the card rather than from the button's label (a question offering an option called "Allow" was being sent as an approval), and the timeout note no longer being filed as the person's answer.

**2. Its answer mechanism is the right one, and this PR's is not.**

This PR delivers the answer on the permission contract's **deny** channel, because a bare `{behavior:"allow"}` makes the CLI report "The user did not answer the questions." That works — the model reads the text and acts on it — but it arrives as `is_error: true`, which is a lie about what happened.

#344 fills in the tool's own `answers` field instead. I had not tested that, and it is correct. Against claude 2.1.260:

```
proxy → {"behavior":"allow","updatedInput":{...,"answers":{"Do you prefer tea or coffee?":"Tea"}}}
model ← Your questions have been answered: "Do you prefer tea or coffee?"="Tea".
        You can now continue with these answers in mind.
```

No `is_error`. That is the documented path and it should be the one that ships.

**What is worth keeping from this PR**, if anything: the tabbed multi-question card (`QuestionCard.tsx`) — #344 asks the questions one at a time through the existing option card, where this draws all of them at once with the model's own descriptions — plus `answeredText`, so a settled question card still shows what was chosen after a reload, and the iOS (#954) and Android (#955) ports of that card.

Holding for @milind-soni to decide whether that rides on top of #344 or is dropped.

milind-soni (changes requested):
Reviewed head 7c81aba8c28f516a65ad1f2e10fd2e8e4baa0919. Three reproduced correctness issues need fixing before this can land:

1. shared/ask-question.ts:182 splits answers on blank lines. A valid Other answer such as `Deploy Tuesday.\n\nDo not delete the backup.` reaches the provider as only `Deploy Tuesday.` Please use stable question IDs and an unambiguous structured answer payload, not formatted display prose.
2. server/permission-proxy.ts:140 returns original questions but builds answers keyed by normalized/truncated questions. For a question longer than 400 characters, answers[questions[0].question] is missing. Preserve the provider's original identity.
3. server/store.ts:346 redacts displayed questions after the proxy retains the original strings as identifiers. A multi-question card containing `Rotate token=abcdef1234567890?` silently loses its answer after redaction. Identity must survive redaction independently of the display text.

Three failing regressions reproduce these paths, including the actual spawned proxy. Existing focused TypeScript tests (203 passed, 1 skipped), typecheck, and an isolated ordinary-chat smoke pass, but do not cover these failures. There are also current-main conflicts in OptionCard.tsx/store.tsx. Please add author Signed-off-by trailers per CONTRIBUTING.md; I cannot certify those on your behalf. #954 and #955 depend on this PR.

milind-soni:
Rebased onto main (7d6844f7) so this can merge. Notes for the record:

- Linear rebase drops the merge commit; its reconciliation content is preserved by replaying the second commit as a 3-way merge of the PR's reconciled tree, so the file list and net diff are identical to the PR head (33 files).
- Conflicts resolved keeping both sides: the redacting `askInputSummary` from main plus the question branch; main's `allowSession` card field plus `questionRequest`; `threadId` and `groupId` both on the card actions.
- Main has since removed every auto-approval guard except Full access. The PR's "a question is never auto-answered" check is placed first in `autoVerdict`, so it applies to Full as well, matching the comment in `index.ts` that even Full never invents an answer.
- One lint adaptation: `...(asRecord(input) ?? {})` → `...asRecord(input)` in `permission-proxy.ts`, since `pnpm lint` now denies warnings.

Verified: typecheck, lint, i18n:check, the targeted suites (185 tests) and the full `server/index.test.ts` (205). iOS and Android unit tests were not run here.

PR #954, opened 2026-09-08
feat(ios): answer a structured ask with the model's own questions

aivsomkar:
Stacked on #953 (base is `fix/ask-user-question-card`, so this diff is iOS only).

#953 turned Claude's `AskUserQuestion` into a real question card on the desktop. The phone got the **single-question** path for free — the harness also sends flat option labels, which drive the buttons `ios/App/ChatView.swift` already draws — but a **multi-question** ask arrived with nothing tappable. One flat button cannot say *which* question it answered, so the card just sat there until the ask timed out into "use your best judgment".

`QuestionCardView` is the desktop's card in SwiftUI: a tab per question, the model's options with their glosses, an "Other" row for a reply the model did not think of, and one submit that sends every answer at once.

## The part worth reviewing

`AskQuestionAnswer.format` is a port of the formatter in `shared/ask-question.ts`, and it has to be exact: that string reaches the model as the tool's own result, so an answer given on the phone must be indistinguishable from one given on the Mac. `AskQuestionTests` pins it against the same literal strings the TypeScript test uses rather than trusting the two to stay in step. If they ever drift, that test says so.

Decoding is deliberately forgiving — a question with no options, a missing `version`. These payloads are bot-authored and arrive inside a transcript, so a surprising shape must cost one card and never the whole conversation.

## Platforms

|---|---|---|
| macOS     | n/a | #953 |
| Windows   | n/a | #953 |
| iOS       | yes | in this PR |
| Android   | yes | #955 (the Kotlin twin, same commit shape) |
| Companion | n/a | no new `/api/*` route — the answer goes through the existing `POST /api/threads/:id/respond` with `behavior: "answer"` |

## Test plan

- `cd ios && swift build && swift test` — 345 tests pass, including 7 new in `AskQuestionTests`: decoding a real card payload, an ordinary approval keeping the ordinary card, a question with no options, the exact answer text, an unanswered question being omitted rather than implied, and the settled card dropping the model-facing lead-in.
- `xcodebuild -project OpenMausCompanion.xcodeproj -scheme OpenMausCompanion -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 17 Pro' CODE_SIGNING_ALLOWED=NO build` — **BUILD SUCCEEDED** (after `xcodegen generate`, which this PR's new `App/` file requires).
- **Not yet seen on a device.** Stages 1 and 3 of `ios/TESTING.md` pass; stage 4 needs a paired phone and a bot that asks, which is a person's hands. The card's shape is the desktop one, which was tested on a real bot in an OMB2 build.

### DIFF
--- a/ios/App/Cards/QuestionCardView.swift
+++ b/ios/App/Cards/QuestionCardView.swift
@@ -1,0 +1,308 @@
+import CompanionCore
+import SwiftUI
+
+/// The question card: what the bot actually asked, and its own answers.
+///
+/// A structured ask (Claude's `AskUserQuestion`) reaches the harness through
+/// the permission channel, so before this it drew the ordinary card — a flat
+/// row of buttons over "which model should this bot run on?". One tap could
+/// not say WHICH question it answered, so a multi-question ask had nothing
+/// tappable at all and the phone could only watch it time out.
+///
+/// The desktop's `src/components/QuestionCard.tsx`, in SwiftUI: a tab per
+/// question, the model's options with their glosses, an "Other" row for a
+/// reply it did not think of, and one submit that sends every answer at once.
+/// The answer text is built by `AskQuestionAnswer.format`, so an answer given
+/// here is byte-for-byte the one the Mac would have sent.
+struct QuestionCardView: View {
+    let chat: Chat
+    let message: Message
+    @EnvironmentObject private var session: Session
+
+    /// Per question: the option labels ticked, and the free-text reply.
+    @State private var picked: [Int: Set<String>] = [:]
+    @State private var custom: [Int: String] = [:]
+    @State private var other: Set<Int> = []
+    @State private var active = 0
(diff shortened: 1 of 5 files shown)


milind-soni (changes requested):
Reviewed head 55a8b97dcef90180900d0fb80a69512bc772b467. QuestionCardView.swift:293 sets sent before awaiting Session.answer, but Session.answer swallows transport errors through perform. A failed send permanently leaves the card locally answered and prevents retry until the view is recreated. Please only settle after success/server acknowledgement, or return a result and clear the submitting state on failure. Add a failed-send/retry regression.

The eight Swift question tests pass. This PR also depends on #953's answer-encoding fixes; after that lands, retarget to main and rerun integration checks. Please add your author Signed-off-by trailer per CONTRIBUTING.md.

PR #955, opened 2026-09-08
feat(android): answer a structured ask with the model's own questions

aivsomkar:
Stacked on #953 (base is `fix/ask-user-question-card`, so this diff is Android only). The Kotlin twin of #954.

#953 turned Claude's `AskUserQuestion` into a real question card on the desktop. The phone got the **single-question** path for free — the harness also sends flat option labels, which drive the buttons `MessageRow.kt` already draws — but a **multi-question** ask arrived with nothing tappable. One flat button cannot say *which* question it answered, so the card just sat there until the ask timed out into "use your best judgment".

`QuestionCardView` is the desktop's card in Compose: a tab per question, the model's options with their glosses, an "Other" row for a reply the model did not think of, and one submit that sends every answer at once.

## The parts worth reviewing

`AskQuestionAnswer.format` is a port of the formatter in `shared/ask-question.ts`, and it has to be exact: that string reaches the model as the tool's own result, so an answer given here must be indistinguishable from one given on the Mac or the iPhone. `AskQuestionTest` pins it against the same literal strings the TypeScript and Swift tests use rather than trusting the three to stay in step.

The radio/checkbox markers are drawn with `Box` rather than taken from the icon set — the app bundles `material-icons-core` only, and `CheckBox`/`RadioButtonChecked` live in the extended pack. Twelve lines beat a new dependency for four glyphs.

`QuestionCardRules` exists so the routing decision — which card a transcript row draws — is testable without a Compose harness. The mistake it guards is exactly the one the desktop shipped: a structured ask routed to the approval card, offering Deny / Always allow / Allow once over a question that has no such answer.

## Platforms

|---|---|---|
| macOS     | n/a | #953 |
| Windows   | n/a | #953 |
| iOS       | yes | #954 |
| Android   | yes | in this PR |
| Companion | n/a | no new `/api/*` route — the answer goes through the existing `POST /api/threads/:id/respond` with `behavior: "answer"` |

## Test plan

- `cd android && ./gradlew cleanTest :core:test :app:testDebugUnitTest :app:assembleDebug` — **BUILD SUCCESSFUL**; 493 core and 863 app tests, 0 failures. New: 8 in `AskQuestionTest` (decoding, the exact answer text, an unanswered question omitted rather than implied, the settled card's lead-in) and 3 in `QuestionCardRulesTest` (which card a row draws, and the tab labels).
- **Not yet seen on a device or emulator.** Compile and unit gates pass; putting it in front of a real bot needs a paired phone and a person to answer, and the card's shape is the desktop one, which was tested on a real bot in an OMB2 build.

### DIFF
--- a/android/app/src/main/kotlin/com/openmausbot/companion/ui/MessageRow.kt
+++ b/android/app/src/main/kotlin/com/openmausbot/companion/ui/MessageRow.kt
@@ -393,7 +393,13 @@ private fun MessageContent(
 ) {
     when (message.kind) {
         Message.Kind.TEXT -> TextBubble(chat.threadId, message, endsRun, openLink, openAttachment)
-        Message.Kind.OPTIONS -> CardView(chat, message, haptics)
+        // A structured ask draws its own card: its answers are the model's
+        // questions, not an allow/deny a tap could stand for.
+        Message.Kind.OPTIONS -> if (QuestionCardRules.drawsQuestionCard(message)) {
+            QuestionCardView(chat, message, haptics)
+        } else {
+            CardView(chat, message, haptics)
+        }
         Message.Kind.ACTIVITY -> ActivityChip(message.tool)
         Message.Kind.SCREEN -> ScreenShot(chat.threadId, message)
         // A message kind from a newer computer. Almost everything the harness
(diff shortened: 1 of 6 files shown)


milind-soni (changes requested):
Reviewed head 1b470f90e6dedb9653d4dc48527e706ab599f552. QuestionCard.kt:127 sets sent before calling session.answer; that call handles transport errors internally and never resets sent. A temporary connection failure therefore leaves the card looking answered and blocks a retry. Please settle from server acknowledgement or propagate a result and restore the card on failure, with a failed-send/retry regression.

Focused Kotlin question/core tests (8) and app question-card rules (3) pass, including compilation. This PR depends on #953's answer-encoding fixes; retarget to main afterward. Please add your author Signed-off-by trailer per CONTRIBUTING.md.

Issue #963, opened 2026-09-08
Android companion app has no FCM — scheduled routine/reminder notifications don't survive app being closed


buttonsjasper360-lang:
## Summary

The Android companion app has no push notification mechanism. `LocalNotificationPoster.kt` says so explicitly in its own header comment:

> "Local-only notifications from live/replayed notify frames. No FCM."

Notifications only fire while the app has a live connection to the desktop, or briefly after backgrounding via `SessionLingerService`, which is deliberately *not* a foreground service and only keeps the process alive for a short grace window (minutes) to avoid Android's cached-process reclamation — not to survive the app being closed for any real length of time.

This means a **scheduled routine** (`routines.json`, `schedule.type: "once"` or recurring) that's meant to notify the user at a specific future time — e.g. a reminder set for a week out — will only actually produce a notification if the Android app happens to be open (or very recently backgrounded) at the moment it fires. Otherwise the notify frame just sits there and gets "replayed" whenever the app is next opened, which defeats the purpose of a time-sensitive reminder.

For comparison, iOS companion notifications go through real APNs push (per `notification-and-proactivity-qa.md`), so this is an Android-only gap — the two platforms aren't at parity for anything that needs to reach the user away from the app.

## Why this matters

A core use case for OpenMausBot's routines/reminders is exactly this: "remind me about X later, even if I'm not near my computer or the app isn't open." Without FCM, that use case silently doesn't work on Android — there's no error, the routine just fires server-side and nothing reaches the phone. A user relying on this (e.g. migrating reminders off a calendar app into OpenMausBot specifically to reduce clutter) would have no visible indication that their reminder never arrived.

## Request

Add FCM-based push delivery to the Android companion app for notify frames (mirroring what APNs already does for iOS), so that:
- A routine/reminder notification reaches the phone as a normal system notification (with sound, like Telegram/Signal/etc.) regardless of whether the app is open, recently backgrounded, or fully closed.
- Behavior is at parity with the iOS companion app described in `notification-and-proactivity-qa.md`.

Happy to test against a real device (Samsung, Android) once there's a build to try.

## Relevant files
- `android/app/src/main/kotlin/com/openmausbot/companion/notifications/LocalNotificationPoster.kt`
- `android/app/src/main/kotlin/com/openmausbot/companion/lifecycle/SessionLingerService.kt`
- `docs/notification-and-proactivity-qa.md` (iOS APNs behavior, for comparison)

[...]

PR #966, opened 2026-09-08
Android: opt-in background connection so scheduled reminders survive the app being closed

buttonsjasper360-lang:
## Problem

The Android companion has no FCM (confirmed in `LocalNotificationPoster.kt`'s own header comment: "Local-only notifications from live/replayed notify frames. No FCM."). A notify frame only reaches the phone while the app has a live connection, or within the ~25s post-background grace window `SessionLingerService` buys.

That means a **scheduled routine/reminder** meant to notify the user later — the core use case for routines — silently does nothing if the app isn't open (or recently backgrounded) at the moment it fires. There's no error; the routine just runs server-side and nothing reaches the phone. See #963 for the original report.

## What this adds

An opt-in **"Background connection"** toggle in Settings (off by default):

- `AlwaysOnConnectionService`: a real foreground service (`foregroundServiceType="connectedDevice"`, not `dataSync` — that type faces a 6h/24h cap on Android 15+) that holds the SSE stream open indefinitely instead of only the 25s linger window. `stopWithTask="false"` so it survives the app being swiped from Recents; `START_STICKY` so the OS restarts it if it's ever killed under memory pressure.
- `AlwaysOnConnectionState`: a small main-thread-confined flag `SessionLingerController.onStop()` checks so the two mechanisms coordinate instead of racing — when the always-on service is running, the 25s linger timer is skipped entirely rather than potentially disconnecting a stream the service is trying to keep open.
- `AlwaysOnBootReceiver` restarts the service after a reboot if the user left it enabled; `OpenMausApp.onCreate()` covers the "process relaunched, not rebooted" case the same way.
- Battery-optimization exemption is requested once, only when the user turns the toggle on.

This is deliberately **not** a real push (FCM) implementation — it trades a persistent low-priority notification and some battery for reliability using only what the app already has (its own SSE connection over the paired route), no new external dependency. A real FCM implementation would still be the more battery-friendly long-term fix; happy to discuss if that's wanted instead of or alongside this.

## Also included

Bumped the "done" notification channel (bot replies / finished tasks) from `IMPORTANCE_DEFAULT` to `IMPORTANCE_HIGH`, so it pops up with sound like a normal messaging app instead of landing silently in the shade — the same silent-delivery problem as above, just for messages that *do* arrive while the app is reachable. Channel importance can only be set at first creation on a given device, so this ships as a new channel id (`bot_messages`) with the old `done` channel deleted on upgrade rather than trying to reconfigure it in place.

## Test changes

`SessionLingerWiringTest`'s manifest-wide "never contains FOREGROUND_SERVICE" assertion is narrowed to the per-element checks it already had on `SessionLingerService` specifically — that assertion was always a proxy for "the linger *service* doesn't need one," not a blanket ban on the app ever using `FOREGROUND_SERVICE` for a different, user-opted-in purpose. The four `assertEquals` checks on `SessionLingerService`'s own manifest declaration are untouched.

Full `:app` and `:core` unit test suites pass locally (56 tests), `:app:assembleDebug` succeeds, and I've installed and tested this on a real device (Samsung, Android) — service runs in foreground, notification posts, battery-optimization exemption grants correctly, and a test notification delivered to the "bot_messages" channel with the app fully closed.

## Summary by CodeRabbit

- **New Features**
  - Added an optional “Background connection” setting to keep the connection active when the app is closed.
  - Background connection can restart automatically after device reboot when enabled.
  - Added guidance about battery usage and persistent notifications.
  - Notification access is required before enabling background connection.

- **Improvements**
  - Completion notifications now use the bot messages notification channel.
  - Existing notification channel settings, including custom sounds, are preserved.
  - Updated notification wording to reflect bot messages more clearly.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -7,6 +7,17 @@
     <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <!--
+      Opt-in "always on" background connection (Settings toggle, default off).
+      FOREGROUND_SERVICE_CONNECTED_DEVICE rather than dataSync: this service
+      exists to keep the pairing to the paired computer alive, not to move
+      data in bulk, and dataSync foreground services face a ~6h/24h system
+      timeout on Android 15+ that connectedDevice does not.
+    -->
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_CONNECTED_DEVICE" />
+    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
+    <!--
       Pairing QR only, and asked for only when the scanner is opened. The app is
       fully usable without it: the NSD list and a typed address are the other two
       ways in (§6), so a camera the user never grants costs nothing.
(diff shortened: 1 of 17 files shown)


milind-soni (changes requested):
Please preserve existing notification choices when migrating the legacy done channel. LocalNotificationPoster.kt currently reads only its importance, creates bot_messages with defaults, then deletes done; a custom sound or intentionally silent sound=null, vibration, and lock-screen settings are lost. The latest fix protects an already-created bot_messages channel but not users upgrading from done. Retain the existing channel for those users or explicitly preserve its sound/audioAttributes, vibration and visibility behaviors, with silent/custom-sound upgrade tests. Also, all three author commits still need their genuine DCO sign-offs. Full CI currently awaits maintainer approval rather than having passed; holding approval until this migration is corrected. No physical-device verification was performed in this review.

PR #1009, opened 2026-09-09
The tool ladder: connect, find, or build what a job needs

aivsomkar:
From Omkar's recording of `hilo.cx/onboarding`: you say what you want done, and the connecting happens **in the conversation**, one step at a time, with a way out at every step.

A bot that needed a tool it did not have used to have one move — say so in prose and stop. Everything past that was the user's problem: find the integration, connect it in a settings panel, come back, ask again.

Hilo's flow has one rung. This has four, and the last three are the point: **a bot that cannot do something should not be a dead end.**

| Rung | When | What happens |
|---|---|---|
| 1 · **Use** | the toolkit is already connected | nothing is shown; the job runs |
| 2 · **Connect** | the catalog has it | picker → name the account → OAuth → resume |
| 3 · **Find** | not in the catalog | research an MCP/CLI and **propose** it |
| 4 · **Build** | nothing exists | generate one with `cli-printing-press`, and propose that |

**No rung is silent about failing.** Falling off one is a message naming what was missing and offering the next.

## Rungs 1–2

`need_tool("calendar")` — the model names a **capability** in plain words, never a vendor and never a slug. The harness answers with the apps it can actually connect, because a model-authored list can name a provider that does not exist or that we have no way to authorize, and a person cannot tell those from a real one until they have clicked it.

Then the account name — which we have always stored (`normalizeAccountAlias`, multiple accounts per toolkit) and had **never once asked a person for**. It is asked at the moment it means something, with the reason attached: it is the name the bot will say back to you when it asks to act.

The existing `ConnectorCard` and `maybeResumeConnectors` finish the job, so a connection reached this way is indistinguishable from one the model asked for by slug.

Escape hatches are load-bearing, not decoration. **"I'll connect it later"** resumes the bot with the truth — *they chose not to, carry on and say what you can't do* — rather than letting the turn time out into a guess.

## Rung 3 — and why the approval is real

This is the rung with teeth. "Search the web, find an MCP, install it" is mechanically a supply-chain attack: rank a malicious server for "google calendar mcp" and you have code on the user's computer holding whatever they connect to it.

So the bot **proposes**. It never installs, never runs, never writes a server.

- **A version range is refused before anyone sees the card.** `^2.0.1` can be a different program tomorrow, under the same click.
- **A proposal with no source is refused.** "The model said so" is not evidence anyone can act on.
- **The publisher is labelled as the page's claim, not ours.** We did not check it and the card must not lend it our credibility.
- **The consent names the consequence** — *this command will run on your computer, with access to what you connect* — rather than asking whether to "add an integration".
- **The approval is hash-bound to the exact command shown**, and the fingerprint covers only what *executes*: rewording the pitch cannot invalidate an approval; changing the program, version, arguments or environment must, and does.

`parseMcpServerMutation` still writes new servers **inert**, and that guard is untouched — it exists because a command added through the panel has been reviewed by nobody. This one has, so the approve route enables it explicitly. A necessary condition added, not a guard relaxed.

## Rung 4

`cli-printing-press` generates a CLI **and an MCP server** from an OpenAPI spec, URL or HAR file. The bot already has a shell, so this needed a prompt and a proposal shape rather than new machinery. It lands on rung 3's card, because from the user's side it is the same decision.

Generated code has no package and no publisher, and inventing either would be the one dishonest field on a card built to be honest — so its provenance is the **documentation it was generated from**, required, https, and shown as a link. The consent says who wrote it: *the bot generated this itself; nobody else has reviewed it.*

If `cli-printing-press` is not on the machine the bot says exactly that and stops. We do not install it for them.

## What testing changed

Omkar ran this against his real catalog and it was wrong three times. Each fix is a commit:

- **The matcher was tuned on the wrong data.** The curated fallback writes terse phrases (`"Analytics, feature flags, experiments"`) so I scored blurb matches by position. The live catalog writes sentences — `"PostHog is an open-source product analytics platform"` puts the word seventh — and the position gate dropped exactly the app he uses. Position is a ranking hint now, never a gate.
- **The catalog arrives sorted by usage and I threw that away.** Asked for email it offered Benchmark Email, BlueFox Email and Bulk Email Checker, and left **Gmail off the end**. For a category word, having it in your name is weak evidence; being the one everybody uses is strong.
- **`searchTerms` singularises before looking up synonyms**, and the table was keyed `"analytics"` — so every analytics synonym was dead code.
- **Rung 1 only checked Composio**, so a tool installed by rung 3 or 4 could never satisfy it. It counts the bot's own enabled MCP servers now.
- **The prompt described a principle and the model did not recognise the moment it was in.** Asked to pull PostHog data it requested an API key in chat; asked to log a weld inspection it asked which system to use. It now names those moments as things to do *instead of* the message the bot was about to write.

## Setup mode

The plan called for suggested first-job chips. Building them fought #833, which deliberately replaced the onboarding quiz with setup mode — a test says so in its name. The chips are **not** here. What was actually missing was the end of the setup prompt, which sent people to Settings → Access to authorize apps by hand. It calls `need_tool` now.

## Platforms

|---|---|---|
| macOS     | yes | in this PR, tested locally (OMB2) |
| Windows   | yes | in this PR (shared server + desktop UI, no platform gate); NOT built on Windows — `Package Windows` run started on this branch |
| iOS       | yes | follow-up: the cards decode fine but carry no `requestId`, so they render inert — title and subtitle, no buttons. Port in progress |
| Android   | yes | same as iOS. Port in progress |
| Companion | yes | in this PR — the allowlist is default-deny, so the phone ports need these entries and both would collide on the one file |

- [ ] iOS: picker, naming and proposal cards in `ios/App`
- [ ] Android: the same in `android/app/.../ui`

## Test plan

- Full suite **5160 passed**, 39 skipped. `typecheck`, `lint`, `i18n:check` clean.
- `shared/tool-request.test.ts` — matching, ranking, the score floor, and a catalog built in usage order with the category-named apps two hundred entries down, which is the shape that actually broke.
- `shared/tool-proposal.test.ts` — version ranges and sourceless proposals refused; the fingerprint changes when the program changes and not when the words do.
- `server/index.test.ts` — capability → real slugs; an app the card never offered is refused; approving without echoing the displayed hash is refused; nothing is installed until it is approved; a generated tool needs its documentation.
- `companion/test/routes.test.ts` — a paired phone may answer a card, never invent a verb on one.
- Manual: packaged as OMB2 and run against real bots on macOS across several rebuilds.

## Known limits

- **This is a text heuristic and it will keep having gaps.** `chat` does not find Slack: its blurb says "messaging", the matcher looks for "message", and the stemmer does not join them. The proper fix is the category metadata Composio returns and our catalog mapper discards.
- **Tool choice belongs to the model.** The prompt is much harder to miss now, but a small model (Haiku 4.5) skipped it — along with the pre-existing rule about never asking for credentials in chat.
- **Nothing re-checks an approved server.** A package pinned today can be yanked or change hands, and we would not notice.

## Summary by CodeRabbit

- **New Features**
  - Added in-chat assistance for connecting apps needed to complete a task.
  - Added connectable app suggestions with guided selection, account naming, connection, defer, and follow-up options.
  - Added tool proposal cards for reviewing MCP servers, command-line tools, and generated tools before approval.
  - Added clear proposal details, sources, commands, approval/decline actions, and status updates.
  - Added a visible catalog of apps that can be connected from eligible conversations.


shared/tool-proposal.ts
```diff
@@ -0,0 +1,143 @@
...
+      return "a generated tool needs the https documentation it was built from";
+    }
+    return null;
+  }
+  if (!proposal.packageId?.trim()) return "a proposal needs the package or repository it comes from";
+  // A range is not provenance: "^1" can become a different program tomorrow,
+  // and the approval was for what was on screen today.
+  if (!proposal.packageVersion?.trim() || /[\^~*x]|latest/i.test(proposal.packageVersion)) {
+    return "a proposal needs one exact version, not a range or \"latest\"";
+  }
```
milind-soni, line 135: [P1] Bind the approved version to the executable. This accepts packageVersion >=1.2.3, and also accepts version 1.2.3 paired with npx args selecting @latest. Both produced approvable cards in the isolated fixture. Hashing the proposal does not pin what a later invocation downloads. Require an exact version and verify supported launcher arguments actually select that package/version; reject unsupported or unpinned forms.

server/index.ts
```diff
@@ -13593,6 +13779,157 @@ const handleRequest = async (req: IncomingMessage, res: ServerResponse) => {
...
+      if (fresh !== reviewed) return json(res, 409, { error: "this proposal changed since it was shown" });
+
+      const name = mcpServerNameFor(proposal.packageId);
+      const servers = { ...(cfg.mcpServers ?? {}) } as Record<string, unknown>;
+      if (servers[name]) return json(res, 409, { error: `an MCP server called ${name} already exists` });
+      const stored = parseMcpServerMutation(name, { command: proposal.command, args: proposal.args, env: {} });
+      if (!stored.ok) return json(res, 400, { error: stored.error });
+      // parseMcpServerMutation writes a new server INERT on purpose — a
+      // command added through the panel has been reviewed by nobody. This one
+      // has: the user just approved this exact command with its provenance in
```
milind-soni, line 13828: [P1] Do not install ordinary CLI proposals as enabled MCP servers. The route ignores proposal.kind: approving kind=cli persisted its command as an enabled MCP entry. The consent describes using a CLI when needed, but MCP entries are mounted on subsequent supported turns and can run for other bots. A normal CLI cannot answer the MCP handshake and may repeatedly perform its operation. Provide a separate CLI installation path or reject this kind until supported.

server/index.ts
```diff
@@ -13593,6 +13779,157 @@ const handleRequest = async (req: IncomingMessage, res: ServerResponse) => {
...
+      const echoed = typeof body.reviewedSha256 === "string" ? body.reviewedSha256 : "";
+      // A card too old to carry a hash stays decline-only rather than
+      // becoming an approval nobody can prove the contents of.
+      if (!reviewed) return json(res, 409, { error: "this proposal cannot be approved — ask the bot to find it again" });
+      if (echoed !== reviewed) return json(res, 409, { error: "this proposal changed since it was shown" });
+      const fresh = createHash("sha256").update(executableFingerprint(proposal)).digest("hex");
+      if (fresh !== reviewed) return json(res, 409, { error: "this proposal changed since it was shown" });
+
+      const name = mcpServerNameFor(proposal.packageId);
+      const servers = { ...(cfg.mcpServers ?? {}) } as Record<string, unknown>;
```
milind-soni, line 13822: [P2] Give generated tools distinct installation identities. Generated proposals omit packageId, so they all become found-tool here. In the real-server fixture the first approval succeeded and the second distinct generated tool returned 409. Derive a collision-safe identity from the proposal instead of using the same fallback for every generated tool.

src/components/ChatView.tsx
```diff
@@ -747,6 +753,14 @@ const MessagesList = memo(function MessagesList({
             case "connector":
               return m.connector ? <ConnectorCard botId={bot.id} threadId={bot.threadId} message={m} /> : null;
             case "options":
+              // The tool ladder's own card: not a decision about an action,
+              // so never the approval box.
+              if (m.card?.toolRequest) {
+                return <ToolRequestCard threadId={bot.threadId} message={m} />;
+              }
```
milind-soni, line 760: [P2] Render these approval cards in room conversations too. The server accepts room-owned tool requests/proposals and asks the agent to wait, but only ChatView handles the new fields. GroupView still recognizes only the existing permission cards, so room users get no usable approve/decline UI and cannot continue the request.

milind-soni (changes requested):
Reviewed against main 168753ef. Existing focused tests pass, but a disposable real-server fixture reproduced the issues below. All nine contributor commits also still need author sign-offs. Please address these before merging; no live workspace or credentials were used.

PR #1018, opened 2026-09-09
Phones show who opened a thread and open it from the chip

milind-soni:
## In plain terms

Phones learn about threads a bot opened. Part of thread-aware bots (`docs/plans/2026-09-09-thread-aware-bots.md`); the server tools and the desktop chat links land in sibling PRs. Both fields are optional on the wire, so this is safe against older servers and can merge first.

- Thread lists show a quiet "opened by Scout" label when a bot opened the thread.
- The "Opened thread #Title on Scout" chip is tappable and goes to that thread, using the same switch route a thread row uses. Viewing never redirects running work. A thread that is gone lands on the bot with a short notice; an unknown bot likewise.
- Decoding tolerates the fields being absent, with fixtures from real payloads asserting nothing regresses.

## Test plan

- [x] iOS: `swift test --package-path ios` 365 tests green; unsigned simulator-target build succeeds. Decoding tests cover with-field, without-delegation-id and without-field; mutation-checked (computed nil → 5 assertions red).
- [x] Android: core 496 and app 862 unit tests green, debug APK builds. Session test drives the chip through the real switch route with a mock server, including the 404 fallback and the no-request case when already on the thread; mutation-checked (`@Transient` → 2 red).
- [x] Companion sidecar already allows the switch route (`companion/src/routes.ts`), asserted by its existing test; no change needed.

No live device or simulator run.

## Summary by CodeRabbit

- **New Features**
  - Activity messages can now open referenced threads by tapping their thread chips.
  - Navigation switches to the relevant conversation or opens another bot’s chat when needed.
  - Threads opened by another bot now display an “opened by” label in thread lists and task details.
  - Missing threads fall back gracefully with an explanatory error message.
  - Added selection haptic feedback when opening a thread chip.
- **Bug Fixes**
  - Improved handling of unavailable bots and deleted threads during thread navigation.

### DIFF
--- a/android/app/src/main/kotlin/com/openmausbot/companion/ui/ChatScreen.kt
+++ b/android/app/src/main/kotlin/com/openmausbot/companion/ui/ChatScreen.kt
@@ -109,6 +109,7 @@ import com.openmausbot.companion.core.Dictation
 import com.openmausbot.companion.core.DisplayedMessageAttachment
 import com.openmausbot.companion.core.DownloadedFile
 import com.openmausbot.companion.core.Message
+import com.openmausbot.companion.core.ThreadRef
 import com.openmausbot.companion.core.TranscriptRow
 import com.openmausbot.companion.core.target
 import com.openmausbot.companion.core.transcriptRows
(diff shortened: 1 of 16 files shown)


milind-soni:
Review follow-up: this merged while the mobile review was in progress. Three reproduced edge cases are fixed in #1020: non-404 failures opening the wrong thread, a late hydration from the previous computer changing current state, and late switch responses navigating/updating the wrong computer. The follow-up preserves the 404 fallback and adds cancellation/connection guards plus regressions. Android500 core +43 focused app tests and unsigned iOS app compilation passed; no live phone data used.

PR #1023, opened 2026-09-09
fix(chat): let an inline code span break so it cannot leave the bubble

ihabmurad:
## What

An inline code span can be longer than the bubble is wide, and an unbreakable
token has nowhere to go but outside it. Links in `ChatMarkdown.tsx` already
carry `break-words`; inline code did not.

Reported from real use, where a bot reply contained:

    `dist/{download,privacy,terms,license,support,presskit,changelogs,docs,about,feedback}`

— 85 characters, no break opportunity.

## Not a regression from #994

Worth stating plainly, since that is where it was noticed. I rendered the same
paragraph through the current component and through the one from `cb4e458b`
(immediately before the bidi work) at identical bubble geometry, and measured
the overflow:

| | overflow |
| --- | --- |
| current — Arabic paragraph (RTL) | **87px left** |
| `cb4e458b` — same Arabic paragraph | 87px right |
| current — English paragraph (LTR) | 87px right |
| `cb4e458b` — same English paragraph | 87px right |

Same magnitude in all four. The bug predates #994; what that PR changed is the
side it escapes on. A right-to-left line ends on the left, so the token now
spills off the left edge of the message column instead of the right — which
reads as text escaping the message rather than a long line running on, and is
why it surfaced now.

## How

`break-words` on the inline code span, matching what the anchor and
`LocalFileLink` in the same file already do. `overflow-wrap: anywhere` also
zeroes the overflow; `break-words` was chosen for consistency with the
existing utilities here.

Fenced blocks are unaffected — `CodeBlock` already has `overflow-x-auto` and
its own wrap toggle.

## Verification

Measured in the browser against a control: with the change, the current
component overflows 0px in both the RTL and LTR paragraphs, while the
extracted `cb4e458b` component rendered beside it still overflows 87px — so
the difference is attributable to this change and not to the harness.

The new test fails without the fix (`1 failed | 34 passed`) and passes with it.
Full `vitest run src/`: 144 files, 1171 tests, all passing. `tsc -b` clean,
`oxlint` clean on the touched files.

### DIFF
--- a/src/components/ChatMarkdown.test.ts
+++ b/src/components/ChatMarkdown.test.ts
@@ -401,6 +401,17 @@ describe("bidi: message content carries its own direction", () => {
     expect(fenced).toContain('<div dir="ltr"');
   });
 
+  it("lets an inline span break, so a long path cannot leave the bubble", () => {
+    // an unbreakable token wider than the bubble has nowhere to go but
+    // outside it, and in an RTL paragraph that is off the left edge, where
+    // the line ends — the direction that reads as text escaping the message
+    const html = renderToStaticMarkup(createElement(ChatMarkdown, {
+      text: `${ARABIC} \`dist/{download,privacy,terms,license,support,presskit,changelogs,docs,about,feedback}\` ${ARABIC}`,
+    }));
+    const inline = /<code [^>]*class="([^"]*)"/.exec(html)?.[1] ?? "";
+    expect(inline).toContain("break-words");
+  });
+
   it("gives links a base direction, not isolation alone", () => {
     // isolate keeps a link from disturbing the sentence around it, but the
     // link's own contents still lay out along its inherited direction — a URL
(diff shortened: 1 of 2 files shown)


ihabmurad:
The only failing job, `package + smoke (Ubuntu 24.04 x64)`, fails in `Install package validation and native smoke tools` — `apt-get` hits a `Hash Sum mismatch` on Google's Chrome APT index (`dl.google.com/linux/chrome-stable`), exit 100, before anything in this repo is built. Same filesize and timestamp, different SHA256: a stale index served mid-update.

Nothing here touches packaging — the change is one `break-words` class on an inline `<code>`. All three `typecheck + test` jobs pass, as do the Android and iOS builds. Could you re-run that job when convenient? I don't have rerun rights on the repo.

milind-soni (approved):
Reviewed against current main. The change is scoped to inline code; fenced code keeps its separate scroll/wrap behavior. Combined chat/code-block tests pass, renderer typecheck and lint pass. In an isolated real-renderer preview, 280px English and Arabic bubbles both had scrollWidth = clientWidth (zero overflow). Retried the Linux packaging job after confirming the original failure was an upstream apt index checksum mismatch; the entire CI run is now green. Ready to merge.

PR #1025, opened 2026-09-09
ci: run oxlint with --deny-warnings in the required check

milind-soni:
## What

`pnpm lint` ran in no workflow and `.oxlintrc.json` only held ignore patterns, so the linter could never gate a merge. This adds `pnpm lint` to the `test` job right after typecheck on all three OSes, flips the script to `oxlint --deny-warnings .`, and fixes the 31 baseline warnings so the gate starts green.

- 14 `no-useless-spread` sites iterate live listener/session Sets whose loop bodies mutate them; kept the snapshot as `Array.from(...)` (the repo's existing idiom) rather than iterating live. Dropped the spread only where the collection provably cannot change under the loop.
- Targeted `oxlint-disable-next-line` with a reason for the five control-character regexes outside `server/`, one intentional variation-selector class, and one test that records `this`.
- Removed `browserConnectionStore` in `electron/main.mjs`: dead since #850 removed the Electron browser surface, the factory is lazy, nothing else references it.
- CONTRIBUTING gains one checklist line.

## How it was verified

```
pnpm lint                 → exit 0 (no output)
pnpm typecheck            → exit 0
vitest run <18 adjacent test files> → 18 passed, 615 tests passed, 2 skipped
pnpm check:electron       → Syntax-checked 98 Electron modules
pnpm control-plane:test   → 2 files, 42 tests passed
```
Mutation check: a scratch file with one `no-useless-spread` warning passes `oxlint` and fails `oxlint --deny-warnings` (exit 1), so the gate really denies.

No behavior change intended. No lockfile changes. No new dependencies.

## Summary by CodeRabbit

- **Improvements**
  - Improved scheduled routine reliability, including result tracking, recovery after deleted or repurposed conversations, and clearer unread-status handling.
  - Installation responses now provide more specific unsupported-operation messaging.

- **Chores**
  - CI now runs lint checks alongside typechecking on all supported operating systems.
  - Lint warnings are now treated as errors.
  - The contribution checklist now includes a lint verification step.

- **Refactor**
  - Simplified internal processing without changing application behavior.
  - Removed unused desktop connection setup.

- **Tests**
  - Updated test and lint annotations to support stricter validation.


milind-soni:
Integrated current main and preserved the peer-approval drain snapshot with Array.from, matching the original iteration behavior while satisfying the lint rule. Final head 5fa019998dd13955020156c8522284f6197d9ac4: strict lint, app/server typechecks, all 98 Electron syntax checks, 482 focused tests (2 skipped), and 2 native desktop-header tests passed on Node 24. A disposable control-omb fixture completed doctor → new-bot → send → settled → transcript with the expected fake-engine response and was torn down. A temporary warning probe confirmed plain oxlint permits a warning while --deny-warnings fails; the probe was removed. Fresh CI is running. No dependency or release changes.

PR #1031, opened 2026-09-09
Stop a conversation from paying for context it never asked for

aivsomkar:
A 35-message food-ordering chat cost 21.6M tokens. Not because of the model, the soul, or the number of turns — because of what the harness lets into the model's context and then re-reads on every call. By the end each call was carrying 330,000 tokens.

Four independent fixes, measured rather than estimated. The [full teardown is here](https://claude.ai/code/artifact/b97547ad-6b96-45dc-ad5b-eee1470be2d7).

## Where the tokens went

| Bucket | Tokens | At Sonnet 5 list price |
|---|---|---|
| Cache reads | 19.6M | $3.92 |
| Cache writes | 1.96M | $4.91 |
| Output | 25K | $0.25 |

87 model calls for 35 messages. On a vendor-CLI engine the harness sends only the new sentence and a resume id — the CLI owns the session, runs the MCP servers itself, and pastes their raw answers into the file it re-sends every call. So none of this is visible from inside the harness, and none of it is fixable there either.

## 1. A bot no longer inherits this machine's Claude Code setup

Every bot turn also carried the desktop's own MCP servers and claude.ai connectors (407 tools on the measured machine), its 78-entry skill listing, its agent listing, hooks, and personal CLAUDE.md. A food-ordering bot paid for Gmail, Slack and Figma tools 87 times, and had them reachable.

`--strict-mcp-config` + `--setting-sources project`. Strict mode makes the harness config the CLI's only MCP source, which would also drop a server the bot's *own* project declares, so `<cwd>/.mcp.json` is now forwarded verbatim (http/sse entries included) and merged last — it can add servers, never shadow a harness mount. Project servers are not pre-allowed; they ride the permission broker like any custom server.

**Measured: 38,171 → 31,774 input tokens on one trivial turn, before any tool runs.**

`OMB_CLAUDE_INHERIT_USER_CONFIG=1` restores the old launch for a bot that genuinely depended on a user- or local-scope server.

## 2. A memory edit no longer re-uploads the whole session

The driver keys its per-thread CLI process on the spawn contract, which includes the system prompt. MEMORY.md is pasted into that prompt, so saving a memory changed the contract, relaunched the process, and made the provider re-cache the entire conversation at the write rate. Mentions did it on any turn that tagged a bot. Two of the six full re-uploads in the measured thread were this, ~400K written tokens.

The prompt is now reported in two halves: `systemStable` and `systemVolatile` (memory + mentions). The driver keys on the stable half; a live session is told what changed inside the next turn, in the CLI's own out-of-band form. `system` is unchanged, so every other driver reads exactly what it read before.

## 3. A gate between the bot's tools and the CLI

One Instamart search returns 60–140 KB of JSON; a menu 70–105 KB. Four searches in one turn added 67,000 tokens that were then re-read ~80 more times.

`server/mcp-gate.ts` is a pass-through MCP server. Every frame is relayed untouched except the response to a `tools/call`, where an oversized result is cut to a budget. The cut keeps **whole records** — the payload is parsed and its bulk arrays truncated, so the model is never handed half an object — and `structuredContent` is trimmed the same way rather than left as a full second copy. A non-JSON payload falls back to a text cut that says so. A result the trimmer cannot handle is relayed whole: costing context is recoverable, dropping a tool's answer is not.

Only bot-owned servers are gated. Harness-owned mounts (permission broker, computer, browser, agents, dweb) are already bounded and stay direct.

**Measured end to end against a live CLI turn with a 136 KB result: 129,809 → 98,087 tokens, identical answer.**

### One result worth reading

The first version told the model where the untrimmed result had been saved, so nothing would be lost. Measured, the model read that file straight back in and the gated turn cost **more** than no trimming at all — 210,913 tokens against 196,183. A recovery path offered to a model is not a safety net, it is a slower way to pay for it. The marker now says to call the tool again with a narrower query. The file is still written, for the person and the harness; `OMB_GATE_SPILL_HINT=1` puts the path back for debugging.

`OMB_MCP_RESULT_BUDGET` sets the budget; `0` mounts bot servers directly.

## 4. The harness picks the compaction window

The CLI compacts when it nears a window, and left alone that window is the model's own — so a Sonnet 5 session runs toward a million tokens before anything happens. At 330K the measured session looked perfectly healthy to it.

`--autocompact 200000` by default, delegating the compaction itself to the CLI, which owns the session and already has a summarizer for it. With tool results gated most threads never reach it; this is the backstop. `OMB_CLAUDE_AUTOCOMPACT` takes a count, `auto`, or `off`; a configured value is clamped to the 100k–1M the CLI accepts, because out of range is an argument error that would fail every turn rather than degrade.

## Platforms

|---|---|---|
| macOS     | yes | in this PR; full suite green, and verified end to end against a real `claude` process. NOT built as OMB2 yet |
| Windows   | yes | in this PR (shared server code, no platform gate). The parity check caught a real Windows bug on the way: the gate spawns the bot's MCP server itself, and `npx -y mcp-remote …` is an npm `.cmd` shim CreateProcess cannot exec — it now resolves through `resolveCliSpawn` like the drivers do, with `windowsHide` and a `taskkill /T` reap. NOT built on Windows |
| iOS       | yes | n/a: no `/api/*` route, event payload, or wire shape changed. `SendTurnInput` is an internal driver contract |
| Android   | yes | n/a: same reason |
| Companion | yes | n/a: no phone-called route changed, so no `routes.ts` allowlist entry is needed |

## Test plan

- `vitest` 5,314 passed / 39 skipped, `broker:test` 8, `test:electron` 155, all on Node 24.14.1
- `test:packaged-server` — the only thing that exercises the esbuild bundle: all **12** spawned proxy paths resolve inside the packaged server dir, now including `mcp-gate`
- `oxlint` 0 errors, `tsc -p tsconfig.server.json` clean
- New: `server/mcp-trim.test.ts` (10 cases, on the shapes real MCP servers return), `server/mcp-gate.test.ts` (8, the gate as a real process with a scripted upstream), plus driver cases for isolation, project `.mcp.json`, the gate wiring, the volatile split, and the compaction window
- Live comparison with a toy MCP server returning 136,091 chars, run through a real `claude` turn, gated vs direct

**Not yet done:** no OMB2 build, so this has not been driven by hand through the Swiggy bot. Worth doing before merge — the change most likely to surprise is that a bot no longer sees this machine's personal MCP servers or global CLAUDE.md.

## Summary by CodeRabbit

* **New Features**
  * Added safeguards that limit oversized MCP tool results while preserving structured data and providing access to full output when needed.
  * Added automatic compaction controls for Claude sessions.
  * Claude sessions now remain active when only memory or mention context changes, reducing unnecessary restarts.
  * Claude integrations now use project-specific MCP settings by default and support forwarding project MCP servers.
* **Bug Fixes**
  * Improved isolation from local Claude Code configuration to ensure more consistent bot behavior.

### DIFF
--- a/scripts/bundle-server.mjs
+++ b/scripts/bundle-server.mjs
@@ -58,6 +58,7 @@ const ENTRY_POINTS = [
   "vps-container-mcp.ts",
   "permission-proxy.ts",
   "connector-proxy.ts",
+  "mcp-gate.ts",
   "browser-proxy.ts",
   "drivers/agents-proxy.ts",
   "drivers/dweb-proxy.ts",
--- a/server/contracts.ts
+++ b/server/contracts.ts
@@ -187,7 +187,7 @@ export interface SendTurnInput {
    * (memory today): `systemStable` is everything else, `systemVolatile` is
    * those sections' text. A driver that keeps one CLI process per thread keys
    * that process on the stable half, so a memory edit no longer respawns the
-   * session and make the provider re-cache the entire prompt; the changed half
+   * session and makes the provider re-cache the entire prompt; the changed half
    * is delivered inside the next turn instead. Drivers that rebuild their
    * request every turn ignore both and keep reading `system`. */
   systemStable?: string;
--- a/server/contracts.ts
+++ b/server/contracts.ts
@@ -183,6 +183,15 @@ export interface SendTurnInput {
   transcript?: Array<{ role: "user" | "assistant"; text: string }>;
   /** Bot persona (name/title/description) as a system prompt. */
   system?: string;
+  /** `system` split at the sections that legitimately change mid-conversation
+   * (memory today): `systemStable` is everything else, `systemVolatile` is
+   * those sections' text. A driver that keeps one CLI process per thread keys
+   * that process on the stable half, so a memory edit no longer respawns the
+   * session and make the provider re-cache the entire prompt; the changed half
+   * is delivered inside the next turn instead. Drivers that rebuild their
+   * request every turn ignore both and keep reading `system`. */
+  systemStable?: string;
+  systemVolatile?: string;
   /** Per-bot integrations the driver may hand to the agent as tools. */
   integrations?: {
     /** A local stdio bridge owns the remote Composio transport. Keeping the
--- a/server/drivers/claude.test.ts
+++ b/server/drivers/claude.test.ts
@@ -630,6 +630,81 @@ describe("ClaudeDriver turns (fake CLI)", () => {
     expect(seen.mcpConfig.mcpServers.ogb.alwaysLoad).toBe(true);
   });
 
+  it("launches isolated from the machine's own Claude Code configuration", async () => {
+    await create();
+    const dump = join(scratch, "isolation.json");
+    process.env.FAKE_CLAUDE_DUMP = dump;
+
+    await instance.adapter.sendTurn({ threadId: "t-isolate", text: "hi" });
+    await recorder.until((e) => e.type === "turn.completed");
+
+    const seen = JSON.parse(readFileSync(dump, "utf8"));
+    // Without these two the CLI also mounts the desktop's own MCP servers
+    // and connectors, and lists the desktop's skills, on every turn of every
+    // bot — tens of thousands of tokens per model call that no bot asked for.
+    expect(seen.argv).toContain("--strict-mcp-config");
+    expect(seen.argv[seen.argv.indexOf("--setting-sources") + 1]).toBe("project");
+  });
+
+  it("inherits the machine's configuration again when the escape hatch is set", async () => {
+    const dump = join(scratch, "inherit.json");
+    await create(undefined, { FAKE_CLAUDE_DUMP: dump, OMB_CLAUDE_INHERIT_USER_CONFIG: "1" });
+
+    await instance.adapter.sendTurn({ threadId: "t-inherit", text: "hi" });
+    await recorder.until((e) => e.type === "turn.completed");
+
+    const seen = JSON.parse(readFileSync(dump, "utf8"));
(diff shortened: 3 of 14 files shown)


milind-soni:
Thank you for this — the teardown is the kind of measurement this codebase has been missing, and the four fixes hold up on re-review against current `main`: the volatile/stable split leaves `system` untouched for every other driver, the gate keeps whole records, and the compaction clamp is exactly right. Checks are green and the branch merges cleanly.

Two follow-ups, which I've put on a branch stacked on this one rather than pushing to yours: #1037.

1. **A version floor for the new flags.** `--strict-mcp-config`, `--setting-sources` and `--autocompact` are unknown-flag argument errors on an older CLI, which would fail every turn rather than degrade, and the harness had no floor anywhere. Bisecting the published binaries (the changelog never records `--autocompact`): `--strict-mcp-config` ≤ 1.0.60, `--setting-sources` 1.0.122, `--autocompact` 2.1.122. #1037 reads `claude --version` once per instance, passes each flag only to a CLI that accepts it, and surfaces the existing `ProviderSnapshot.update` notice on the Engines page below 2.1.122.
2. **A release note** for the isolation change in `docs/custom-mcp-servers.md`: what a bot stops seeing, the supported fix, and `OMB_CLAUDE_INHERIT_USER_CONFIG=1`.

The intent is to merge this PR and then #1037. If you'd rather fold either commit into this branch, cherry-pick freely — they're `3d6cc54c` and `8a1e8127`.

milind-soni:
Small revision to #1037 after its first CI run: the version is now read by `snapshot()` (which the harness calls whenever it lists engines) rather than probed inside the turn path, so no extra CLI start-up per first turn and no shift in recorded call order. The two commits are now `006e8d1c` and `adab869e` if you want to cherry-pick.

PR #1059, opened 2026-09-10
fix(usage): enforce spend caps for room turns

casquijo:
## What changed

- Capture immutable usage attribution when each room member is dispatched and record it when the provider turn completes.
- Recheck the workspace spend cap immediately before room and calendar provider dispatches.
- Preserve attribution across queued messages and identify routine-driven room goals correctly.
- Treat capped room goals as blocked instead of retrying them as provider failures.
- Add regression coverage for rooms, goals, queued users, and calendar calls.

## Why

Direct turns were accounted for and capped, but provider-backed room and calendar turns could proceed without equivalent ledger entries or dispatch-time enforcement.

This ensures settled room usage is visible before the next member dispatch and that all affected paths observe the same monthly cap.

## How it was verified

- `pnpm typecheck`
- `pnpm lint`
- Targeted spend-cap tests: 6 passed
- Affected and review suites: 147 passed, 1 todo
- `pnpm test`
  - Vitest: 461 files passed, 2 skipped; 5,731 tests passed, 40 skipped, 1 todo
  - Broker tests: 8 passed
  - Electron tests: 164 passed
  - Packaged-server smoke passed

## Screenshots (UI changes)

Not applicable — no UI changes.

## Checklist

- [x] No `dist-server/` edits
- [x] No macOS-only code or shell command construction added

### DIFF
--- a/docs/verification/spend-cap.md
+++ b/docs/verification/spend-cap.md
@@ -3,10 +3,11 @@
 ## Sub-features
 
 - Refuse every new turn once the month's reported cost reaches the workspace
-  cap: the message routes answer 409 with `code: "spend_cap"`, and a routine,
-  peer hop or webhook stops in `startTurn` with the same refusal.
+  cap: the message routes answer 409 with `code: "spend_cap"`; routines, peer
+  hops and webhooks stop in `startTurn`; and every room member, chained mention,
+  goal step, retry, queued send or calendar call checks again at dispatch.
 - Count a turn the moment it settles, not when the ledger's append lands or a
-  cache expires.
+  cache expires, including each provider-backed room or calendar member.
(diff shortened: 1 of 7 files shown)


milind-soni:
Triage update: #1135 is now merged, so the room spend-cap guard, room ledger entries, blocked-goal outcome, and live budget fields already exist on main. This PR is not wholly redundant: capturing the initiating user on each operation/queued item still matters. Current main reads turnTriggers at settlement, which a later queued sender can overwrite. Please rebase this as the narrower immutable-attribution follow-up, preserving the queued-user, calendar-owner, and routine-goal regressions, without adding a second spend/ledger path. Holding the current overlapping version rather than closing it as a duplicate.

PR #1083, opened 2026-09-11
fix(openai-compat): replace hardcoded absolute timeout with renewable idle stream timeout

silvestreaac:
### Problem
In the current OpenAI-compatible driver, turn requests enforce a static, hardcoded absolute timeout (`AbortSignal.timeout(options.timeoutMs)` with 120s default). Long-running streaming turns (such as deep reasoning models or verbose tool/text generations) are aborted midway even while stream chunks are actively arriving with zero idle lag.

### Solution
1. Replace the static single timeout in `createOpenAIChatRuntime` with a renewable idle timer (`resetIdleTimer`) that resets on each received streaming SSE chunk.
2. Maintain standard abort signals when idle period elapses or upstream disconnects.
3. Expose `OPENMAUS_OPENAI_COMPAT_IDLE_TIMEOUT_MS` (default: 180s) for customizable idle duration.
4. Add unit test coverage in `openai-compat.test.ts` verifying:
   - Abort when stream goes idle beyond threshold.
   - Successful completion when stream duration exceeds 120s but receives regular active chunks.

### Verification
- Ran `npx vitest run server/drivers/openai-compat.test.ts` (18/18 PASS).

## Summary by CodeRabbit

- **Bug Fixes**
  - Streaming responses now remain active while chunks continue arriving, preventing premature interruptions during longer completions.
  - Incomplete streams are reported as interrupted instead of being treated as successfully completed with partial content.
  - Improved handling of stalled, interrupted, malformed, and unsuccessful responses to prevent hangs or false-success results.

- **Configuration**
  - OpenAI-compatible connections now support configuring the streaming idle timeout, with a default of three minutes.


milind-soni:
Maintainer triage update: pushed cea87148 after merging current main. Fixed the premature timer cleanup around HTTP-error and non-stream JSON body reads; removed an abort-listener leak; EOF before the completion marker now reports interruption. The renewal test now honors cancellation and lasts 300 simulated seconds, so it fails an absolute 180-second timer. Restored the unrelated provider-routing test and removed the unrelated client deployment record from this PR. Added real loopback HTTP tests for stalled JSON, stalled error body, truncated SSE, and explicit cancellation. 37 focused OpenAI-compatible/Grok/MiniMax tests passed, plus typecheck and lint. Fresh cross-platform CI is the merge gate.

milind-soni:
The idle timer is wired in the right place: the single outer `finally` at `server/drivers/openai-chat.ts:187` clears it on every exit path, including the `response.text()` and `response.json()` reads, and the new `openai-compat-idle.test.ts` drives a real loopback socket rather than a stubbed fetch. (The `omits provider routing when none is configured` test is back too — thanks for restoring it.)

Two things block it.

**1. The idle abort is named `AbortError` (`openai-chat.ts:102`).** That is the same name a user's stop produces (`interruptTurn` → `abort.abort()`), so `aborted` at `openai-chat.ts:253` is true, no `runtime.error` is emitted, and the turn settles as `stopReason: "interrupted"`. Someone whose model stalls for three minutes sees "interrupted" and no reason why. On main the abort comes from `AbortSignal.timeout`, whose reason is a `TimeoutError`, so the message reaches the user — and since grok shares this runtime with `retryScale` set (`grok.ts:51`), `classifyError` matches timeout as transient (`retry.ts:46`) and retries. Both behaviours go away. Abort with `"TimeoutError"` and the existing error path handles it.

**2. A stream closing without `[DONE]` now fails the turn (`openai-chat.ts:148-151`).** Main does `if (done) break;`, returns what arrived and emits `item.completed`. Here the partial reply is discarded and the turn ends `ok: false`; the `truncated` fixture at `openai-compat-idle.test.ts:21` locks that in. If failing is intended, raise a real error instead of an AbortError; otherwise keep main's behaviour.

Worth a line in the description as well: this is the shared runtime, so grok (120s) and minimax (180s) move from absolute to idle deadlines while only openai-compat gets the env knob.

CI on `cea8714` is still running — the green run in the checks list is from an earlier commit. The red Vercel check is a deploy-authorization prompt, not your change.

Push those two and I will re-review.

PR #1087, opened 2026-09-11
test(server): prove adding people end to end

milind-soni:
## What

An end-to-end test for Settings → People, plus its verification recipe.

`server/people-invite.test.ts` boots the real server with no sign-in list and a stubbed control plane, then walks the requests the People card sends:

- the owner (loopback) names the first admin; until then email sign-in is off (404) and the descriptor says so
- that admin signs in with the emailed code and invites a member through the same `PUT /api/config { signIn }` the card uses; the card's other reads (sessions, usage by person, public address) answer in the shape it renders
- the member's invite link serves the sign-in page; the member gets a chat-only cookie session, sees bots, and cannot change the list (403)
- a promotion applies to the next sign-in while the device already issued keeps its scopes
- a removal refuses new sign-ins (403) while the old devices keep working until an admin revokes them (then 401)
- a `@domain` entry welcomes everyone there and nobody at a look-alike domain

`docs/verification/people.md` documents the recipe and the browser walk-through, and the index links it.

## Proof

- Mutation-checked: members never welcome → 4 tests fail; revoke made a no-op → 1 fails; saver drops `signIn` → 4 fail.
- The served UI was also driven once through headless Chrome (owner invites from the card, a second browser through a forwarding proxy opens the link, gets the prefilled sign-in page, signs in with the stubbed code, lands in the app as a member, shows as a device on the card, is removed). 25/25 checks. That driver depends on a local Chrome and is not part of this change.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for managing people in hosted workspaces, including invitations, roles, sign-in access, device management, and member removal.
  - Documented initial administrator setup and People settings, with verification details linked from the guide.

- **Tests**
  - Added end-to-end coverage for administrator setup, email-code sign-in, invitations, role changes, member removal, device revocation, domain-based invitations, session access, and authorization restrictions.


milind-soni:
Reviewed against current main. The required CI failure on all three platforms was the verification recipe citing people-invite.test.ts without its server/ prefix; fixed that reference. Also use the existing free-port-pair helper, bound fixture HTTP requests, drain child stdout, and fail promptly if the owned fixture exits during startup. The real-server invite/sign-in/role/revocation tests and neighboring tests now pass locally: 22 passed, plus typecheck and lint. No production behavior or live users changed. Enabling merge only after fresh required CI passes.

milind-soni:
Follow-up review addressed: documented this as an API fixture below the control-omb Drive map, not as an already-supported CLI/UI workflow. HTTP fixture calls now use an absolute AbortSignal deadline and handle response errors. Re-ran all 22 relevant tests, typecheck, lint, and the verification-docs checker successfully.

PR #1098, opened 2026-09-11
fix: accept successful Windows CLI termination

rahul-vanyar:
On Windows, `killCliTree` in `server/procs.ts` runs `taskkill /PID <pid> /T /F` and then waits for the child process's `close` event to resolve the returned promise. But when `taskkill` succeeds, the callback returns without resolving anything:

```ts
execFile("taskkill", ["/PID", String(pid), "/T", "/F"], { windowsHide: true }, (err) => {
  if (!err) return;
  ...
});
```

That leaves the promise waiting on the `close` event (or the timeout) even though the process tree is already gone. If `close` fires late — or doesn't fire in time for whatever reason — the timeout elapses and the caller sees a false "did not shut down" error, even though `taskkill` reported success.

The fix calls `done(true)` as soon as `taskkill` succeeds, instead of falling through to wait on `close`:

```ts
if (!err) return done(true);
```

Added a regression test (`server/kill-tree-windows.test.ts`) that mocks `execFile` to resolve successfully without ever emitting `close`, and asserts `killCliTree` still resolves `true` promptly. A second test confirms a genuine `taskkill` timeout still reports `false`.

https://claude.ai/code/session_01MDxACCHTR4dJELExVRvEMo

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows process termination to wait for confirmed process exit before reporting success.
  * Prevented premature completion and unnecessary delays when stopping command-line processes.
  * Improved handling of delayed, missing, or timed-out process exits.

* **Tests**
  * Expanded Windows termination coverage across platforms, including event-order and listener-cleanup scenarios.

* **Documentation**
  * Documented Windows process shutdown behavior and verification scenarios.


milind-soni (changes requested):
Holding this until the Windows lifecycle failures are resolved. The required Windows run 34595232595 fails both Antigravity closeAndWait and Claude logout disposal: the latter returns while the sign-in lock is still held. This change resolves killCliTree on taskkill success before the child close event, so callers that release runtime/account state on close can observe success before cleanup has finished. Please reconcile that lifecycle contract and add a Windows test proving immediate reuse/sign-out after disposal, then obtain a green required Windows run. I am not treating these failures as unrelated flakes.

milind-soni:
You read `server/procs.ts:202` correctly: the success branch of the `taskkill` callback does nothing, so only `close` or the timeout can settle that promise.

**The Windows job is red and this change caused it.** `typecheck + test (windows-latest)` in [run 34595232595](https://github.com/milind-soni/OpenMausBot/actions/runs/34595232595/job/103249337240) has two failures; main's Windows job is green at e7c300a7. Both are downstream of the changed line:

- `server/drivers/antigravity.test.ts:254` — "closeAndWait resolves only once the process is gone, so a rename on Windows cannot hit a running executable": `AssertionError: expected false to be true`. `closeAndWait()` now returns `true` while `exitCode` and `signalCode` are still null.
- `server/drivers/claude-login-auth.test.ts:217` — "disposal stops an in-progress logout before returning": `promise rejected "A Claude sign-in is running for th…" instead of resolving`. The home stays reserved until that child's `close` fires (`claude-login-auth.ts:240`), and `dispose()` now returns before it does.

**The premise has counter-evidence on main.** #841 fixed real EPERM reports with exactly this: "close() asks taskkill and returns before the process is gone, so the rename of the staging folder into place fails with EPERM while the exe still runs." A zero exit from `taskkill` means the kill was dispatched, not that the tree is gone. Your test asserts the case where `close` never fires; nothing here shows that it happens.

**It also reverses #1093**, merged an hour before this PR opened and present in your branch: confirm the process is gone before reporting success.

If you have a Windows case where `close` genuinely never arrives, post it and I will reopen the question — the fix would then confirm absence after the kill (poll the PID) so `true` still means reaped. Without that evidence, I would rather close this than weaken the guarantee.

milind-soni:
Holding this during triage. The Windows CI failure is directly relevant: server/drivers/antigravity.test.ts (closeAndWait resolves only once the process is gone) observes the child still running after closeAndWait returned true. Returning done(true) immediately on successful taskkill breaks the lifecycle guarantee callers rely on before renaming/replacing the executable. The fix needs to retain bounded process-exit/handle-closure confirmation, not equate taskkill success with the owned child being fully closed. Not merging this version.

milind-soni:
Updated the fix after reproducing the native Windows CI regression from the earlier version. Successful taskkill now waits until the child exit is observed; it does not wait for an inherited stdout pipe to close afterward. Both event orders are covered, as is a successful taskkill with no observed exit (still uncertain/false). All event-order tests now run on every host, and the required Windows job still exercises the real child-tree/Antigravity process checks. 62 local process/driver/isolated-server Stop tests passed; typecheck and lint passed. Auto-merge remains gated by fresh required checks, especially Windows.

PR #1101, opened 2026-09-11
fix: security hardening batch (CI permissions, secrets, dependency pinning)

rahul-vanyar:
A batch of small, independent security-hardening fixes found while going through the build, CI, and self-hosting paths. Each is scoped to one concern:

- **Pin and verify Android platform tools** — `scripts/prepare-android-tools.mjs` downloaded `platform-tools-latest-<os>.zip` from Google over an unpinned "latest" URL with no integrity check. Now it fetches an explicit, pinned release and verifies its SHA-256 before extracting.
- **Isolate npm publishing permissions** (`.github/workflows/npm-package.yml`) — the pack-and-prove job carried `contents: write` and `id-token: write` for its whole run, including the fork-triggered pull-request path. Publishing is split into its own job that only runs on release tags and only holds `id-token: write`; the GitHub Release update (which needs `contents: write`) is a separate job again, so no single job holds more permission than its step needs.
- **Close unauthenticated broker registration** (`cloudflare/composio-broker`) — the Cloudflare Worker's default config shipped with `REGISTRATION_MODE: "open"` and `workers_dev: true`, so a freshly deployed broker would accept new installation tokens from any caller before anyone configured it. Defaults now ship closed (`workers_dev: false`, `REGISTRATION_MODE: "closed"`), with a test asserting the registration endpoint never touches the database unless a caller explicitly opts into open mode.
- **Keep webhook secrets out of request URLs** — webhook delivery secrets were accepted (and shown in the UI/docs/curl snippets) as part of the URL path, which leaves them in proxy access logs, browser history, and referrer headers. Webhook credentials are now delivered as a bearer token; the old path-secret form is rejected unless an operator explicitly sets `OMB_WEBHOOK_LEGACY_PATH_SECRETS=1` for a migration window. The secret-redaction helper also gained a pattern for the old-style URL so any secret that leaked into logs before upgrading gets masked.
- **Pin production container inputs** — the Docker/Podman images and compose files referenced `node:24-bookworm-slim`, `caddy:2`, and `ghcr.io/milind-soni/openmausbot:latest` by floating tag, and the CLI engines (`@anthropic-ai/claude-code`, `@openai/codex`, etc.) installed whatever the registry currently resolved to `latest`. All of these are now pinned to a specific version plus digest (or exact npm version), and the Debian package sources are pinned to a snapshot date so a base-image rebuild doesn't silently pull newer, unreviewed packages.
- **Override vulnerable build dependencies** — `pnpm audit` now runs in CI, and two flagged transitive dependencies (`fast-uri`, `@xmldom/xmldom`) are pinned to patched versions via `pnpm.overrides` until their parent packages pick up the fix upstream.
- **Harden server error and static responses** — unhandled server errors echoed the raw error message (and sometimes internal file paths or fixture-specific detail) straight to the HTTP caller; they now return a generic message plus a correlation id, with the real error logged server-side. The static file server resolved paths without checking the result stayed inside the configured static root, so a crafted request path could potentially read files elsewhere on disk; it now validates the resolved path is contained within the static directory before serving. Static and SPA responses also get `X-Content-Type-Options: nosniff` and a restrictive `Content-Security-Policy`.

Each commit is independent and was tested against `main` individually (typecheck + the relevant vitest/node:test suites).

https://claude.ai/code/session_01MDxACCHTR4dJELExVRvEMo

## Summary by CodeRabbit

- **New Features**
  - Webhooks now use endpoint URLs with Bearer-token authentication.
  - Local Docker deployments can override the container image with `OMB_IMAGE`.
  - Android tools and browser components now verify downloaded archives.

- **Security**
  - Static content includes stronger security headers and path-traversal protection.
  - Server errors return sanitized messages with tracking IDs.
  - Hosted broker registration is closed by default.

- **Bug Fixes**
  - Deployment images, runtimes, and CLI tools now use fixed versions and verified sources.

- **Documentation**
  - Self-hosting and webhook migration guidance has been updated, including temporary legacy authentication support.


milind-soni:
The `npm-package.yml` split is the right call and cleanly done: top level drops to `contents: read`, `publish-npm` holds only `id-token: write`, `update-github-release` only `contents: write`, both gated on `refs/tags/v`, actions SHA-pinned. Main still carries the wide top-level `contents: write` + `id-token: write`, so nothing has superseded it.

What blocks the batch:

**1. The `pnpm audit` step reds ubuntu — `.github/workflows/ci.yml:39-43.`** Run `34597686731` fails there in 29s on 7 advisories, and none of them are the two you pin. They are `apps__docs>next` (2 critical), `wrangler>miniflare>sharp`, `electron-builder>app-builder-lib>js-yaml`, `vitest`, `@vitest/mocker`. So the job is red on arrival, and any new advisory published against an unchanged tree turns main red. Drop the step; if you want audit coverage, make it a separate non-gating job at `--audit-level high --prod`.

**2. `server/index.ts:14693` breaks a test it doesn't update.** Every 500 now returns `"unexpected server error"`, but `server/vps-routing.test.ts:300` still asserts `/fixture capture failed/`. That is the macOS failure.

**3. `scripts/prepare-android-tools.test.mjs` uses `node:test`, and vitest collects `*.test.mjs`** → "No test suite found in file" on macOS and Windows. This repo's node:test files are named `*.node-test.mjs`. Rename it or rewrite it against vitest.

I have not attributed the other failures (`independent-threads-api:390`, `post-to-room:755`) to you — that suite flakes on main too.

**Ask:** open `5a80c6d8` as its own PR and I will take it. Drop `f5c98957`. Rebase the rest — it conflicts with main now that #1099 moved `pnpm-workspace.yaml` — and push with 2 and 3 fixed.

rahul-vanyar:
Rebased onto current `main` (past #1099's `pnpm-workspace.yaml` move) and addressed the review:
- Dropped `f5c98957` entirely, so the gating `pnpm audit` step is gone.
- `server/vps-routing.test.ts` now asserts the generic `unexpected server error` 500 body (issue 2).
- Renamed `scripts/prepare-android-tools.test.mjs` → `.node-test.mjs` so vitest stops collecting a `node:test` file (issue 3).
- Fixed the `credential.url` → `credential.endpointUrl` + `Authorization: Bearer` leftover in `independent-threads-api.test.ts` and `webhook-restart.e2e.test.ts` (CodeRabbit).
- Split `5a80c6d8` out to #1158 as requested.

typecheck clean; the `independent-threads-api` delegation-approval flake is the pre-existing one you already excluded from scope (verified: that test file is byte-identical to upstream `main`, and it fails in isolation there too).

PR #1112, opened 2026-09-11
fix(ui): mascot right-sidebar settings + one-line composer

maxkongerskov:
## Summary

Restore two pieces of familiar chrome people already use — **not** bundled with Keep chatting.

1. **Mascot → right sidebar settings**  
   Bot settings open in the **right sidebar** again (accordion categories), matching the original layout. Re-click a category (name or chevron) to collapse it. On mascot open, **every category starts closed**; only intentional deep-links (e.g. the usage chip) expand a row. No centered popup.

2. **One-line composer**  
   Drop the extra tray under the input (working-folder path + Tools wrench). That duplicated Access / mascot settings and made the chat input two lines. Back to a single-line card.

## Demo

Mascot right-sidebar settings + one-line composer

## Test plan

- [ ] Mascot click opens **right sidebar** settings (not a centered modal)
- [ ] All categories start **collapsed**; click opens; re-click collapses
- [ ] Usage chip still deep-links into Usage expanded
- [ ] Composer is a **single-line** input (no folder/Tools tray)
- [ ] Access / MCP still reachable via mascot → Access

## Summary by CodeRabbit

- **New Features**
  - Bot settings now open in a right-hand sidebar with expandable accordion sections.
  - Settings opened to a specific section automatically expand that section.
  - Usage can be opened directly from the chat header cost shortcut without closing settings.
  - Memory drafts are preserved when navigating between sections.

- **Changes**
  - Removed the composer’s Tools tray.
  - Access settings are now reached through the bot profile/settings control.
  - Settings search clears when a section is opened.
  - Settings are now non-modal, allowing keyboard focus to return to chat.


milind-soni:
The accordion mechanics are right where they're easy to get wrong: re-clicking the open row collapses it (`BotSettingsDialog.tsx:362-366`), and Memory stays mounted while collapsed so an in-progress draft survives a section switch.

CI is green. The one red mark is Vercel asking for deploy authorization on our side — unrelated to your changes.

**1. The direction is a call I need to make, not something for you to fix.** This reverses two decisions from #956 (merged Sep 10): bot settings as a centered dialog, and the ComposerTray. It also changes where the mascot lands — today a bare open shows Overview; here every row starts closed, so the mascot opens onto a list and nothing else. I need to settle that before reviewing this as a change we keep. Please pause the UI work until I come back on it. The two below are worth fixing either way.

**2. `docs/keep-chatting/05-mascot-sidebar-one-line-composer.gif` is 22.7 MB** — more than twice the largest file in the repo (`docs/screenshots/composer-one-line.gif`, 9.8 MB), and no doc references it, so merging puts it in the tree permanently for nothing. Remove it from the branch and drag the GIF into the PR description instead; GitHub hosts it and the demo link keeps working.

**3. `collapsed` never re-syncs.** `BotSettingsDialog.tsx:43` reads `state.botSettingsExpandAccordion` once in a `useState` initializer, and `App.tsx:265` keys the component on `bot.id`, so it stays mounted the whole time the panel is open. Open the mascot, then click the usage chip (`ChatView.tsx:1444`): the store sets the section and the expand flag, but `collapsed` is still `true`, and line 352's `!collapsed && section === id` keeps Usage shut. Same for the identity item (`Sidebar.tsx:760`) and Access from the computer panel (`ComputerPanel.tsx:1024`). An effect on `[section, botSettingsExpandAccordion]` calling `setCollapsed(!expand)` covers all three.

Asking for: a push dropping the GIF and fixing the sync, then hold while I decide the #956 question.

milind-soni:
Holding for the mounted-panel deep-link issue: collapsed is initialized from botSettingsExpandAccordion but never synchronized after an external section dispatch. I reproduced this in the actual isolated renderer at 6af3e24c: select a bot with usage, open its profile (all rows collapsed), then click the header usage chip. The panel stays unchanged and Usage remains collapsed. Please add a renderer regression for that sequence, including repeating the same deep link after manually collapsing it. Also, the demo GIF adds ~22.8 MB to git history; an attached PR asset would keep this UI change lighter. Current CI is green, but it does not cover that interaction.

milind-soni:
Addressed the reproduced sidebar regression in this PR and updated against main. Accordion expansion now uses the existing store state rather than a mount-only local copy. Header shortcuts reopen the requested section after collapse, clear a conflicting search, and scroll that section into view. Collapsed Memory does not perform an active read, while its mounted draft survives. The real isolated renderer test passed repeatedly, covering live usage-header clicks, repeated collapse/reopen, search, role creation, MCP configuration, and failed preset recovery; a retained screenshot was visually checked. 91 store tests, typecheck and lint also passed. Fresh CI must pass before merge.

milind-soni:
Windows CI completed with one failure in the unchanged Claude steering fixture (server/drivers/claude.test.ts:1531), not the settings UI: slow mode finishes after 800ms, so the child can settle before the test resumes from its tool-event wait. Mac/Linux passed, and the same unchanged driver test passed the other selected Windows PRs. Replaced that wall-clock window with the existing finish gate plus an explicit child-received marker, preserving assertions for one turn, steered reply text, and rejection of a late steer. No production Claude changes. Also merged current main (including #1128 and #1098) into this branch. The Claude suite passes 111 tests (one skip), typecheck/lint pass. Before this fixture-only follow-up, the combined tree passed 6,168 tests, 174 desktop tests, packaged-server smoke and the forced real-app browser recipe. Required CI must pass again; auto-merge stays gated.

PR #1108, opened 2026-09-11
fix(threads): fold bot-closed threads out of the sidebar

AdityaUmale:
Fixes [MOCA-92](https://linear.app/supamaus/issue/MOCA-92/threads-a-bot-closes-stay-in-the-sidebar-and-helper-threads-pile-up): threads a bot closes stay in the sidebar, and helper threads pile up.

## What was wrong

`close_thread` (PR #1021) wrote a "Closed by @Bot" activity chip into the transcript and cleared unread, and that was all. The task record carried no closed state, so a closed thread reached every client looking exactly like an open one. A PM bot that opened several helper threads and closed them left every row behind, and because bot-opened threads are newest they also spent the sidebar's six recent slots, pushing the person's own threads out of the default list.

## What changed

- **Task record** gains `closedBy` (`{ botId, name, at }`), stamped by `close_thread` and persisted like `openedBy`. It is not a task PATCH field, so the HTTP task PATCH cannot forge or clear it.
- **Reopen**: the moment a new turn starts in a closed thread (`startTurn`), the stamp is cleared, so a thread the person or the opener picks back up is simply open again.
- **Sidebar**: `visibleSidebarThreads` folds closed threads out of the default list without counting them against the six recent rows. They still appear under "Show all N threads" and in search, dimmed, with "closed by <bot>" under the title and "Closed" in the tooltip. A closed thread that goes busy, unread, waiting, or is the selected one stays on screen. The All-threads picker shows the same byline.
- **Bots**: `list_threads` reports state `closed`; closing an already-closed thread returns `{ closed: true, alreadyClosed: true }` with no second chip. Tool descriptions updated.
- **Team backups** carry `closedBy` with the closer remapped like the opener, so a tidied pile does not come back as a pile after import.
- Nothing is deleted. Deletion stays the person's confirmed action from the row menu.

## Reproduce and verify

`node scripts/repro-closed-threads.mjs` launches an isolated fake-engine server on a throwaway data dir, opens three threads on a bot and two on a teammate through the same internal endpoints the agents tools hit, closes all five, and prints whether the stamp holds plus a URL for the real sidebar. It never touches the user's app data.

Manual, in the desktop app: ask a bot "Use start_thread to open a thread on yourself titled Scratch test with the message 'Reply with the single word done.' Wait for it with list_threads, then close it with close_thread." Before: the row stays under the bot. After: it leaves the default list when the bot reports the close, and reappears if you send a message in it.

## Mobile (second commit)

[...]

PR #1117, opened 2026-09-11
fix(threads): hide bot-closed threads from sidebar

### LINKED ISSUE #1074: Bots can close thread but that doesn't remove from the sidebar
### Before filing

### Area

Chat UI (src/)

### What happened

Image

### Version and install

.71 MacOS 

### Logs

```text

```

tahodev:
## Summary

Persist the close_thread result on the task and hide closed tasks from both recent and all/search sidebar lists without deleting their transcript.

Closes #1074

## Validation

- `pnpm exec vitest run src/components/SidebarThreadRow.test.ts server/thread-aware-bots.e2e.test.ts` - 18 passed
- `pnpm typecheck` - passed

## Notes

The task record remains available to server-side history and deletion paths. Only sidebar visibility changes.

## Summary by CodeRabbit

* **New Features**
  * Bot-closed threads are now marked as closed while retaining their transcripts.
  * Closed threads are hidden from sidebar history and do not occupy recent-thread slots.

* **Bug Fixes**
  * Closing a thread consistently updates its status and clears its unread indicator.

* **Tests**
  * Added coverage confirming closed-thread persistence, status updates, and correct sidebar ordering.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -8936,7 +8936,7 @@ const handleRequest = async (req: IncomingMessage, res: ServerResponse) => {
           from: { botId: from.id, name: from.name, color: from.color },
           tool: { name: `Closed by @${from.name}`, ok: true },
         });
-        if (task.unread) store.patchTask(owner.id, threadId, { unread: false });
+        store.patchTask(owner.id, threadId, { unread: false, closed: true });
         return json(res, 200, { closed: true, threadId, title: task.title, botName: owner.name });
       }
       if (method === "GET" && path === "/api/internal/rooms") {
--- a/server/store.ts
+++ b/server/store.ts
@@ -317,6 +317,8 @@ export interface TaskRecord {
   autoApprove?: boolean;
   alwaysAllow?: string[];
   unread?: boolean;
+  /** A bot marked this finished; the transcript remains until the person deletes it. */
+  closed?: boolean;
   rewound?: boolean;
   pinnedMessageId?: string;
   /** Runtime-only state, reset on load and never written to bots.json. */
--- a/server/store.ts
+++ b/server/store.ts
@@ -343,7 +343,7 @@ export interface TaskRecord {
 
 const TASK_PATCH_FIELDS = [
   "title", "projectId", "modelSelection", "approvalMode", "autoApprove", "alwaysAllow",
-  "unread", "rewound", "pinnedMessageId", "resumeCursors", "lastInstanceId", "cwd",
+  "unread", "closed", "rewound", "pinnedMessageId", "resumeCursors", "lastInstanceId", "cwd",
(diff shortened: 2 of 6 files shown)


milind-soni (changes requested):
Holding in favor of the fuller lifecycle approach in #1108. This patch sets closed=true permanently but has no reopen path. If a user sends another message to that task (or it becomes busy/unread/waiting), visibleSidebarThreads still filters it out, including search and Show all. Please preserve discovery and attention states and clear the closed marker when a new turn starts; add a close -> resume -> sidebar-visible regression. #1108 already attempts those behaviors and should be evaluated instead of merging both competing representations.

milind-soni:
The cleanup path is the part long-press implementations usually get wrong, and this one gets it right: window-level move/up/cancel/blur/Escape plus the early `clearPress` dispose (Sidebar.tsx:1237) mean a missed `pointerup` cannot glue the float to the cursor.

Two things before it lands.

**Long press does not survive a touch drag.** The bot row carries no `touch-action` — `rowClass` at Sidebar.tsx:1087 sets only `select-none [-webkit-touch-callout:none]` — and `touch-none` reaches the list at Sidebar.tsx:2132 only once `draggingBot` is set, i.e. after the 450ms lift. Chromium fixes touch-action when the touch begins and ignores later changes, so the first finger move after the lift is claimed as a pan, `pointercancel` fires, and Sidebar.tsx:1329 routes it to `finishDrag(false)`. The float dies on the input the feature is named for. Blunt fix: `touch-action: none` in `rowClass` under `reorder?.enabled`, at the cost of flick-scrolling over rows. Better: a non-passive `touchmove` listener added on lift that calls `preventDefault()` — scrolling survives, because a press that moves early is already cancelled by `botLongPressShouldCancel`.

**Bot order is dropped from workspace backup.** `SIDEBAR_BOT_ORDER_KEY` (src/lib/sidebar-preferences.ts:8) is not in `WORKSPACE_BACKUP_CLIENT_KEYS` (shared/workspace-backup-client.ts:3-15), which #1072 added to main yesterday, though its three sidebar siblings are listed there. Backup skips bot order and a restore comes back without it. One line.

CI is green — the red Vercel check is a deploy-authorization failure on our side, not this branch.

Also worth knowing: #931 reorders sidebar rows and touches the same two helpers, but it conflicts with main and this does not, so I would land yours. Nothing for you to do there.

Push those two fixes and I will re-review.

milind-soni:
Superseded by #1108, which is now merged. That implementation folds quiet bot-closed threads without losing search/history, reopens them on a new turn, and covers desktop, iOS, Android and backups. Closing this duplicate rather than layering a second closed-state mechanism on top.

PR #1119, opened 2026-09-11
fix(hermes): give the ACP model catalog probe a 15s deadline

### LINKED ISSUE #947: Hermes model catalog never populates: fetchHermesAcpModels probe killed at 5s while 'hermes acp' session/new takes ~6-7s
## Summary

The Hermes instance's model picker only ever shows the "Hermes default (config)" fallback entry. The full catalog hermes advertises (all `<provider>:<model>` ids) never appears — not after restart, not after refresh.

## Root cause

`server/drivers/acp/hermes.js` → `fetchHermesAcpModels()` hard-caps the probe at 5 seconds:

```js
timer = setTimeout(() => done([]), 5_000);
```

On a Hermes install with several authenticated providers (~143 models), `hermes acp` answers `initialize` in ~1.1s but `session/new` — whose result carries `models.availableModels` — completes at **~6.3–7.2s**, cold or warm. The probe is killed before the catalog arrives, so `resolveModels` gets `[]` for the remote catalog every time, and only `hermesConfiguredModel` (the config.yaml default) survives in the picker.

## Repro (same JSON-RPC handshake the driver uses)

Spawn `hermes acp`, send `initialize` (protocolVersion 1), then `session/new` {cwd, mcpServers: []} and time the responses:

- run 1: initialize @ 1.20s, session/new @ 7.16s → 145 availableModels
- run 2 (warm): initialize @ 1.09s, session/new @ 6.46s → 143 availableModels

Both exceed the 5s cap, so the picker permanently shows "No local models found" + the config fallback.

## Suggested fix

Raise the probe cap (e.g. 15s), make it configurable, or keep the 5s for `initialize` only and allow a longer window for `session/new`. Even a retry-once would mostly fix it.

Related: #715 (runtime handshake caps in core — same family, different constants).

## Secondary observation

Hermes is registered `access: "custom"`, so every model it serves lands in the picker's "Local models" pane ("Run this agent with a model already on your machine") even when the models are cloud BYOK providers behind hermes (z.ai, DeepSeek, OpenRouter, …). A "BYOK / custom" pane label might describe those better than "Local".

tahodev:
Fixes #947

## What changed

`server/drivers/acp/hermes.ts` → `fetchHermesAcpModels()` hard-capped the whole initialize + `session/new` exchange at 5 seconds:

```js
timer = setTimeout(() => done([]), 5_000);
```

On a Hermes install with several authenticated providers, `initialize` answers in ~1.1s but `session/new` — whose result carries `models.availableModels` — takes ~6.3–7.2s (cold or warm), so the probe was killed before the catalog arrived and the model picker only ever showed the "Hermes default (config)" fallback entry.

- The probe deadline is now 15s by default (`HERMES_ACP_MODELS_DEFAULT_TIMEOUT_MS`).
- It can be overridden with `HERMES_ACP_MODELS_TIMEOUT_MS`; missing, non-numeric, or non-positive values fall back to the default.
- Failure stays non-fatal and returns `[]`, exactly as before.

## Tests

`fetchHermesAcpModels` is exported for a probe-deadline contract test in `server/drivers/acp/hermes.test.ts`, using a minimal fake ACP CLI that answers `initialize` at once and delays `session/new`:

- catalog returned when `session/new` answers inside the deadline
- `[]` returned when the probe outlives the deadline, without waiting for the child
- invalid timeout override rejected, falling back to the default

`pnpm exec vitest run server/drivers/acp/hermes.test.ts` → 40 passed; `pnpm typecheck` clean.

## Summary by CodeRabbit

- **New Features**
  - Added configurable timeout support for Hermes ACP model discovery through the `HERMES_ACP_MODELS_TIMEOUT_MS` environment setting.
  - Set the default model discovery timeout to 15 seconds.
  - Invalid timeout values safely fall back to the default.
  - Improved command-line tool resolution across custom environments, including Windows command shims.
- **Bug Fixes**
  - Model discovery now handles slow or unavailable catalog commands more reliably through timeout enforcement.


milind-soni:
Maintainer triage: rebased by merging current main and pushed 81ed27f0. The Windows failures were in the new probe tests: raw spawn could not execute the Node-shebang fixture. The probe now uses the existing shell-free CLI resolver (also handles supported Windows wrappers), preserves SystemRoot in its isolated test environment, rejects fractional/overflow timeouts, and asserts the deadline lower bound and 15-second default. 45 Hermes tests passed locally; typecheck and lint passed. Waiting for the fresh Windows/Linux/macOS checks before merging.

milind-soni:
Addressed the follow-up review: the catalog resolver now uses the same supplied PATH/PATHEXT as the child process (without mutating the app environment or introducing a shell). Added a real spawned catalog test for a CLI found only on that PATH, including a Windows npm shim, plus a Windows resolver isolation test. Latest local run: 64 passed, 8 platform-specific skips; typecheck/lint/diff check passed. Fresh Windows CI remains required.

milind-soni:
The diagnosis is the valuable part here: measured timings for both halves of the handshake (`initialize` ~1.1s, `session/new` ~6-7s) instead of a guess that 5s was tight, and a deadline made injectable through `HERMES_ACP_MODELS_TIMEOUT_MS` so the behaviour is testable rather than a bumped constant.

Status, so nobody re-chases it. The Windows failure on your original 12ebcd93 was real, not a flake: `server/drivers/acp/hermes.test.ts` failed 2 of 40 in run 34646572555 with

```
AssertionError: expected [] to deeply equal [ { …(3) }, …(1) ]
```

on "returns the advertised catalog when session/new answers inside the deadline" and "ignores an invalid timeout override…" — a raw `spawn` cannot execute a Node-shebang fixture on Windows. That is fixed on the current head (ad627942), and windows/ubuntu/macos typecheck+test all pass in run 34693102058. The red Vercel check is team authorization and fails identically on every open PR here, so ignore it.

What is left is a maintainer call, not a push from you:

The PR now also edits `server/env-path.ts` and `server/procs.ts` — the CLI resolution every driver spawn goes through — to thread an optional `env` parameter. The title and description still cover only the timeout, so that change would merge undescribed. Worth splitting out or writing up.

One sharp edge if it stays: at `server/env-path.ts:244`, supplying `env` makes PATH discovery read `env.PATH ?? env.Path ?? ""` and skip `augmentedPath()`. Right for today's caller (`drivers/acp/core.ts:298` sets `PATH: augmentedPath()`), but a later caller passing an env without PATH would search nothing. A fallback or a line of comment would stop that being rediscovered.

@milind-soni — split or describe, then this is ready.

Issue #961, opened 2026-09-08
Can't Switch Between LLM Provider when "Full Access" Is Enabled


geoff-wp:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Chat UI (src/)

### What happened

If you attempt to switch between LLM providers when full access is enabled, you receive an error, so you have to change it to "Approve For Me", then change the LLM, then change it back to full access. 

<img width="1400" height="178" alt="Image" src="https://github.com/user-attachments/assets/de8bbc56-5554-447d-b562-d49e91bd34a5" />

### Version and install

0.1.63 MacOs

### Logs

```text

```

jamiojala:
Related scope/UX case confirmed in the official macOS **0.1.75** desktop build after upgrading from 0.1.74. PR #1103 adds the scoped confirmation, but it remains unavailable when the existing thread uses a different engine from the bot default.

Reproduction setup (generic names, no private data):

1. A bot's default model uses Codex and its profile approval mode is Full access.
2. An existing conversation has its own Grok model selection and Auto approval mode.
3. Open that conversation's approval menu. Full access is not offered; the footer directs the user to desktop bot settings.
4. Bot settings already show Full access, but for the Codex default. The Grok conversation remains Auto and continues surfacing native permission requests.

The current source explains this: `Composer.tsx` only shows `Use bot's Full access for this thread` when the thread and default engine driver kinds match. `projectBotForTask` correctly retains the explicit thread override, and HTTP thread settings correctly reject elevation.

The cross-provider grant must **not** silently transfer. The missing part is an actionable UI explanation and/or a fresh, explicit, conversation-scoped desktop consent flow for the actual selected provider. Ideally the user could grant Full access to that one existing conversation without changing the bot default, future conversations, or sibling threads. Preserve private-channel authorization, provider validation, busy checks, cancellation and interrupted-grant rollback; this is not a request to bypass those safeguards.

Observed evidence: the live profile reports `full`, the selected thread reports `auto`, and its composer displays `Approve for me for Grok`. No grant was edited or manufactured to test this case. I have not tested a same-provider Full-access regression here.

PR #1120, opened 2026-09-11
fix(approvals): let a provider switch ride with an atomic downgrade

tahodev:
Refs #961

## What changed

A bot running Full access (or Custom) cannot switch LLM providers: the server refuses with "Choose Ask first", so the person has to open Permissions, lower approval, switch providers, then raise it again. The refusal is deliberate and correct — an elevation granted for one engine's tool semantics must not silently transfer to another. This PR keeps that invariant and removes the errand:

- **`server/index.ts` (generic bot PATCH):** a single request that lowers `approvalMode` and changes `modelSelection` together is now accepted. `patchBot` applies both atomically, so the new provider never runs elevated — which is exactly what "choose Ask first" asks for. Cases that keep the refusal:
  - a model-only switch while elevated (no downgrade in the request)
  - a per-thread Full/Custom override, which a bot-level patch cannot lower
  - leaving Custom, which still requires the packaged desktop app's private channel (now also for the combined request)
- **`shared/approval-mode.ts`:** the guard's condition is now `providerSwitchBreaksElevation(mode, fromDriver, toDriver)`, shared between the server route and the model picker so the UI warns about exactly the requests the server would reject.
- **`src/components/ModelPicker.tsx`:** picking a different provider while the bot runs Full/Custom now opens a confirmation ("Switch provider and lower approval?") instead of dispatching into a server error; confirming sends the combined downgrade + switch. Thread-level switches keep the direct dispatch.

## Tests

- `server/approval-mode.test.ts` — `providerSwitchBreaksElevation` across elevated/non-elevated and same/other-engine cases
- `server/index.test.ts` (trusted-mode model guard fixture) — model-only switch still 400; combined downgrade + switch succeeds and lands Ask + the new provider; per-thread Full override keeps the 400; Custom combined request still 403 (packaged desktop app)
- `pnpm exec vitest run server/approval-mode.test.ts src/components/ModelPicker.test.ts` → 32 passed; the focused `server/index.test.ts` guard test passes; `pnpm typecheck` clean

## Summary by CodeRabbit

- **New Features**
  - Added a confirmation prompt when switching a bot’s model provider would reduce its approval level.
  - Confirming the switch automatically changes the approval level to **Ask** while applying the new model.
  - Model switches that preserve provider compatibility continue without an extra prompt.

- **Bug Fixes**
  - Improved handling of approval levels when changing model providers, preventing unsupported elevated modes from being retained.


milind-soni:
You stated the invariant correctly and the code keeps it: `providerSwitchBreaksElevation` in `shared/approval-mode.ts` gives the route and the picker one predicate instead of two drifting copies, and the four cases you list as still-refused are the right list.

What blocks it:

**1. The escape hatch can't be reached for threads the app creates.** `server/store.ts:2065` stamps `approvalMode: approvalModeFor(bot)` and `autoApprove` onto every thread `createTask` makes, and `server/index.ts:1602` pins it again for a thread-scoped grant. So for a bot at Full, any thread opened after its first already has `taskRow.approvalMode === "full"`. Your check at `server/index.ts:11840-11844` reads that, sets `projectedMode` to `"full"`, and returns the same 400 the combined request was meant to pass. `patchBot` (`server/store.ts:1662`) copies only resumeCursors/rewound/pinnedMessageId/unread onto the task, so the bot-level downgrade never reaches that row. The new test goes green because the `full-model-guard` fixture has no `tasks` array and the loader back-fills one with no `approvalMode` (`server/store.ts:985`) — a shape real bots don't have. Give a fixture a task row with `approvalMode: "full"` and the 200 turns into a 400.

**2. The dialog promises a switch the server rejects.** `src/components/ModelPicker.tsx:365` gates on `approvalModeFor(bot)`, the bot-level mode, while the server judges the active thread's row. Combined with (1): confirm, the level drops to Ask optimistically (`src/state/store.tsx:2402`), the 400 lands, the patch queue reconciles it back with an error banner. The two predicates also disagree the other way — bot at Full, thread row at Ask — where the dialog forces a downgrade the server never asked for.

**3. Rebase.** Conflicting in `src/components/ModelPicker.tsx` against 50362a94. CI has not run on this head (it needs a maintainer to approve the workflow for a fork PR); the one red check is the Vercel deploy authorization that every fork PR here shows, and it is not caused by your changes.

Before you push: I owe you a decision on (1). Either a bot-level downgrade also lowers the selected thread's row in the same write, or combined requests stay refused for any stamped thread — the record can't currently tell a stamp inherited from the bot apart from a deliberate per-thread grant, and that's mine to settle, not yours. Tell me which you'd rather implement and I'll confirm, then the fix and the rebase can land together.

milind-soni:
Further review against current main: the new confirmation is gated on !threadId, but both current ModelPicker entry points pass threadId (ChatView and bot-settings/ModelSection). This means users cannot reach the advertised confirmation. The current picker also has This bot / Only this thread scopes; the default-update task route has its own elevation guard. Please integrate the combined downgrade into those scoped routes and exercise both picker scopes, including a sibling thread with an explicit Full override and a Custom-mode transition. Keeping this out until that workflow is covered rather than merging unreachable UI.

PR #1121, opened 2026-09-11
fix(rooms): keep thread-only 1:1 picker off group turns

maxkongerskov:
## What

#1130 made model scope explicit: **This bot** updates the visible thread, groups, and future threads; **Only this thread** leaves groups alone.

The earlier version of this PR projected whichever private 1:1 tab was selected into room execution. That undoes the thread-only contract and makes room routing depend on which tab was last focused.

This revision locks the current contract instead:

- Group / Goal turns copy the **bot default** (`bot.modelSelection`), not the visible 1:1 task.
- A thread-only picker change stays on that private thread.
- `busyProviderSelections` already treats an in-flight room turn as using the profile default.

The original report (1:1 Grok worked, the room hit unsigned Claude `/login`) is what happens when the belt change is thread-only and the bot default is still Claude. After #1130, picking Grok with **This bot** is the way to move rooms; this PR does not restore the old coupling.

## Test plan

- [x] `pnpm exec vitest run server/member-turn.test.ts server/room-picker-model.e2e.test.ts`
- [x] `keeps a Group's profile model fixed while allowing independent idle task settings` (the 409 vs 200 case) passes
- Two fake engines, separate dump files: a thread-only picker dump is not written when the room speaks

## Notes

Standalone on current `main`. No change to #1130's This bot / Only this thread UI.


milind-soni:
Triage: #1130 is now merged with an explicit This bot / Only this thread selector. Bot scope updates the current thread, groups, and future threads; thread-only scope intentionally leaves groups unchanged. Having rooms automatically follow the selected 1:1 task would undo that contract and make room routing depend on which tab was last selected. Holding this overlapping PR; any remaining room-selection regression should be reproduced against current main and fixed without losing explicit scope.

milind-soni:
The e2e test is the strongest part of this: two fake engines writing separate dump files, so the assertion is which instance actually ran, not which one a mock was asked for. Threading `readySpeaker` through the post-wait `setupChanged` re-check is right too.

Three things before it can land.

**1. CI is red on all three runners and this PR is the cause.** `server/independent-threads-api.test.ts:302`, "keeps a Group's profile model fixed while allowing independent idle task settings" — `AssertionError: expected 409 to be 200`. It comes from the new guard at `server/index.ts:13169`. That test asserts the opposite rule to yours: editing the visible 1:1 task's model stays allowed while the bot speaks in a room, and only the bot default locks. Not a flake.

**2. Settle the premise before changing the routing.** On main the picker's scope already defaults to "This bot" (`src/components/ModelPicker.tsx:262`), which sends `updateBotDefault` (:369) into the default rooms run on; the hint at :507 reads "This thread, groups, and new threads." At "This thread" scope it deliberately does not reach rooms. Which scope was selected when Kiwi & co. hit unsigned Claude Code? If per-thread picks should follow a bot into rooms, that is a maintainer decision and it rewrites the test in (1).

**3. `busyProviderSelections` (`server/index.ts:13690-13696`) still pushes `bot.modelSelection` for an active room turn**, under a comment that says rooms run from the profile default — no longer true after this change. `PATCH /api/instances/:id` and `claude-update` gate on that list alone, so the account a live room turn is really on can be edited or removed underneath it.

Answer (2) in a comment first; I would rather you not rewrite code before we agree on the bug. Once the rule is settled, push (1) and (3) together.

milind-soni:
Holding this version during triage because current main now has explicit model scopes: This bot updates the visible thread plus the default used by groups/new threads; Only this thread deliberately leaves that default alone (see docs/verification/threads.md and bot-continuity.md). This PR instead projects whichever private thread is currently selected into room execution, including its approval mode. That would make a thread-only override affect rooms and change room behavior simply by navigating private threads. Please rework against the current scope contract rather than restoring that coupling.

maxkongerskov:
Reworked against the #1130 scope contract rather than restoring “whatever private tab is selected drives rooms.”

**Which scope the original bug used:** the report was a 1:1 belt change to Grok while the room still ran the bot default (unsigned Claude `/login`). That belt is the visible thread’s picker — after #1130, that path is **Only this thread** unless **This bot** is chosen. There was no evidence the profile default itself was Grok.

This PR now:
- runs group turns on `bot.modelSelection` (bot default / groups)
- leaves a thread-only 1:1 override on that private thread
- keeps idle task PATCH at 200 while a room is speaking (the 409 case in `independent-threads-api.test.ts` is green)

`busyProviderSelections` on current main already pushes `bot.modelSelection` for an active group turn.

PR #1128, opened 2026-09-12
Delegations: wait for a busy teammate for 24h instead of giving up, and say when one is waiting on you

aivsomkar:
## What changed

Two related changes to bot-to-bot handoffs, both server-side.

**1. A queued handoff waits for its target, then expires after 24 hours.**
Previously it gave up once the target bot had finished three turns without picking it up (`MAX_BUSY_ATTEMPTS`, `attempts`, receipt status `busy_gave_up`). Chatting with the target a few times was enough to silently kill a teammate's handoff. Now each item carries `queuedAt`, waits through any number of the target's turns, and posts one waiting chip per handoff rather than one per busy period. Its only bound is `DELEGATION_TTL_MS` (24h), producing an `expired` receipt that wakes the delegator through the existing `onSettled` path. Expiry is checked in `processOne` and by an hourly `expireStaleDelegations` sweep started after the boot drain.

The 24h rule also bounds fresh-thread handoffs (`targetThreadId`), which previously waited with no bound at all.

**Downtime does not consume the 24 hours.** Every bot loads `idle` at startup, so an expiry checked before asking "can the target take this now?" would make the boot drain expire leftovers instead of delivering them — quit the app on Friday, lose the queue on Monday. Expiry therefore applies only to a handoff that still cannot be delivered, decided by one `targetCanTakeTurn` helper shared with `holdWhileTargetBusy` so the two can never disagree.

**2. Peers can tell "working" from "waiting on the user".**
`peerStatus()` maps the harness's existing `BotActivity` to `available | working | waiting-on-user | not-responding | unavailable`, and feeds the team roster in system prompts, `list_bots`, the waiting chip, and `check_delegation` (which now also reports when the handoff expires). A bot parked on your approval card no longer reads as "working right now" to its teammates.

## Why

A handoff that dies because you happened to chat with the target three times is indistinguishable, from the delegating bot's side, from one that was never sent — and "busy" hid the one state a delegator could actually act on: the target is stuck waiting on a person. Design borrowed from October Bus's durable-inbox model; no code from it is used.

## How it was verified

- `pnpm vitest run` — 6017 passed, 44 skipped, 1 todo (487 files). Baseline on `main` was 6013.
- `pnpm typecheck` — exit 0.
- `pnpm lint` (`oxlint --deny-warnings`) — exit 0.
- New tests cover: waiting through five busy periods and still delivering; one chip per handoff across repeated drains; the "waiting on you" wording; expiry at the TTL boundary for both classic and fresh-thread handoffs; the boot-drain case (an over-age item whose target is idle is **delivered**, not expired); the sweep skipping a thread mid-drain; loading a queue saved before `queuedAt` existed; and a future `queuedAt` being clamped.
- Not yet done: the manual OMB2 side-by-side check (below).

## Screenshots (UI changes)

None — no UI changes. The user-visible surface is the wording of activity chips in chat.

## Platforms

|---|---|---|
| macOS | yes | in this PR; suite/typecheck/lint green, manual OMB2 check pending |
| Windows | yes | in this PR (shared server code, no platform gate); `Package Windows` workflow triggered on this branch |
| iOS | no | n/a: no phone code calls `/api/internal/*`; the changed payloads are read only by the local agents MCP proxy; waiting chips are existing `kind: "activity"` messages |
| Android | no | n/a: same as iOS |
| Companion | no | n/a: no phone-called route changed; the sidecar 404s `/api/internal/*` by design |

## Follow-ups

- [ ] Manual OMB2 check: (a) delegate to a busy bot → one waiting chip, delivered when it frees up; (b) open an approval card on the target → chip says it is waiting on you, `list_bots` agrees; (c) chat with the target 3+ times → still delivered; (d) quit with a handoff queued, move `queuedAt` back more than 24h, relaunch → delivered, not expired.
- [ ] **Decision needed:** a routine run stays `waiting` up to ~25h behind a queued handoff to a persistently busy peer, skipping that routine's interval fires meanwhile. Either give routine-thread handoffs a shorter bound or accept and document it. Not changed here.
- [ ] **Worth knowing:** `approvalAlreadyGranted` (a user's `ask_bot` "allow") can now be honoured up to 24h later instead of within a few of the target's turns. Reachability, section and allow-list are still re-checked at dispatch, so no authority is gained.

## Compatibility

- `delegations.json` gains one numeric field; old files load (backfilled `queuedAt`, stored `attempts` ignored) and an older build ignores the new field.
- Receipts on disk with `busy_gave_up` still load; nothing produces that status any more.
- `/api/internal/agents` keeps `busy` alongside the new `status`/`statusText`, and the proxy falls back to `busy` when they are absent.

## Checklist

- [x] `pnpm typecheck` passes locally; `pnpm vitest run` passes in full (6017). `pnpm test` additionally runs broker/electron/packaged-server suites, which were not run here.
- [x] No `dist-server/` edits
- [x] No platform-gated code added; no `shell: true` / cmd.exe string-building

## Summary by CodeRabbit

- **New Features**
  - Added clearer teammate status labels, including working, waiting for user input, not responding, and unavailable.
  - Enhanced delegation status messages with the responsible teammate and remaining expiration window.
  - Delegated tasks now remain queued through busy periods for up to 24 hours.
  - Added expiration receipts and notifications when queued handoffs expire.
  - Added compatibility for previously queued delegations after upgrades.

- **Documentation**
  - Added implementation and design documentation for delegation mailbox behavior.

- **Tests**
  - Expanded coverage for status reporting, queue expiration, legacy data, and restart scenarios.


milind-soni:
Maintainer review: updated against main and fixed a reproduced restart bug. Two overdue jobs for one target previously dispatched the first and immediately expired the second. Elapsed restored windows now renew before any boot dispatch; renewed deadlines are persisted and quick repeat restarts do not reset a still-valid window. Added a failing-before/passing-after backlog regression and an isolated real-server workflow: peer blocked on an approval, Chief delegates and becomes available, roster/chip report waiting on the user, approval releases one delivery and its attributed result returns automatically. 186 initial focused tests, 152 revised queue/proxy/roster tests, and 97 final delegation/concurrent-thread/comms tests passed; typecheck and lint passed. No real provider or user data was used. The verification guide now states the actual wall-clock/restart rules and the routine overlap tradeoff. Waiting for the required fresh platform checks before merge.

PR #1129, opened 2026-09-12
fix(access): Off means no browser, and connected apps are per bot

AdityaUmale:
Fixes [MOCA-94](https://linear.app/supamaus/issue/MOCA-94/bot-ignores-its-connected-composio-apps-until-a-hidden-toggle-is)

## Problem

Two settings tell you a bot can't do something, and the bot does it anyway.

**Works on: Off kept the browser.** `resolveSurface` withheld the computer for
Off but still mounted the built-in browser, so a bot set to Off browsed. Two
places already described Off as covering both surfaces: the bot overview drops
"Has the built-in browser" when the destination is Off, and the settings prompt
preview dropped the browser block on the same condition. The dispatch was the
odd one out, and the picker never said what Off covers.

**Connecting an app didn't hand it to a bot.** The connected-apps grant is per
bot, and a bot a Chief of Staff created, a package brought in, or a backup
restored starts with that grant off. Until it's on, the turn never mounts the
tools and never tells the model they exist, so the bot reaches for a browser
instead. The Apps panel said nothing about this.

Reported in #bugs: a user connected a calendar through Composio, set the bot to
Off, and watched it try the browser without knowing about the connection.

## Fix

Surface gating:

- `resolveSurface` returns no browser for Off, plus a prompt note that names the
  setting and tells the model to say so rather than narrate a browser it can't
  call.
- Room turns resolve their surface through `resolveSurface` too, so a room can't
  hand a bot the browser its own settings withhold. They carry the same note.
- The settings prompt preview uses that policy instead of spelling the browser
  rule out a second time. Duplicating the rule is what let the two drift apart.
- The bot-facing browser proxy refuses a bot whose destination is Off, in the
  gate and in the live re-check, so changing the destination mid-turn revokes
  the browser the way the per-bot switch already does.
- Bot settings say what Off covers, and the Browser switch stops offering a
  browser that Off withholds.

Connected apps:

- The Apps panel names the bots whose own grant is off, each with a button that
  turns it on. Bots whose engine can't mount the tools are left out, because
  their switch is disabled and naming them would move the dead end, not end it.

## How it was verified

- `pnpm typecheck`, `pnpm lint`, `pnpm i18n:check`, and the full `vitest` suite.
- New unit tests cover the Off branch of `resolveSurface`, the preview and
  dispatched prompt for an Off bot, and `botsMissingConnectedApps`.
- An isolated `control-omb` fixture, recorded under **Surface gating: Works on =
  Off** in `docs/verification/chat-turns.md`. A bot on Auto settled with no
  Works-on sentence in the system prompt the engine received. The same bot
  patched to `computer: "off"` settled with it. A channel send carried it in the
  room prompt, and an Auto bot mentioned in the same room did not.

Proving the browser server's own absence from `mcpConfig` needs a fixture with a
real browser engine installed. The dump shows what the model was told, not what
a native browser would do.

## Summary by CodeRabbit

* **New Features**
  * Added clearer “Works on: Off” guidance, explaining that bots operate without a screen or browser while connected apps, MCP servers, files, and chat remain available.
  * Added per-bot controls for enabling connected apps, with clearer availability states and action labels.
  * Account disconnect controls are now consistently available, including for remote clients.

* **Bug Fixes**
  * Prevented browser access when a bot’s computer setting is Off, including during active sessions.
  * Updated prompts and browser settings to accurately reflect available capabilities.

* **Documentation**
  * Added verification guidance and recorded fixture results for Off-mode behavior.


milind-soni:
Reviewed and refreshed against current main, preserving the explicit /setup behavior and remote connected-account fixes. Fixed readiness detection so the connected-apps grant appears only for a configured, fresh, successfully loaded catalog with an active connection; expired accounts and stale/error states do not count. Local typecheck, lint, localization checks, and focused policy tests pass. An isolated fake-provider fixture using the native browser engine verified browser MCP mounting in four cases: Auto/direct present, Off/direct absent, Off/room absent, and browser-only present. No live user app, browser profiles, or data were changed. Fresh required platform CI must pass before merging.

milind-soni:
The full suite caught a fixture-order issue: the new browser toggle coverage persisted browser:false before the pre-existing skill opt-out test, whose disk assertion incorrectly assumed no sibling flags had ever been saved. Fixed that assertion to require preservation of the actual prior sibling flags (without relaxing the skill opt-out checks). Re-ran the entire server/index.test.ts: all 206 passed, including exact learned-skill approval enforcement. Fresh CI is now the merge gate again.

PR #1134, opened 2026-09-12
feat(windows): frameless window with titleBarOverlay

v01tage25:
Removes the default Windows title bar and switches the desktop shell to a frameless window with a skin-matched overlay.

What changed:
- electron/window-chrome.mjs: win32 now returns { frame: false, titleBarStyle: 'hidden', titleBarOverlay: { color: '#070707', symbolColor: '#b5b5b5', height: 32 } }
- electron/window-chrome.test.mjs: updated Windows expectation

Why:
- The native white title bar breaks the dark header design.
- Overlay keeps native min/max/close buttons while the renderer draws its own header.
- Colors match the midnight skin (--color-app / --color-ink-secondary); desktop:skin can recolor it per skin.

Tested:
- pnpm dev:desktop on Windows, frame is gone, caption buttons visible on dark background.
- Before/after screenshots attached by author.

## Summary by CodeRabbit

* **New Features**
  * Windows app windows now use a hidden title bar with renderer-drawn minimize, maximize/restore, and close controls.
  * Windows headers, sidebars, and panels accommodate caption controls with improved spacing and draggable areas.
  * Window controls include localized labels and reflect maximize-state changes.
  * Desktop workspace settings can open from environment requests or links.
  * Added workspace menu/state access and computer-sharing management capabilities.

* **Bug Fixes**
  * Corrected Windows window chrome behavior and capability reporting.

* **Tests**
  * Updated coverage for Windows caption controls and capability reporting.

### DIFF
--- a/electron/capabilities.cjs
+++ b/electron/capabilities.cjs
@@ -170,7 +170,8 @@ function desktopCapabilities({
       // this computer's users
       homeDir: remote ? "" : homeDir,
     },
-    windowChrome: isMac ? "mac-inset" : "native",
+    windowChrome:
+      isMac ? "mac-inset" : hostPlatform === "win32" ? "win-overlay" : "native",
     screenPreview,
     dictation,
     localComputer,
--- a/electron/capabilities.cjs
+++ b/electron/capabilities.cjs
@@ -171,7 +171,7 @@ function desktopCapabilities({
       homeDir: remote ? "" : homeDir,
     },
     windowChrome:
-      isMac ? "mac-inset" : hostPlatform === "win32" ? "win-overlay" : "native",
+      isMac ? "mac-inset" : hostPlatform === "win32" ? "win-caption" : "native",
     screenPreview,
     dictation,
     localComputer,
--- a/electron/capabilities.test.mjs
+++ b/electron/capabilities.test.mjs
@@ -41,14 +41,14 @@ describe("desktop capabilities", () => {
     });
   });
 
-  it("reports the overlay window chrome on Windows", () => {
+  it("reports the renderer-caption window chrome on Windows", () => {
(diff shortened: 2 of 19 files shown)


milind-soni:
Maintainer note while landing this — I resolved the `src/locales/en.json` conflict (both sides appended keys; kept main's `canvas.*` set and your `window.*` labels) and fixed the CI failure, which was unrelated to the conflict: the new `useCaptionChrome` export broke four test files that mock `./DesktopCapabilities` with a bare literal, so every render of ChatView/GroupView threw `No "useCaptionChrome" export is defined on the mock`. Three now spread the real module via `importOriginal`; `BotIdentityAvatars.test.ts` has to stay a whole-module literal because it renders with no DOM and the real module reads `window.ogb` at import time.

Verified locally on the merged branch: 719 renderer tests, 190 electron node tests, `check:electron`, typecheck and lint all pass.

One limitation worth recording, not a blocker. On Windows the window is frameless for its whole life, but `windowControls` is deliberately outside `REMOTE_SAFE` in `preload.cjs`. So when the main window navigates to a hosted workspace, that page reports `win-caption` (capabilities derive it from the host platform, not from `remote`), keeps the drag regions and the 16px control shift, and correctly hides the buttons because `window.ogb?.windowControls` is absent. The window stays draggable, but there is no in-app minimize, maximize or close until the user switches back to Local — taskbar and Alt+F4 still work. Your `visible` gate is what keeps those from becoming dead buttons, so the degradation is graceful; it just leaves Windows plus a remote workspace without window controls.

Also flagging that the description still describes the `titleBarOverlay` approach with `#070707`, which the last commit replaced with `titleBarStyle: "hidden"` plus renderer-drawn buttons. I'll write the squash message from the code rather than the body.

PR #1140, opened 2026-09-12
fix(electron): add macOS Preferences menu role

### LINKED ISSUE #1012: macOS: Command+, does not open app Settings
### Before filing

- [x] Searched open and closed issues for settings, shortcut, keyboard, comma, Command, and Cmd. No matching report found. Related issues #845 and #932 concern closing bot Settings, not opening app Settings with a keyboard shortcut.

### Area

Desktop shell / packaging (electron/)

### What happened

The macOS desktop app does not open app Settings through the expected Command+, shortcut.

### Steps to reproduce

1. Open the OpenMausBot desktop app on macOS.
2. Make its main window active, with app Settings closed.
3. Press Command+, (Command and comma).

### Expected behavior

App Settings opens, following the standard macOS Settings shortcut.

### Actual behavior

Settings does not open through the shortcut.

### Shortcut clarification

The original report named Command+Q. This issue tracks Command+, as clarified by the reporter; Command+Q is the Quit shortcut.

### Version and install

macOS desktop app. App version, macOS version, keyboard layout, and install method were not supplied.

### Logs

No logs supplied. Based on a user report; the corrected shortcut has not been independently tested.

santhiprakash:
## Problem
On macOS, pressing Command+, does not open app Settings because the application menu built in `electron/menu.mjs` omits the `preferences` role.

## Triage
Electron's `preferences` role on macOS automatically registers the `Command+,` accelerator and renders the "Preferences…" menu item in the app menu. The macOS app menu currently has `about`, `hide`, `quit`, etc., but no `preferences`.

## Fix
- Insert `{ role: "preferences" }` into the macOS app-menu submenu in `electron/menu.mjs`.
- The `isMac` guard ensures the change only affects macOS; Windows/Linux continue to receive the same menu as before.
- Added a focused unit test in `electron/menu.test.mjs` that verifies the app menu contains a `preferences` item on `darwin` and that the app menu is not generated on `linux`/`win32`.

## Verification
- `pnpm typecheck` passes.
- `pnpm vitest run electron/menu.test.mjs` passes.
- Manual / smoke: the `preferences` role only appears in the `process.platform === "darwin"` branch.

## Notes
Closes #1012

## Summary by CodeRabbit

* **New Features**
  * Added a macOS **Preferences…** option with the `CmdOrCtrl+,` shortcut to open app Settings.
  * Added support for connecting to hosted workspaces from the desktop app.
  * Added computer-sharing controls, including folder selection, sharing status, and revocation.
  * Improved workspace management with validation, confirmation prompts, and clearer error reporting.

* **Bug Fixes**
  * Improved cleanup when forgetting workspaces or quitting the app.
  * Preserved unrelated workspace sessions when removing one workspace.

* **Tests**
  * Added coverage for settings access and platform-specific menu behavior.

### DIFF
--- a/electron/menu.mjs
+++ b/electron/menu.mjs
@@ -35,7 +35,7 @@ export function buildApplicationMenu({ environments, activeId, onSwitch, onAddFr
     ],
   };
   const template = [
-    ...(isMac ? [{ label: app.name, submenu: [{ role: "about" }, { type: "separator" }, { role: "hide" }, { role: "hideOthers" }, { role: "unhide" }, { type: "separator" }, { role: "quit" }] }] : []),
+    ...(isMac ? [{ label: app.name, submenu: [{ role: "about" }, { role: "preferences" }, { type: "separator" }, { role: "hide" }, { role: "hideOthers" }, { role: "unhide" }, { type: "separator" }, { role: "quit" }] }] : []),
     { role: "fileMenu" },
     { role: "editMenu" },
     server,
--- a/electron/menu.test.mjs
+++ b/electron/menu.test.mjs
@@ -1,0 +1,47 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("electron", () => ({
+  app: { name: "OpenMausBot" },
+  Menu: { buildFromTemplate: (template) => template },
+}));
+
+import { buildApplicationMenu } from "./menu.mjs";
+
+describe("buildApplicationMenu", () => {
+  const originalPlatform = process.platform;
+
+  function withPlatform(platform, fn) {
+    Object.defineProperty(process, "platform", { value: platform, configurable: true });
+    try {
+      return fn();
+    } finally {
+      Object.defineProperty(process, "platform", { value: originalPlatform, configurable: true });
+    }
+  }
+
+  const environments = [{ id: "x", name: "X", origin: "http://localhost" }];
+  const build = (platform) =>
+    withPlatform(platform, () =>
+      buildApplicationMenu({
+        environments,
+        activeId: "x",
+        onSwitch: vi.fn(),
+        onAddFromClipboard: vi.fn(),
+        onForget: vi.fn(),
+      }),
+    );
+
+  it("macOS app menu contains a Preferences item", () => {
+    const template = build("darwin");
+    expect(template[0].label).toBe("OpenMausBot");
+    expect(template[0].submenu.some((item) => item.role === "preferences")).toBe(true);
+  });
+
+  it.each(["linux", "win32"])("does not add an app menu on %s", (platform) => {
+    const template = build(platform);
+    expect(template[0].role).toBe("fileMenu");
+    for (const item of template) {
+      expect(item.label).not.toBe("OpenMausBot");
+    }
+  });
+});


santhiprakash:
Verified against Electron v43.4.0's `lib/browser/api/menu-item-roles.ts`: `preferences` is absent from `roleList`, so `getDefaultLabel` returns `''`, `getDefaultAccelerator` returns `undefined`, and `canExecuteRole` returns `false` (`execute()` no-ops). The item rendered blank and inert, and the previous test passed vacuously — it asserted the role string in our template, not Electron's resolution of it.

Fixed on `b56b65b4`:

- `electron/menu.mjs` — explicit `Preferences…` item (label + `CmdOrCtrl+,` + click) wired through a new `onOpenSettings` callback, replacing the unknown role.
- `electron/main.mjs` — `onOpenSettings` forwards to the main window over an `app:open-settings` channel.
- `electron/preload.cjs` + `src/types/ogb.d.ts` — `onOpenAppSettings` bridge listener; local shell only (not in the remote-safe subset, so a remote server's page never receives it).
- `src/App.tsx` — dispatches `toggleAppSettings` on the event, so Cmd+, now actually opens app Settings (#1012's expected behavior).
- `electron/menu.test.mjs` — replaced the vacuous role assertion with label/accelerator/click-wiring assertions. Flagging explicitly: the original test encoded the bug (it passed against the broken role-based item).

Verification: `npx vitest run electron/menu.test.mjs` → 3/3 (and the new test fails on the pre-fix code — sabotage run); `pnpm typecheck` clean; `oxlint --deny-warnings` clean on the changed files; `pnpm check:electron` clean (101 modules syntax-checked).

santhiprakash:
Valid gap — the menu test stopped at the `onOpenSettings` callback, so the `app:open-settings` hop and the preload subscription were untested. Added `electron/preload.node-test.mjs` (`1abb995`):

- Loads the real `electron/preload.cjs` in the existing `node --test` harness (`pnpm test:electron`), satisfying `require("electron")` from a cache-injected fake — no production code touched.
- Pins the literal `app:open-settings` channel on the receiving side, asserts every emit reaches the callback, and asserts the returned unsubscribe removes the listener.
- One boundary, stated plainly: the sender literal in `electron/main.mjs` (`webContents.send("app:open-settings")`, main.mjs:1400) isn't exercised because `main.mjs` can't be imported in a test harness; the receiving side now pins the same literal, so a rename on either side fails this test on the receiver first.

Verification:
- `pnpm test:electron` → 176 pass / 0 fail (174 pre-existing + 2 new).
- Sabotage: renaming the preload's channel to `app:open-settings-WRONG` fails exactly the new test; restored → 176/176 again.
- `electron/menu.test.mjs` still 3/3 (vitest); `oxlint --deny-warnings` clean on the new file.

The App-side dispatch leg (`App.tsx:184` → `{ type: "toggleAppSettings", open: true }`) remains covered by types + the existing action-shape contract test (`SidebarPhoneButton.test.ts`); mounting `App` would need a DOM test runner the repo doesn't carry, which didn't seem worth adding for this.

milind-soni:
Maintainer note while landing this. Two conflicts against current main, both resolved as a union rather than a choice:

- `electron/menu.mjs`: main added the `onConnect` callback for "Connect hosted workspace…"; this branch adds `onOpenSettings`. The signature now takes both.
- `electron/main.mjs`: main also wrapped Forget in `workspaceMenuAction`. Kept that wrapper and added the Preferences sender beside it.

Verified locally on the merged branch: typecheck, lint, `check:electron`, and 192 electron node tests all pass. The full vitest suite finished with only server end-to-end failures this branch cannot reach (HTTP 409 port conflicts from a second suite running beside it on the same machine); nothing here touches the server.

For the record on the earlier red CI: those three jobs were cancelled at exactly their `timeout-minutes` cap, not failed on an assertion. The suite runs with `fileParallelism: false` and healthy ubuntu runs land at 17 to 19 minutes against a 20-minute cap, so that was a slow runner rather than anything in this diff.

The `app:open-settings` channel is local-shell only, as your comments say, so Cmd+, is inert while the window shows a hosted workspace. That matches how the rest of the bridge is scoped and is the right call.

PR #1160, opened 2026-09-13
feat(windows): stage the CUA driver and enable local control on Windows

v01tage25:
## What this does

Windows builds had no CUA runtime at all, so **"This computer" could never work there**:
`resources/cua-driver.exe` and the native SDK were absent from the package, `startCua()`
was never called on `win32`, and `localComputerReady()` had no Windows branch — so the UI
always answered *"CUA Driver is not ready for local computer control"* no matter what the
user installed. This PR stages the runtime and completes the wiring.

### Packaging

- `scripts/prepare-cua-win.mjs` — checksummed (SHA-256) download of the win32
  `cua-driver` executable plus the Uniffi SDK, staged outside ASAR because native
  modules cannot load from an archive. Version-gated to the installed
  `@trycua/cua-driver`.
- `electron-builder.yml` — ships `cua-driver.exe` and `cua-sdk/` for win32.
- `package.json` — adds `build:cua:win` and wires it into `package:win`.
- `scripts/smoke-cua-win-embedded.mjs` — starts and stops a real embedded host, run
  as a `cua-win-smoke` leg on `windows-latest`.

### Wiring

- `electron/cua.mjs` — resolves the win32 driver, the standalone endpoint, and the
  embedded host. Windows has no TCC prompts to satisfy, so the embedding path skips
  the macOS permission gate.
- `electron/capabilities.cjs` — `localComputerReady` and `connectionEnabled` treat a
  live Windows connection as ready instead of always failing closed.
- `electron/main.mjs` — `startCua()` now runs on `win32` at boot, not only darwin/linux.
- `src/lib/local-computer.ts` — keeps "This computer" selectable on Windows before the
  driver is live (same behaviour macOS already has) and names the driver in the
  disabled-reason copy.

## The named-pipe detail

A Windows daemon endpoint is a named pipe, not a filesystem socket. `fs.existsSync`
returns `false` for `\.\pipe\cua-driver` even while it is accepting connections, so the
liveness probe now connects first on Windows rather than trusting a stat. This was the
difference between "driver installed and healthy" and "app reports nothing is there".

## Verification

On a real Windows 11 machine, with the driver installed:

- the named pipe connects (`fs.existsSync` false, `connect` true);
- the driver lists every tool the server requires — `click`, `get_window_state`,
  `list_apps`, `type_text` — and its descriptor passes `decodeLegacyDescriptor`;
- `capabilities.localComputer.available` is `true` with a live connection and stays
  `false` with none, so the failure mode remains fail-closed;
- the embedded host starts: `pnpm build:cua:win && node scripts/smoke-cua-win-embedded.mjs`
  → `smoke:cua-win OK — staged bundle drives a real embedded host`.

`pnpm typecheck` is clean. The new capability and UI tests pass (25 cases across
`electron/capabilities.test.mjs` and `src/lib/local-computer.test.ts`).

## Tests

- `electron/capabilities.test.mjs` — Windows reports ready with a live connection, and
  stays unavailable without one.
- `src/lib/local-computer.test.ts` — "This computer" stays selectable on Windows before
  the driver is live, and the disabled reason names the driver.

## Scope

This is Windows local control only. It shares no code with the macOS/Linux CUA paths
beyond the two capability predicates, and it does not touch the harness or any other
platform.

## Summary by CodeRabbit

- **New Features**
  - Added Windows support for local computer control using an embedded CUA driver.
  - Windows packaging now includes the required driver and native SDK resources.
  - Windows local computers remain selectable while the driver is starting, with clearer status messaging.
  - Restored computer sharing, hosted workspace navigation, window state synchronization, background updates, and dictation with clipboard support.
  - Added improved workspace validation, sharing lifecycle handling, logout warnings, and hosted-link navigation.
- **Documentation**
  - Added Windows local-control verification guidance.


milind-soni:
Maintainer cleanup pushed in 40f88cab:

- Merged current main and resolved conflicts while preserving hosted-workspace sharing, navigation guards, Android preview artifacts, and current CI timeouts. Removed the unrelated dictation/clipboard edits (including the missing import that crashed Linux packages).
- Added the missing Windows build/smoke scripts and wired staging into package:win.
- Windows now uses the owned EmbeddedCuaDriverHost in development and production. No detached standalone daemon or shared-pipe adoption; startup cancellation and late completion use the existing host cleanup path.
- Added behavioral Windows isolation tests and bounded the real Windows startup/shutdown smoke. Verified the pinned release checksum against the upstream GitHub asset digest.

Local verification: 91 tests passed, 1 platform skip; typecheck, lint, Electron syntax (106 modules), locale validation, and diff whitespace checks passed. The server proxy checks use inert children and disposable loopback fixtures, including the real Electron executable; no user app/data or computer input was touched.

Waiting for the revised Windows CI host smoke and cross-platform checks before merging. This verifies packaging/lifecycle, not arbitrary Windows input or elevated/UAC screens.

PR #1161, opened 2026-09-13
fix(ci): unblock main — refresh the stale Ukrainian strings and give the serial test job margin

milind-soni:
`main` is red on two counts right now, and every open PR inherits both. This fixes each.

## 1. `pnpm i18n:check` fails, 25 seconds into the ubuntu job

`Validate locale catalogs` is the first step of `typecheck + test (ubuntu-latest)`, so a stale catalog reddens every PR for a reason none of them caused:

```
uk.json: sidebar.section.channels: English source changed; refresh the translation or remove it for English fallback
… 23 keys total
```

The Ukrainian catalog (#1141) was translated against an older English source. Since then the product renamed **channels → group chats**, **contexts → teams**, and **Team Library → Templates**, so those 23 strings do not merely lag the English, they name things the app no longer has.

Retranslated the 23 and re-stamped their hashes with `node scripts/generate-locale.mjs uk --accept`:

| Key | English | Was | Now |
|---|---|---|---|
| `sidebar.section.channels` | Group chats | Групи | Групові чати |
| `sidebar.bot.moveToSection` | Move to team | Перемістити в розділ | Перемістити в команду |
| `sidebar.teamLibrary` | Templates | Команди | Шаблони |
| `composer.mention.channel` | Group chat | Група | Груповий чат |

`sidebar.section.removeFromContext` reads "Перемістити в General": the default team's name is the literal string `General` from `src/lib/team-map.ts`, untranslated in every language.

Nothing else in the catalog moves. The 90 keys Ukrainian has never carried still fall back to English, and no other locale is touched.

## 2. `Run tests` is cancelled at its 20-minute cap

With the locale step fixed, this PR's ubuntu and macOS jobs went on to be cancelled mid-`Run tests` at exactly 20m0s. That is not this diff: two JSON files cannot slow the suite.

`vite.config.ts` sets `fileParallelism: false` deliberately, so wall time is the sum of every test file and grows with each one added. The 20-minute cap was set when healthy macOS/Linux runs took about 15 minutes; they now reach 20. The cap exists to catch a hang, not to ration ordinary growth, so macOS/Linux go to 30 and Windows to 35.

**If the sum outgrows that too, the real fix is splitting the job** — server suites on one runner, renderer on another — rather than raising the ceiling again.

## Verification

- `node scripts/generate-locale.mjs --check` → `locale catalogs valid (10 languages, 2049 English strings)`
- `vitest run src/locales src/components/SettingsModal.appearance.test.ts` → 8 passed
- On the previous push, `Validate locale catalogs` passed and the job proceeded to `Run tests`, which is what proved the second problem.

A Ukrainian speaker should still read the renamed terms; this restores a valid catalog and correct terminology rather than dropping the strings to English fallback.

## Summary by CodeRabbit

- **Localization**
  - Updated Ukrainian translations across group chats, teams, templates, composer, rooms, people, and backup settings.
  - Added Ukrainian labels for moving chats between teams and creating team-associated group chats.
  - Clarified messaging about workspace member removal, including session termination and blocked sign-ins.
  - Refined terminology for consistency across the interface.


milind-soni:
Reviewed and updated against main ab66e55f. Resolved the overlap with #1164 by restoring the 23 corrected Ukrainian translations and their matching English-source hashes. No other locale changes. All catalogs validate (10 languages, 2,073 English strings), 19 focused locale/appearance tests pass, and lint passes. The CI change only raises the overall job limits to 30/35 minutes; all checks and per-test timeouts remain intact. Fresh cross-platform CI is running on 5868db1a before merge.

milind-soni:
Follow-up review found two existing CI issues and keeps the fixes test-only: POSIX 0600 checks now run on POSIX while all persistence assertions remain on Windows; the MCP editor renderer test waits for its asynchronously loaded Paste config control and mounted textbox. Verified 10 team-computer tests and the complete isolated renderer workflow (11.2s), plus focused lint and diff checks. Fresh required CI is being rerun; no checks are bypassed.

PR #1162, opened 2026-09-13
fix(security): close two confirmed holes in shared-folder access

milind-soni:
Two high-severity holes in the computer-sharing feature from #1145, both proved by executing the real code. A companion PR gates the whole feature off by default; these make it safe when someone turns it on.

## 1. The credential guard was a string compare on a case-insensitive filesystem

`executeSharedOperation` decided containment with `path.relative`, which is byte-wise. APFS and NTFS are not. Asking for `openmausbot/credentials.bin` instead of `OpenMausBot/credentials.bin`, or the same name in NFD rather than NFC, produced a relative path starting with `..`, so the guard read it as "outside, allow" and the OS then resolved it to the same file.

That was step one of a chain: read the credential blob, read the grant file, rewrite it with `terminal: true` and `computer: true`, and the next launch re-armed it after only a shape check. A reviewer got a working shell this way from a grant the user approved as read-only folders.

Containment is now decided by `{dev, ino}`: protected roots are resolved to filesystem identities, and the target's resolved ancestor chain is walked against them. A write may create a file that does not exist yet, so the walk falls back to the nearest existing ancestor. The check also runs against the opened descriptor's own identity, so no spelling the OS accepts can defeat it. The existing `O_NOFOLLOW`, hardlink and symlink defences are untouched.

Worth knowing: APFS is normalization-preserving, so `realpath` of an NFD spelling returns NFD. Comparing realpath text would still have been defeated; only identity closes it.

## 2. "Not the entire computer or home folder" accepted the parent of every home

The picker refused exactly two strings: the filesystem root and `os.homedir()`. `/Users`, `C:\Users`, `/home` and `/System/Volumes/Data/Users` were all accepted, with write enabled, which made the read-only tier alone enough to reach `~/.ssh` and cloud credentials.

It now refuses any folder whose identity is home's or an ancestor of home, seeded from both `os.homedir()` and the macOS data-volume path, and any volume root. On this host `/Users` and `/System/Volumes/Data/Users` are the same `dev:ino` with different realpath text, and `/home` is autofs on its own device, which is why both rules are identity-based.

## Also: the app's own data directory was unprotected

`protectedPaths` held one entry, Electron's `userData`. `~/.openmausbot`, which holds provider API keys and `sessions.json`, was not in it. `createComputerSharing` now takes the roots from `main.mjs` (so this module still does not import electron), and a root that does not exist yet on a fresh install is skipped rather than thrown, which previously would have aborted the connect loop.

## Exploit, before and after

```
before                                    after
LEAK   lowercase -> SECRET-BLOB           refused
LEAK   NFD -> GRANT                       refused
ACCEPTED /Users                           refused
ACCEPTED /home                            refused
ACCEPTED /System/Volumes/Data/Users       refused
```

## Verification

```
pnpm typecheck / pnpm lint / pnpm check:electron   clean
pnpm test:electron                                 196 pass, 0 fail
vitest shared-computers + e2e                      9 passed
```

Six new tests, each proved to bite by reverting the guard it covers: case-variant spelling, NFD spelling (including a write of a not-yet-existing file, which exercises the ancestor fallback), the picker refusing home and volume roots, a legitimate nested share still working read and write, the data directory refused through a broad share, and a missing data directory not breaking save or connect. Reverting guard one fails two tests; reverting guard two fails one; dropping the new protected root fails one; making the resolver throw on a missing root fails two.

## Deliberate trade-offs

- Any mount point is now refused, so the top level of an external drive can no longer be shared, only a folder inside it. That follows from refusing volume roots and is a real behaviour change.
- Children of home such as `~/Library` remain selectable by explicit pick; they are no longer reachable through a broad share. Refusing them would break sharing `~/Documents`.
- `{dev, ino}` is not stable on every network or FUSE mount. Identities are recomputed per operation so nothing stale persists, and any stat error other than a missing path fails closed. A text fallback was not added, since text comparison is the bypass being fixed.
- TOCTOU is narrowed, not eliminated: a rename of an ancestor between the walk and the first read remains theoretically live. Closing it needs `openat`-style traversal, which Node does not expose.

## Summary by CodeRabbit

- **Security Enhancements**
  - Computer sharing prevents access to home directories, filesystem roots, volume roots, and the application’s data directory.
  - Shared access recognizes alternate path spellings, symbolic links, and files that do not yet exist.
  - Applications can designate additional protected folders that cannot be shared.
  - File access is rechecked before reading or writing to prevent indirect access to protected locations.
  - Protection now handles large filesystem identities reliably and fails closed when protected locations cannot be verified.

- **Localization**
  - Added Ukrainian translations for channels, team organization, accessibility labels, and backup guidance.


milind-soni:
Reviewed and updated on 1596aa5c with the reviewed #1161 prerequisite. Additional regressions now prove that an unreadable home/ancestor fails closed and that 64-bit filesystem identities are compared exactly without Number rounding. Missing protected roots are revalidated when created. Verification: 199 Electron node tests, 9 isolated shared-server tests, typecheck, lint and Electron syntax checks pass. No live user files or app were used. Remaining ancestor-rename/unsupported-filesystem limitations are not claimed solved; #1165 keeps the unfinished sharing surface off by default. Waiting for required platform CI before merge.

PR #1165, opened 2026-09-13
fix(security): gate computer sharing behind a flag that defaults off

milind-soni:
Ship-blocker. The computer-sharing feature from #1145 has seven confirmed high-severity holes, including one where a reviewer obtained a working shell from a grant the user approved as read-only folders. This makes the whole surface unreachable so a release can be cut today. No code is removed.

## The flag

`features.sharedComputers`, defaulting **off**. Deliberately not a Settings toggle: it is a maintainer escape hatch for an unfinished feature, enabled by hand in `~/.openmausbot/config.json`. The accessor says so.

## What is gated

Fifteen places, not just the obvious ones:

- The `/api/shared-computers/*` family, `/api/desktop/shared-computer-control`, and the bot-facing `/api/internal/shared-computers`.
- Both agent tools, `list_shared_computers` and `shared_computer`, omitted from the tool list entirely (the same way skill tools already are) with the handlers refusing anyway.
- The public descriptor no longer advertises the capability, so a client is never told the server speaks a protocol it will refuse.
- The Electron connector never starts, the "Share this computer?" dialog is never offered, and the four `sharing:*` IPC handlers refuse.
- The Settings panel and its entry point render nothing.

## Indistinguishable from a build that never had it

A disabled route answers exactly like an unknown one: `404 {"error":"no route: POST /api/shared-computers/connect"}`, byte-identical to `POST /api/not-a-feature/connect`. The internal route gives the same `unknown internal endpoint` as any other. The descriptor omits the key rather than sending `false`.

One subtlety worth the extra line in `request-auth.ts`: the route family sits in the client-scope allow-list, so without gating that entry a paired client session would get 404 here and 403 on genuinely unlisted paths — a distinguisher. The rule is now inert while the flag is off, so both give 403.

## How Electron learns it

From the local server's `/api/config`, the path `gatherDiagnostics()` already uses; read before the connector could start and re-read on each window load and each `sharing:*` call. An unreachable or older server means off. This is a start-time check rather than a renderer signal because `start()` runs in `app.whenReady()` before a window exists. There is a second, independent gate: the remote workspace's own descriptor, which `identity()` already refuses when the capability is absent.

## Verification

```
pnpm typecheck / pnpm lint / pnpm check:electron    clean
vitest config + gate + agents-proxy + remote-sessions   184 passed
vitest server/index.test.ts                             209 passed
```
Mutation checks both ways: flipping the config accessor to default on fails four tests, including the one asserting a disabled route answers like an unknown route; flipping the agent-tool env gate fails two, including the one asserting the tools are absent from the list. Both restored green.

The full suite has 24 failures in 6 files, all of which reproduce identically on a pristine checkout of the same base — no new failures. The locale check also fails on that base, fixed separately in #1164.

## Not gated, deliberately

`forget(env)` still runs: it only deletes local grant records, so it removes authority rather than granting it, and gating it would strand records.

## Summary by CodeRabbit

* **New Features**
  * Computer sharing is now available only when explicitly enabled by the server.
  * Sharing controls, capabilities, tools, and APIs remain hidden or unavailable when disabled and return standard unavailable responses.
  * Enabling the feature restores sharing controls, capabilities, and tools.

* **Bug Fixes**
  * Improved protection against sharing restricted folders, including application data and filesystem root locations.
  * Sharing now stops safely if availability is disabled during an active session.

* **Localization**
  * Added Ukrainian translations for channels, team features, backup imports, and related interface labels.


milind-soni:
Reviewed and updated to 4dc7ba53, including the reviewed #1161 and #1162 prerequisites. The gate now also withdraws running authority: local flag checks abort the connector and its jobs; server flag withdrawal closes queued/running rendezvous work and native leases, with checks repeated after asynchronous body reads. Regression tests prove the command PID is alive before withdrawal and exited afterward. Combined validation: 220 focused tests, 200 native Electron tests, typecheck, lint, and the real isolated desktop workspace connection smoke pass. Hosted workspace connections remain available; computer sharing defaults off and explicit re-enablement requires restarting the desktop. Waiting for required platform CI before merge.

PR #1179, opened 2026-09-13
fix(coordination): keep a minimum runway when the handoff lifetime ceiling hits

### LINKED ISSUE #1178: Room handoff lifetime ceiling aborts in-flight work and silently starves follow-up requests
## Environment

- OpenMausBot at commit `ef317c5` on `main`, local server started with `pnpm dev:server`, macOS.

## Summary

`server/room-handoffs.ts` bounds each handoff tree with a 30-minute lifetime (`ROOM_HANDOFF_LIMITS.lifetimeMs`) measured from the **root** node's `createdAt`, and `tick()` applies that single ceiling to every non-terminal node in the tree:

```ts
const error = this.hooks.validate(n, parent);
if (error || this.now() - this.root(n).createdAt > ROOM_HANDOFF_LIMITS.lifetimeMs) {
  this.cancelTree(n, error ?? "Room request timed out", "failed");
}
```

Three consequences:

1. **Actively running work is aborted mid-run.** The ceiling is computed from the root's creation for every node, including a node that transitioned to `running` seconds earlier. When the root crosses 30 minutes, `tick()` cancels the node and aborts its provider process via `cancelTree`.
2. **Follow-up requests under an aged root are silently doomed.** The enqueue-time guard only refuses when the root is *already* past the ceiling:

   ```ts
   if (count >= ROOM_HANDOFF_LIMITS.requests || this.now() - root.createdAt > ROOM_HANDOFF_LIMITS.lifetimeMs) throw new Error("Room handoff budget exhausted");
   ```

   A follow-up request enqueued under a resumed root (e.g. at root age 23 minutes) passes the guard and is accepted into a queue where at most 7 minutes of runway remain. It is then killed by the ceiling minutes after it starts.
3. **The fallback message is misleading.** When the only problem is the lifetime ceiling, `validate()` returns nothing and the result is the generic `"Room request timed out"`, which reads like a routing or delivery failure. A sender cannot distinguish "never delivered" from "delivered, ran, killed mid-flight", which makes triage go the wrong way.

## Minimal reproduction (fake clock)

Save as `server/room-handoffs.lifetime-repro.test.ts` and run `pnpm vitest run server/room-handoffs.lifetime-repro.test.ts`. Both tests pass against current `main`, demonstrating the behavior:

```ts
import { mkdtempSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";
import { describe, expect, it, vi } from "vitest";
import { RoomHandoffs, type RoomHandoffHooks } from "./room-handoffs.ts";

const addr = (id: string) => ({ groupId: id, threadId: `${id}-thread`, botId: `${id}-bot` });
const flush = () => new Promise<void>(resolve => setImmediate(resolve));

describe("room handoff lifetime (repro)", () => {
  it("aborts a running child when the root hits the 30-minute ceiling", async () => {
    let nowMs = 0;
    const hooks: RoomHandoffHooks = {
      validate: () => undefined, busy: () => false, report: vi.fn(), changed: () => {},
      run: (_n, _r, signal) => new Promise(resolve => signal.addEventListener("abort",
        () => resolve({ ok: false, text: "aborted mid-run" }))),
    };
    const engine = new RoomHandoffs(join(mkdtempSync(join(tmpdir(), "repro-")), "requests.json"), hooks, () => nowMs);
    const { node } = engine.enqueue(addr("A"), "turn", undefined, addr("B"), "work", "long build");
    engine.sourceSettled("turn", true);
    nowMs = 25 * 60_000; engine.tick(); await flush();
    expect(node.status).toBe("running");            // started 5 minutes before the ceiling
    nowMs = 30 * 60_000 + 1; engine.tick(); await flush();
    expect(node.status).toBe("failed");             // aborted mid-run at the root's ceiling
    expect(node.result).toContain("Room request timed out");
  });

  it("accepts a follow-up under an aged root and kills it minutes later", () => {
    let nowMs = 0;
    const hooks: RoomHandoffHooks = { validate: () => undefined, busy: () => false,
      run: vi.fn(async () => ({ ok: true, text: "done" })), report: vi.fn(), changed: () => {} };
    const engine = new RoomHandoffs(join(mkdtempSync(join(tmpdir(), "repro-")), "requests.json"), hooks, () => nowMs);
    engine.enqueue(addr("A"), "turn", undefined, addr("B"), "first", "first task");
    engine.sourceSettled("turn", true);
    nowMs = 23 * 60_000;                            // guard passes: root is under 30m
    const { node } = engine.enqueue(addr("A"), "turn", undefined, addr("C"), "followup", "follow-up task");
    expect(node.status).toBe("queued");             // accepted with only ~7 minutes of runway
    nowMs = 30 * 60_000 + 1; engine.tick();
    expect(node.status).toBe("failed");
    expect(node.result).toContain("Room request timed out");
  });
});
```

## Expected behavior

- A request that is actively running when the root reaches its ceiling should not be aborted seconds or minutes after it started; it should keep a minimum execution runway (or the budget should be measured per request).
- A follow-up enqueued when the root's remaining lifetime cannot honestly serve that runway should be refused at enqueue time with an error that says why.
- The terminal message should state what actually happened: the lifetime budget, the node's status (queued or running), and elapsed time — not a generic timeout that looks like a delivery failure.

## Actual behavior

Running turns are aborted mid-flight at `root.createdAt + 30m`, follow-ups enqueued in the root's final minutes are accepted and then killed with `"Room request timed out"`, and there is no test coverage for `lifetimeMs`, the timeout message, or the budget-exhausted guard in `server/room-handoffs.test.ts`.

## Impact

Any coordination tree whose work legitimately spans more than 30 minutes total (build, install, and verify cycles routinely do) loses in-flight work with a misleading error. Follow-up requests from resumed turns are structurally unreliable near the end of a root's life, and the abort can leave partially written artifacts, prompting a full retry that duplicates effort.

## Possible directions

1. Give each execution a minimum runway: compute each node's deadline as `max(root ceiling, node start + minimum runway)`, and refuse *new* enqueues when the remaining root lifetime is below that minimum (accurate error at enqueue time).
2. Move the budget to per-request: measure lifetime from each child's own `createdAt`, keeping the existing `requests`/`depth`/`executions` caps as the tree bound.
3. Either way, replace the `"Room request timed out"` fallback with an accurate message naming the ceiling, the node status, and elapsed time.

I'm happy to send a PR implementing option 1 with regression tests if that matches the direction you want.

bradhallett:
## What changed

- `server/room-handoffs.ts`: a node's lifetime deadline is now the later of the tree ceiling (root `createdAt` + `lifetimeMs`, unchanged) and the node's own execution start plus a new 10-minute `minRunwayMs` minimum. `tick()` validates and expires deepest-first so each node is failed with its own status, and an ancestor past its ceiling waits while a descendant is still running inside its runway instead of cascading a cancellation into it. `enqueue()` refuses new work when the root's remaining lifetime cannot serve the minimum runway. The generic `"Room request timed out"` fallback is replaced by a message naming the budget, the node status, and elapsed time (and the enqueue refusal states how much runway remains).
- `server/room-handoffs.test.ts`: four regression tests using an injected clock — queued expiry at the ceiling, running work keeping its runway past the ceiling and then failing with its status, enqueue refusal when the remaining runway is too small, and a follow-up under an aged root still being served when a full runway remains.

## Why

Fixes #1178. The 30-minute lifetime was measured from the root's `createdAt` for every node on every tick, so work that started moments before the ceiling was aborted mid-run, and a follow-up enqueued in the root's final minutes passed the enqueue guard only to be killed minutes later. Both surfaced as the generic `"Room request timed out"`, which reads like a delivery failure rather than a budget expiry. The fake-clock repro in the issue passes against current `main` and this branch implements the issue's option 1 (per-node minimum runway plus an honest enqueue-time refusal).

## How it was verified

- `pnpm vitest run server/room-handoffs.test.ts` — 20 passed (16 existing + 4 new), on macOS at commit 61e52d6.
- `pnpm typecheck` — pass.
- `pnpm lint` — pass.
- Full `pnpm test` — the vitest stage did not finish inside my local window; mid-run it reported 19 failures in `server/drivers/agents-proxy.test.ts` and 1 in `server/index.test.ts`. I re-ran agents-proxy in isolation with the two changed files reverted to `ef317c5`: the identical 19 tests fail on unmodified `main` too, so those are pre-existing local-environment failures, not caused by this change. The single `index.test.ts` failure was not re-verified against `main`. Because vitest exits nonzero, the broker/electron/packaged-server stages did not run locally; CI covers them.

## Screenshots (UI changes)

None — server-only change.

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — typecheck, lint, and the full `room-handoffs` suite pass; full `pnpm test` vitest stage shows only pre-existing failures verified against `main` (see above), later stages deferred to CI

Fixes #1178

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved room handoff lifetime handling to prevent work from being queued or continued when insufficient execution time remains.
  - Preserved active work and required follow-up executions while enforcing per-task expiration limits.
  - Improved processing order for deeply nested handoffs, allowing eligible waiting work to resume before expiration.
  - Added safeguards for execution limits and minimum runtime availability to provide more predictable handoff behavior.


bradhallett:
Thanks for the review — both findings are addressed in 31c6cc1 (added on top of 61e52d6, no force-push).

**1. Waiting parent failed at the ceiling before it can resume (Major, `server/room-handoffs.ts` 211–227)** — verified valid, fixed. In `tick()` the expiry loop ran before the report/resume loop, so once a child completed past the ceiling, its waiting parent had no active descendant (`protectsRunner` false) and was failed before the later loop could flip it to `resume`; the follow-up execution that decides on the children's results never ran. Expiry now also skips a node that still owes that follow-up execution — `resume`, or `waiting` with every child terminal, recursively through descendants (`owesFollowUp`). The parent survives past the ceiling, its children's reports settle it into `resume`, and dispatch sets `startedAt`, so its deadline is the later of the tree ceiling and the follow-up execution start plus the 10-minute minimum runway — the same runway rule as any other execution.

Regression test `"resumes a waiting parent past the ceiling and gives the follow-up its own runway"`: a child finishes at 31m, inside its own runway but past the 30m ceiling; the parent must resume and keep running until its follow-up start + 10m (41m). At 61e52d6 the new test fails (1 failed | 20 passed, parent observed `failed`); with 31c6cc1 the suite is 21/21.

**2. Docstring coverage 0% on touched functions** — added short doc comments to the touched helpers that lacked them: `duration` and `lifetimeError` in `server/room-handoffs.ts`, the `fixture` helper in `server/room-handoffs.test.ts`, and the new `owesFollowUp`.

Verification: `pnpm vitest run server/room-handoffs.test.ts` — 21/21 pass; `pnpm lint` (oxlint --deny-warnings) — pass; `pnpm typecheck` — pass.

PR #1182, opened 2026-09-14
fix(electron): save attachment downloads without dialog

### LINKED ISSUE #1181: Attachment <a download> leaves Electron download pending and disables native input when the save dialog does not map on GNOME Xorg
OpenMausBot 0.1.78 attachment “Save a copy” controls can make the BrowserWindow stop accepting all native X11 mouse and keyboard input on GNOME Shell 46 Xorg, HiDPI DPR 2. The renderer, local server, browser thread, and CDP input remain live.

Static trace: `useLocalFileSave` fetches the attachment into a Blob, creates a hidden `<a download>`, and calls `link.click()`. The Electron main process does not install a `session.defaultSession` `will-download` handler and does not call `DownloadItem.setSavePath` for this flow. There is an explicit parent-owned `desktop:save-file` IPC handler elsewhere, but attachments do not invoke it.

Dynamic trace on Electron 43.4.0 / Chromium 150.0.7871.224: `Browser.downloadWillBegin` fires for the blob and all 2,388,828 bytes arrive within tens of milliseconds. Under default behavior, the item stays `inProgress`, a full valid PNG appears as `~/Downloads/.org.chromium.Chromium.*`, no save dialog maps, and the focused BrowserWindow no longer accepts native XTEST input. CDP `Input.dispatchMouseEvent` still works.

Controlled boundary: CDP download behavior `deny` produces `canceled` and native input survives. `allow` with an explicit destination produces `completed` and native input survives. The default unresolved save-path state alone triggers the freeze. `--gtk-version=3` produces exactly the same failure as the default launcher; GTK3 is not a workaround. Host GTK packages are GTK 3.24.41 and GTK 4.14.5.

An independent re-measurement reproduced both sides: default behavior received all bytes but stayed `inProgress` and stranded native input; `allow` with an explicit path completed and native input survived. A CDP-set download behavior is not an operator workaround because it is dropped when the client that set it disconnects.

Proposed app-level fix: own `will-download` and synchronously assign a safe, collision-resistant Downloads path with `item.setSavePath`, or route the attachment bytes through the app's existing validated IPC save flow. Do not rely on Electron's default download save dialog. Add a Linux integration test asserting that an attachment download reaches `completed` and that BrowserWindow native input remains responsive when no dialog surface is available.

dfmcintosh-hash:
## Bug

Attachment `<a download>` links can remain pending on GNOME/Xorg when Electron's save dialog does not map.
The unresolved download leaves the BrowserWindow unable to receive native input.

## Fix

Own `will-download` in the main process and synchronously assign a collision-free path in Downloads.
Repeated filenames use Chromium-style ` (1)`, ` (2)` suffixes without replacing existing files.

## Tests

- `pnpm test:electron` — 203 passed, 4 skipped
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:electron` — 106 Electron modules syntax-checked
- `pnpm broker:test` — 8 passed
- `pnpm test:packaged-server` — passed
- `pnpm test` — 6,636 passed; 7 unrelated server npm-fixture tests failed because their temporary PATH could not resolve the fixture's `/usr/bin/env node` shebang on this host

Fixes #1181

## Summary by CodeRabbit

* **New Features**
  * Downloads are now saved in the Downloads folder using a collision-free filename.
  * Existing files are preserved by adding a numbered suffix to new downloads.
  * Downloaded filenames are secured within the designated Downloads folder.


dfmcintosh-hash:
The [root-cause clarification](https://github.com/milind-soni/OpenMausBot/issues/1181#issuecomment-5672127635) shows why the merged handler is still worth having as a deterministic guard when a save dialog cannot map in headless, stale-portal, or remote-session environments. If you would prefer dialog-first behavior with a bounded timeout or cancellation path instead, please say which behavior you want.

PR #1183, opened 2026-09-14
Peer approvals: cap how many can wait on you at once

aivsomkar:
## What changed

When a bot wants to message or delegate to a teammate and its "Ask me before contacting other bots" setting is on, `requestPeerApproval` raises a card and waits up to 15 minutes. Nothing limited how many of those could pile up, and each one also holds a bot's turn open.

Now the number waiting on you is bounded: **3 per bot, 25 across the workspace**. Over the cap, the request is refused before any card exists, with a new `too_many` verdict distinct from a deny you actually gave.

- `requestPeerApproval` returns `PeerApprovalVerdict = "allow" | "deny" | "too_many"`.
- The check sits after the existing standing-grant fast paths and before any card is created, so a `too_many` refusal raises no card, sends no notification, starts no peer turn, and neither sets nor consults a standing grant.
- A standing "always allow" grant still allows without a card and does not consume budget — it never asks you anything.
- Callers (`delegations.ts` and three sites in `index.ts`) report the refusal in their own words rather than claiming the user denied it.
- Budget frees itself as cards are answered, expire, or are cancelled.

## Why

A looping or unattended fleet could bury the owner in approval cards, and every unanswered card pinned a bot's turn. The governing rule is that a peer request must never gain authority — so when we cannot ask, we refuse. **It fails closed**, and there is a test asserting exactly that (the over-cap verdict is never `allow`).

Design borrowed from October Bus's escalation caps (100/agent, 1000/scope); no code from it is used.

## How it was verified

- `pnpm vitest run server/peer-approval.test.ts server/delegations.test.ts server/agent-tool-policy.test.ts server/direct-coordination.e2e.test.ts` → 127 passing
- `server/comms.test.ts` (peer-comms e2e) → 27 passing
- `pnpm typecheck` and `pnpm lint` → clean
- New tests: at-cap returns `too_many` with no new card; the fail-closed assertion; answering one frees budget; the per-bot cap is per bot; the workspace cap trips across 13 bots each under their own cap; a standing grant still allows at/over cap; and a capped handoff's receipt does not claim the user denied it.
- A side-by-side OMB2 build of this work was provided to the owner for hands-on testing.

## Screenshots (UI changes)

None — no UI changes.

## Platforms

|---|---|---|
| macOS | yes | in this PR; suites, typecheck and lint green |
| Windows | yes | in this PR (shared server code, no platform gate); NOT built on Windows |
| iOS | no | n/a: server-only; no phone-called route or payload shape changed |
| Android | no | n/a: same as iOS |
| Companion | no | n/a: no phone-called route changed |

## Checklist

- [x] `pnpm typecheck` passes; the suites above pass (`pnpm test` also runs broker/electron/packaged-server suites, not run here)
- [x] No `dist-server/` edits
- [x] No platform-gated code added; no `shell: true` / cmd.exe string-building

## Summary by CodeRabbit

* **New Features**
  * Added safeguards limiting pending peer approval requests per bot and across the workspace.
  * Approval requests exceeding these limits are rejected without creating notifications or approval cards.
  * Added clear error messages when approval capacity is full.
  * Existing standing approvals continue to work without consuming approval capacity.

* **Bug Fixes**
  * Delegations exceeding approval limits now fail safely, record a denied result, and are not dispatched.
  * Completing an approval now frees capacity for subsequent requests.


milind-soni:
Triaged for merge — holding on one interaction with `coordinate_bots`.

**Per-bot cap of 3 vs. a 4-teammate coordination (high)**

`coordinate_bots` accepts 1–4 `bot_ids` and raises one approval per target in a single synchronous `targets.map(...)`, and `requestPeerApproval` registers each pending entry synchronously inside its Promise executor. With `MAX_PENDING_PEER_APPROVALS_PER_BOT = 3`, the 4th target sees `pendingForBot === 3` and returns `too_many` — even though nothing was pending when the call started.

So for any bot with "Ask me before contacting other bots" on (and not Full access), a four-teammate coordination now **always** fails, and it fails in the worst way: the user is asked to answer three cards first, and the work is then discarded.

The mechanism itself is good and fails closed — it's specifically the arithmetic of a single multi-target call against a cap of 3. A few ways out:

- count a multi-target `coordinate_bots` as **one** request against the cap (it is one user decision, really), or
- raise the per-bot cap to at least the max `bot_ids` length, or
- check capacity once for the whole batch up front and refuse the call as a unit, so the user isn't asked to answer cards for work that cannot proceed.

The workspace-wide cap of 25 and the distinct `too_many` verdict at all four call sites both look right to me; it's just this one boundary.

PR #1193, opened 2026-09-14
feat(engines): one-click installation without Node or npm (MOCA-139)

aivsomkar:
## What changed

- Add one-click installation for npm-based engines even when the server has no Node.js, npm, or npx. Download pinned Node 24.14.1 from the official distribution, verify checked-in SHA-256 digests, and keep runtime, engines, and npm cache private to OpenMausBot.
- Show Not installed, Setup required, Sign-in required, Ready, and Active states in the model picker. Onboarding immediately exposes an install action when no engine is ready.
- Show prerequisite preparation and installation progress, prevent duplicate clicks, support retry, and keep sign-in separate. Preserve each engine's installation state when switching panels.
- Add regression tests, a synthetic renderer fixture, and an opt-in isolated real-download acceptance script. No new dependencies, global installs, shell profile changes, or administrator prompt.

## Why

Closes [MOCA-139](https://linear.app/supamaus/issue/MOCA-139/add-model-picker-ui-showing-installactive-status-and-install-prompt). New users should not need to install a package manager or use a terminal before installing an engine.

## How it was verified

- Targeted regressions: 138 passed, 8 platform-specific skips across 11 files.
- Full `pnpm test` passed: main Vitest suite 545 files / 6,755 tests passed, 40 tests skipped and 1 todo; broker tests 8 passed; Electron tests and packaged-server smoke passed.
- Typecheck, lint, locale validation, production build, and packaged-server smoke passed.
- macOS arm64 real-download smoke with an empty prerequisite PATH: private Node bootstrap, Codex installation, update/reuse, startup PATH registration, and CLI execution passed (`codex-cli 0.154.0`). The disposable installation was removed; no provider login or model call occurred.
- Isolated browser fixture: visible onboarding action, preparation/install states, failure/retry, transition to Connect ChatGPT, and switching from installing Codex to missing Qwen passed. The picker fits a 390px viewport.
- Native Windows/Linux download-and-install acceptance remains outstanding. Supported archive targets are not a claim of native acceptance. Unsupported/musl targets retain manual setup.

See [verification instructions](https://github.com/milind-soni/OpenMausBot/blob/feat/moca-139-one-click-engines/docs/verification/one-click-engines.md).

## Screenshots (UI changes)

Clean-machine onboarding after this change (synthetic fixture; no user data):

One-click Codex installation

A baseline screenshot was not retained; previously the npm prerequisite prevented the server install option on a machine without npm.

## Checklist

- [x] No `dist-server/` edits
- [x] Platform-specific code is gated; no shell command string-building

## Summary by CodeRabbit

* **New Features**
  * Engines can automatically prepare required tools and install supported packages without system Node.js or npm.
  * Added clearer setup statuses, including preparing, installing, setup required, sign-in required, and ready.
  * Installation progress now updates automatically, with status refresh and retry options.
  * Model picker now shows engine names, providers, active selection, and setup status.
  * Onboarding can suggest and expand the most relevant engine setup.

* **Documentation**
  * Added guidance for clean-machine setup, supported platforms, troubleshooting, and verification.


milind-soni:
Triaged for merge — holding on two things, one of which is a Windows hazard this repo has already been bitten by and already has a helper for.

**1. `server/node-runtime.ts`: two bare `rename`s right after executing `node.exe` (high)**

`usable(paths(extracted))` runs the freshly extracted `node.exe` to check it works, and the very next statements rename that directory:

```ts
if (!(await usable(paths(extracted)))) throw new Error(...);
const destination = join(parent, release.directory);
if (await usable(runtime)) return runtime;
if (existsSync(destination)) await rename(destination, join(staging, "previous-incomplete-runtime"));
await rename(extracted, destination);
```

On Windows, process teardown is asynchronous and Defender commonly holds a just-executed image for a beat, so `rename` throws `EPERM`/`EACCES`/`EBUSY` transiently. The whole install then fails with the generic message, on the platform this PR exists to serve (no Node, no npm).

There's already an exported helper for precisely this, used for the same extract→rename shape in `server/drivers/antigravity-runtime.ts:463`:

```ts
await retryOnWindowsFileLock(() => rename(runtimeDirectory, destination));
```

Wrapping both renames the same way should be all it needs.

**2. `server/engine-install.test.ts`: install tests can reach the network (medium)**

The pre-existing tests are no longer isolated — depending on the host's Node version they can start a real `nodejs.org` download. That makes the suite dependent on network and on whatever Node the runner happens to have, which on CI is a flake generator and offline is a hard failure. Worth pinning the version check or stubbing the fetch so the tests can't leave the machine.

Everything else looked good — the download is pinned and sha256-verified, the archive size is bounded, the extractor is the OS one with no dev tools required, and the concurrent-completion check before the rename is a nice touch. Happy to push the `retryOnWindowsFileLock` change myself if you'd rather I did.

Issue #1197, opened 2026-09-14
Inbox-style thread management: ordering, archive, snooze, cross-bot attention inbox


bradhallett:
## Plan

Inbox-style thread management for the sidebar, split into four independently reviewable PRs:

1. **Inbox ordering (client-only)** — extract a shared pure orderedSidebarThreads helper from visibleSidebarThreads; attention-demanding threads (waiting-on-you, then working/busy, then queued, then unread) float to the top within a bot while idle threads keep stored order. Applied consistently in BotThreadList, SidebarBotActivity, and the TaskPicker activity listing.
2. **Reversible thread archive (server-persisted)** — archivedAt on TaskRecord (+ TASK_PATCH_FIELDS and the HTTP PATCH whitelist for /api/bots/:id/tasks/:threadId), Archive in the thread row context menu, an archived disclosure with unarchive at the list tail, badges in show-all/search/TaskPicker. Archived threads leave the default list unless they demand attention, mirroring the existing bot-level hidden precedent.
3. **Thread snooze (server-persisted)** — snoozedUntil with the same dual whitelist treatment; default until-new-activity (cleared at the central patch points when a thread becomes busy or unread); presets until activity / later today / tomorrow; expired snoozes self-heal on read against the server clock; a snoozed thread that goes waiting-on-you resurfaces.
4. **Cross-bot attention inbox (client-only)** — an Attention section at the top of the all-threads TaskPicker reusing the ordering helper, plus a clearly labeled sidebar header entry point. No changes to the This computer control (that is a capability grant, verified in review).

Constraints: one concern per PR, no new runtime dependencies, server changes carry colocated vitest tests (temp HOME), no DCO/CLA requirements. Each PR records its upstream base commit.

Track: ordered threads already stay visible (visibleSidebarThreads) but never reorder; archive/snooze have no thread-level analog today (closedBy is bot-initiated and the wrong field); the attention inbox generalizes SidebarBotActivity's collapsed escape hatch across bots.

PR #1201, opened 2026-09-14
feat(threads): reversible archive with attention resurface

bradhallett:
## What changed

Per-thread archive, following the bot-level `hidden` precedent end to end:

- `archivedAt?: number` (epoch ms) on `TaskRecord`, added to `TASK_PATCH_FIELDS` and to the HTTP PATCH allowlist for `/api/bots/:id/tasks/:threadId` (both whitelists, or the field silently no-ops). `null` unarchives; non-timestamps get a 400. The wire task mapping spreads the record, so the flag reaches every client through the existing bot broadcast.
- Archived threads leave the default sidebar list, with the same attention override as bot-closed threads: a working, waiting-on-you, queued, or unread archived thread resurfaces immediately.
- Thread-row context menu gains Archive / Unarchive (disabled while the thread is busy, like delete). The marker is the quiet byline under the title ("Archived"), which `threadByline` already feeds to the sidebar row and the All-threads picker, and the row dims like closed threads.
- A small "Archived (N)" disclosure at the list tail (chevron, default collapsed) lists folded-away archived threads for unarchiving. Show-all and search keep listing archived threads with the marker.

## Why

Part of the inbox-style thread management plan in #1197. Archiving is user intent, so it is server-persisted and syncs across devices instead of living in a renderer preference; the attention override keeps an archived thread from stranding an approval.

## How verified

- `pnpm vitest run src/components/SidebarThreadRow.test.ts server/thread-archive.test.ts` — 16 passed. New: archived threads fold out of the default list but resurface when demanding attention; byline/dim/tooltip behavior; and an isolated-HTTP-fixture API test covering archive round-trip, persistence to `bots.json`, `null` unarchive dropping the key, invalid value 400s, and the allowlist still rejecting unknown settings.
- `pnpm typecheck` and `pnpm lint` green.

## Screenshots

Not produced in this environment (no runnable app); available on request.

## Checklist

- One concern per PR; no new runtime dependencies; no lockfile or dist-server churn.
- Upstream base commit: 83678c9.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to archive and unarchive threads from the sidebar.
  - Archived threads appear in an expandable section and remain visible when active or requiring attention.
  - Archive status persists across sessions and displays updated labels, styling, and tooltips.
  - Thread controls and status updates now target the currently running engine, including routed and grouped threads.

- **Bug Fixes**
  - Invalid archive timestamps are rejected with a clear validation error.
  - Archived threads are handled consistently in filtering, search, and “show all” views.
  - Key verification is skipped for keyless Chatterbox.
  - Remote command length limits are enforced consistently.


bradhallett:
Addressed in 8ff4ac7 (added on top of 53eb334, no force-push): archive presence is now tested with `archivedAt !== undefined` in all five spots (byline, default-list folding, row state, menu toggle, archived disclosure filter), with a regression test for the zero timestamp. Local `vitest run src/components/SidebarThreadRow.test.ts` 14/14, typecheck and lint clean.

bradhallett:
Addressing the CodeRabbit note about threads that report `activity: "working"` without `busy`: the row could demand attention yet render no Working status and still allow archive/delete.

Fixed by unifying every busy-gated behavior on one predicate, `isWorking(task) = task.activity === "working" || task.busy`:
- `demandsAttention` and the Working status label (8ff2a15b)
- the spinner glyph, so an activity-only working thread shows the working indicator (1a16a89b)
- the Archive and Delete menu actions, including the delete-confirm guard, so working threads cannot be archived or deleted from the row (1a16a89b)

Verified with `pnpm vitest run src/components/SidebarThreadRow.test.ts server/thread-archive.test.ts` (17/17), `pnpm typecheck`, and `pnpm lint`, all green.

PR #1202, opened 2026-09-14
feat(board): a task board, with dated cards

v01tage25:
## What this adds

A task board: a durable, per-server board of work where each card points at a
chat run. Cards carry a bot, a brief, a day, an optional deadline, a status and
a run state; the board shows them as columns on a pannable canvas.

This is a self-contained feature. It adds no new provider calls of its own — a
card's Start reuses the existing thread/run routes.

## Why

The app already has chats, rooms, teams and a team map. What it had no place
for is *work that is not a conversation yet*: a thing you intend to do, that
belongs to a bot, that has not been started. A task board is that place, and
because every action goes through the routes that already own those paths, a
card can reference a run without becoming a second scheduler.

## How it works

**Server.** `server/work-items.ts` is a durable store of cards persisted as a
single JSON file, alongside the other stores. It owns creation, patching,
ordering, the routine link, and settling a card when its run ends.

Routes, all under `/api/task-board`:

| Method | Path | Purpose |
|---|---|---|
| `GET` | `/api/task-board` | the whole board, with each card's owner resolved |
| `GET` | `/api/task-board/teams` | sections with bot counts |
| `POST` | `/api/task-board/items` | create a card |
| `PATCH` | `/api/task-board/items/:id` | edit, move, reorder, set or clear dates |
| `DELETE` | `/api/task-board/items/:id` | remove a card |
| `POST` | `/api/task-board/items/:id/run` | start the card's turn |
| `POST` | `/api/task-board/items/:id/stop` | stop the card's own thread |

Cards reach a team through their owner's `section` — the same field the sidebar
groups by and the team map draws, so a board and a map cannot disagree about
who is in what. A card on a hidden bot, or with no owner, falls into the
unsectioned team rather than appearing in every team at once.

**Client.** The board is a canvas of columns you can move, resize and rename.
`src/lib/task-board.ts` holds the pure logic (columns, teams, day filtering,
run availability, status tone) and `src/lib/board-layout.ts` holds the layout
rules, so both are testable without a browser.

## Card dates

A card carries two dates that mean different things, deliberately not merged:

- **`day`** decides which day's board the card appears on. A card with no day
  belongs to the day it was made, which is what makes the day filter safe to
  switch on for a board that already has cards — nothing needed migrating.
- **`dueAt`** is a deadline. **Display only** — nothing starts, moves or fails
  because of it. That restraint is the whole reason a due date is safe to add:
  if it moved cards it would be a scheduler, which is explicitly not what this is.

The day row offers Yesterday, Today and Tomorrow, defaulting to Today, with a
calendar behind the Day label so any date stays reachable. A past day renders
read-only — drops refused, cards not draggable, run and edit controls gone,
with a banner saying why. The work on a finished day is a record, and editing
it would rewrite what happened.

On `PATCH`, `undefined` means "not mentioned, keep it" and `null` means "clear
it". This distinction is load-bearing: coercing a missing field to `null` would
clear both dates on every unrelated patch — which is exactly what dragging a
card between columns is.

Dates parse at **local** midnight. `new Date("2026-09-14")` is UTC and lands on
the previous day for half the world, which a day picker must not do.

## Notable details

- **The run clock counts only real work.** `startedAt` is set when a card is put
  in a working column, but moving a card there is arrangement, not work. The
  timer now requires the card's agent to actually be busy, so a card nobody
  started cannot display hours of elapsed time.
- **A wheel over a column's card list scrolls the list**, not the board. At
  either end of the list the pan resumes, so a short column does not swallow
  the gesture.
- **A board with cards but none on the selected day** gets its own empty state.
  Telling someone with a dozen cards to "create your first" reads as the board
  having lost them.
- **Board updates ride the existing 4s poll.** There is no `task-board` case in
  the SSE reducer, so poll latency is the update latency. Adding one is a
  separate change.

## Localisation

The board's 98 strings are in all nine packs. Packs are partial by design — a
key a pack omits falls back to English — so `sidebar.nav.taskBoard` exists in
English and Ukrainian only, and the rest fall back. `generate-locale.mjs
--check` passes with zero unaccepted strings.

## Tests

127 tests across five files: the HTTP routes (round-trips, the `undefined` vs
`null` distinction, malformed input keeping the old value), the card model, the
day helpers, the layout rules, and the routine link.

## Verification

| Check | Result |
|---|---|
| `npx tsc -b` | clean |
| Board + lib tests | 127 passed |
| `generate-locale.mjs --check` | 0 unaccepted, 10 catalogs valid |
| Full suite | 2 failures, both pre-existing on `upstream/main` |

The two full-suite failures — `server/index.test.ts` "team import is
additive-only" and `server/workspace-backup.test.ts` "refuses external and
arbitrary internal symlinks" — reproduce on a clean `upstream/main` checkout
with none of this code present (verified in a separate worktree). They are
Windows-specific issues in the existing suite, not regressions from this PR.

The wheel behaviour was verified against the running app with a real
input-level wheel, not a synthetic event: mid-list the column scrolled 0→634
while the canvas transform stayed fixed; at the end of the list the canvas
panned while the list stayed put.

## Scope

This PR contains the board and nothing else. The working branch it was ported
from also carried a hermes gateway driver, a Vision key field, dictation, a
prompt library and a Windows local-control path; none of those are here. Files
that are board-only are byte-identical to that branch, and the diff against
`upstream/main` contains zero references to any of those features.

## Not in this PR

Designed but deliberately not built (see `docs/plans/task-board-dates-multibot-todos.md`):
a per-card checklist with an all-done → Done rule, multi-bot cards backed by a
room, and a `runAt` → one-off routine bridge. Each is its own reviewable change.

## Summary by CodeRabbit

- **New Features**
  - Added a Task Board with status columns for organizing work cards.
  - Create, edit, assign, reorder, run, stop, and delete cards.
  - View cards by day or team, with due dates, overdue indicators, and read-only past days.
  - Added a pannable, zoomable board with movable and resizable columns.
  - Routine activity now appears on the Task Board automatically.
  - Added markdown rendering for card briefs and improved bot selection.
- **Localization**
  - Added Task Board translations across supported languages.
- **Documentation**
  - Added Task Board design and handoff documentation.


milind-soni:
Triage decision in #1559: do not merge a parallel task backend alongside #1288/#1332. This PR has a JSON-backed work-items model and its own task-board routes, while the other stack supplies the SQLite claim/heartbeat core. The dated-card/filter/drag UI is distinct work worth preserving, not something we should label already merged. Recommend porting the selected UX onto one canonical task API, then closing this alternative with links to what was retained.

milind-soni:
Closing this alternative implementation as part of the consolidation tracked in #1559. We are choosing one durable task model in the existing SQLite message database, shared by execution, budgets, review and the forthcoming board UI; we will not merge the separate JSON work-items store and parallel /api/task-board API. The dated-card/filter ideas remain recorded in #1559 for the canonical UI. The freeform movable/resizable canvas and read-only-past-day workflow are not being adopted in the initial board. This is an architectural disposition, not a claim that this PR or its UI has already shipped. No source branch is being deleted.

PR #1207, opened 2026-09-14
fix(computer): watchable cloud screen — 1080p frames, live during a turn

guilimasp:
## What changed

Three commits, one symptom: the Computer panel of a **Cloud** bot is not usable for watching the bot work.

1. **`feat(computer)`: capture the cloud panel frame at up to 1920px, quality 85** — `server/box.ts`. The panel/chat frame was `scrot -q 70` and then, for any display wider than 1024px, `convert -thumbnail 1024x -quality 70`. Now: native width up to 1920px, quality 85, `-resize` when a wider display does need scaling. The command is built by `panelShotCommand()` and pinned by `server/box-screenshot.test.ts`.
2. **`fix(computer)`: show the cloud screen while a turn is using it** — `src/lib/local-computer.ts`, `src/components/ComputerPanel.tsx`, `src/lib/remote-desktop.ts`. `resolveBoxPanelAction` takes `busy`; mid-turn a ready box is attached without `provision` (phase → ready, so the turn's live `screen` frames and the busy-cadence screenshot poll show), a not-yet-ready box shows a neutral "your bot is using this computer right now" state, and a provision that races a starting turn (409) falls into that same state. `bot.busy` joins the resolve effect's deps. New locale key `computer.phase.busyBox` (en, pt-br). The cloud-preview fixture now refuses `provision` while Busy is on, like the server, and `docs/verification/cloud-preview.md` gains that step.

3. **`feat(computer)`: draw the bot's pointer into the cloud panel frame** — `server/box.ts`. X11 captures leave the mouse pointer out, so the panel showed the desktop with no trace of where the bot was clicking. `scrot --pointer` (Box image ships scrot 1.10) and ffmpeg `-draw_mouse 1` now include the real pointer; ImageMagick `import` has no equivalent, so that fallback stays pointer-less. Pinned in `server/box-screenshot.test.ts`.

No server route changes: `screenshot` and `join` were never gated during a turn.

## Why

- The frame a person watches was captured at a lower size and quality than the frame the model gets (1280px/q75 in `computer-proxy.ts`), then re-encoded a second time. Browser text was unreadable in the panel and only got blurrier in the image viewer.
- Opening the panel during a turn always POSTed `provision`, which the server refuses with 409 while a turn owns the box. The panel treated that as a fault — "Couldn't reach the computer" + red "being used by an active turn — interrupt it first" — and, because it never reached the ready phase, ignored the live frames the turn was already broadcasting. The one moment you want to watch the screen was the moment the panel refused to show it.

Frame cost: ~3–4× larger (≈100 KB → 300–450 KB). Only the latest frame is kept per bot in the renderer and the chat keeps at most eight with pixels (`MAX_KEPT_SCREEN_FRAMES`), at the same 3–6 s cadence. If that is too much for the chat transcript, a follow-up could pass `?width=` from the panel and keep full size for the viewer only.

## How it was verified

**Tested**
- **On a real Box (ascii.dev), through the desktop dev shell against my own data:** `xdotool getmouselocation` reported x:1050 y:880 and the captured 1920×1080 frame shows the arrow at that spot; during a live turn the panel frame carries the bot's pointer. the Computer panel shows the box's screen at **1920×1080** (was 1024px), the settle frame in the chat is 1080p too, and closing/reopening the panel while the bot was mid-turn showed the live screen immediately — no "Couldn't reach the computer", no active-turn alert — with frames following the bot from page to page.
- **Panel during a turn (client fix)** — the real `ComputerPanel` from this branch on the repo's isolated fake-engine server, with a real bot in a real running turn (`busy=true`) and the provider refusing `provision` with the same 409 the production server sends. Before: "Couldn't reach the computer" + the red alert. After: the phase reaches ready, frames flow, and the panel stays connected when the turn ends. Also covered by the repo fixture (`scripts/verify-cloud-preview.ts`, which now refuses `provision` while Busy is on — step 6 in `docs/verification/cloud-preview.md`).
- Unit tests: `server/box-screenshot.test.ts` (pins the capture command), `src/lib/local-computer.test.ts` (busy → attach / wait), `src/components/remote-desktop-panel.test.ts` (`isActiveTurnRefusal`), i18n. `pnpm typecheck`, `pnpm lint`, `pnpm i18n:check` pass. Full `pnpm exec vitest run`: 6731 passed, 3 failed — `codex-catalog` (reads this machine's Codex setup), `local-computer-proxy` and `engine-install` (electron/npm sandbox); the same three fail on an untouched `upstream/main` here and are unrelated to this change. macOS 26 / Node 22.

## Screenshots (UI changes)

**Real run, no mocks** — this PR's UI and server against a real Box, on a fresh thread. The bot's screen frames are captures from the box at 1920×1080 (the arrow inside the screen is the bot's real pointer, drawn by the capture). The white cursor is the only overlay: it marks where the real clicks were made while recording. Sequence: panel opened while idle → frame at native size → a small task sent → panel closed and reopened mid-turn → screen shown immediately, frames following the bot.

https://github.com/user-attachments/assets/4c7724d6-6837-49f9-98ee-366740d6792d

## Checklist

- [x] `pnpm typecheck` and `pnpm test` pass locally (3 environment-specific failures noted above)
- [x] Server behavior changes come with tests (`server/box-screenshot.test.ts` pins the capture command)

## Summary by CodeRabbit

- **Bug Fixes**
  - Cloud computer panels now connect to an already-ready computer during an active turn instead of showing an error.
  - Added a clear busy-state message while the computer becomes available.
  - Sleep is temporarily disabled during active turns to prevent failed requests.
  - Improved screenshot quality, resizing for wide displays, cursor visibility, and stale-frame handling.

- **Tests**
  - Added coverage for active-turn contention, busy-state handling, and screenshot capture behavior.

- **Documentation**
  - Added verification steps for cloud preview behavior during an active turn.


milind-soni:
Merged current `main` in (this now includes #1211, which drops the local computer relay and runs cloud turns on the box's own agent) and pushed four follow-ups. The short version: **#1211 does not make this branch moot.** Post-#1211 `main` still ships `PANEL_WIDTH = 1024` / `-q 70` / `-thumbnail`, the two PRs touch disjoint regions of `server/box.ts`, and #1211 touches no file under `src/` at all — so both halves here are still fixing live behaviour.

One correction to something I said earlier in review: the client half is **orthogonal** to #1211, not amplified by it. The 409 keys on `bot.busy`, and a cloud bot on the Claude/Codex/pi engines was equally busy before, so the bug fired exactly as often. What #1211 genuinely changes is the *cost* side — `turnInstance()` now routes every cloud-computer bot's ordinary turn to `boxAgent`, so many more turns settle an end-of-turn frame into the transcript, at 3-4x the bytes.

### Pushed

1. **`server/box.ts`** — the comment justifying the new sizing pointed at `computer-proxy.ts`, which #1211 deleted. Reworded to what's now true: this capture is the only frame off the box OpenMausBot sizes at all.
2. **`shared/computer-contention.ts`** (new) — the active-turn 409 was spelled out in four places (server route, panel regex, preview fixture, test). The panel matching the server's *prose* meant a copy-edit server-side would silently turn your new busy state back into "Couldn't reach the computer" — the exact bug this branch removes — and no test would have caught it. One constant now, read by all four.
3. **`ComputerPanel.tsx`** — the resolve catch only recognised the active-turn refusal. But the panel's own screenshot poll holds the box's lifecycle claim while it captures (up to a 60s command timeout), so a provision landing mid-capture is refused with a *different* 409, `"…is being changed — wait for it to finish"`, and fell through to the red fault. It's a wait too, so it re-resolves after 2s using the `isRemoteScreenshotContention` list that already names that string.
4. **`ComputerPanel.tsx`** — `busy-box` drew a static monitor next to copy promising "its screen shows up here as it works", with no spinner. It's a waiting state; it shows one now.

`typecheck`, `lint`, `i18n:check` pass; ComputerPanel, local-computer, remote-desktop-panel and box-screenshot suites are green. I mutation-checked the shared predicate both ways (forcing it false fails the recognition test; narrowing `includes()` to `===` fails the wrapper case).

### Left for you — one design call I didn't want to make

Adding `bot.busy` to the resolve effect's deps means the effect re-runs on `busy` true→false, and for an explicit-Cloud bot `resolveBoxPanelAction` then returns `ensure-box`, so the panel POSTs `/computer/provision` **once per turn end** rather than once per panel open. Server-side that's `provisionBox` → `findBox` → `waitReady` → `mintDesktopUrl` even for an already-ready box. The response's `joinUrl` isn't consumed on that path (only `r.state` is), so skipping the POST when the box is already ready looks safe — but that also changes what opening the panel normally does, so it's your call rather than mine.

Two smaller notes, neither blocking:
- The PR description says "only the latest frame is kept per bot". That's true for *live* frames, but the client also keeps the newest **eight** frames' pixels per thread (`MAX_KEPT_SCREEN_FRAMES`), and the server persists the newest four pngs per thread into SQLite. So the 3-4x is partly a durable-storage and hydrate-payload cost, not purely transient — worth saying plainly in the description.
- Mitigating, and worth noting in its favour: on a display ≤1920 the new command skips `convert` entirely, so box CPU per frame actually goes *down*.

guilimasp:
Review follow-ups landed in de2959fd, one commit for the four points:

- `server/box.ts` — `rm -f "$f"` before the scrot/import/ffmpeg chain, so a stale frame can no longer pass `test -s` as "captured" when every tool fails (pinned in `box-screenshot.test.ts`).
- `ComputerPanel.tsx` — Sleep is disabled while `bot.busy` (the server answers 409 for it mid-turn).
- `ComputerPanel.tsx` — while in `busy-box`, the panel polls `GET /computer` every 5 s and attaches as soon as the box reports a ready state, instead of waiting for the turn to end (`isReadyBoxState` exported from `local-computer.ts`).
- `computer.phase.busyBox` (en, pt-br) now says the screen appears "as soon as the computer is ready".

The description's video is now a real run against a Box (no mocks; the white cursor is the only overlay, marking real clicks).

PR #1211, opened 2026-09-14
Run cloud turns on the Box itself, drop the local computer relay

AnicetNgrt:
## What this changes

When a bot works on its cloud computer (an ascii.dev Box), the whole turn now runs on that computer. The Box already ships every coding agent OpenMausBot supports (Claude Code, Codex, pi, OpenCode, Prime Agent, Kimi) with desktop control built in, so OpenMausBot no longer has to relay clicks, screenshots and shell commands from a local process into the Box.

Removed: `server/computer-proxy.ts`, `server/remote-computer.ts`, their tests, the per-Box install step, and the proxy wiring in the Claude, Codex, pi and ACP drivers. About 2,100 lines gone; Box-related code is 37% smaller.

## Benefits

- Simpler: one way to run a cloud turn, no local relay process, no tooling to install on every Box.
- Faster and more robust: the agent talks to the desktop locally inside the Box instead of through a round trip to the user's machine.
- Works when the user's machine is closed or asleep: routines set to "run on the cloud" complete on the Box alone.
- Clear errors: when the Box cannot run a model (login expired, model not allowed for the account), the reason shows in the chat instead of an empty turn.
- Any model the Box knows can be used: the model list comes from the Box catalog, so a bot moving from a local engine to the cloud keeps its model.

## Two behavior changes to know about

**1. Where a cloud turn runs**

Before: the agent (for example Claude Code) always ran on the user's own machine. When a bot used its cloud computer, OpenMausBot relayed every click, screenshot and shell command from the user's machine into the Box.

Now: when a bot's "Works on" is **Cloud**, the agent runs inside the Box. Nothing is relayed. **Auto** uses the user's own computer if one is available; it no longer attaches a cloud Box to a local agent. Pick Cloud to work on the Box. Local, Local VM and VPS computers behave exactly as before.

**2. Which keys a Box has**

Unchanged: a Box is still created with `noEnv: true`, so nothing from the user's ascii.dev account ever lands in it. Nobody has to visit the ascii.dev dashboard or sign in anywhere new.

New: because the agent now runs inside the Box, OpenMausBot hands the Box the keys it already holds, at creation time: the workspace Anthropic key from App Settings, plus `OPENAI_API_KEY`, `OPENROUTER_API_KEY`, `DEEPSEEK_API_KEY`, `MOONSHOT_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN` and the Kimi token pair when they are set in the server's environment. Those keys are the only ones the Box has.

Limits of this first version:
- A bot whose local engine runs on a subscription login (Claude Pro or Max, ChatGPT) has no key to hand over unless `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) or an API key is configured. Its cloud turn then shows "login required" in the chat instead of running.
- Boxes created by earlier versions of OpenMausBot were created without keys. Their first cloud turn shows "login required"; provisioning a new computer from the bot's Computer panel fixes it.

## ascii.dev features this relies on

- Integrated agents: run a prompt on a Box with any supported agent and model. https://docs.ascii.dev/box/integrated-agents
  - Start a run: https://docs.ascii.dev/box/api/reference/agent/prompt-box-agent
  - Follow a run: https://docs.ascii.dev/box/api/reference/agent/list-box-events and https://docs.ascii.dev/box/api/reference/agent/get-prompt-run-status
  - Stop a run: https://docs.ascii.dev/box/api/reference/agent/interrupt-running-agent-work
- Built-in desktop control for every agent on the Box (the `computer` tool), described on the integrated agents page above.
- Per-Box keys: `POST /boxes` with `noEnv: true` and `env` gives a Box exactly the keys the app passes, nothing from the account. https://docs.ascii.dev/box/integrated-agents#whose-keys-yours-or-your-users
- Model catalog: `GET https://ascii.dev/api/provider-models` (public, no key) lists every agent and the models it can run.

## How it was tested

Live, against ascii.dev, through the normal HTTP API of a running OpenMausBot server (Node 24):

| Scenario | Checks | Result |
| --- | --- | --- |
| Bot set to Cloud, default model (Claude Code, claude-fable-5), Box created with `noEnv` and only the keys OpenMausBot forwarded | reply arrives, Box event log shows the prompt, screenshot, sleep | pass |
| Same bot on Codex and Claude Code with a DeepSeek key | reply, memory across turns, shell command on the Box, interrupt, answer after interrupt | pass (8/8, both agents) |
| Shared team computer (two Auto bots, pi, claude-sonnet-5), same key hand-off | direct turn on the shared Box, room turn on the same Box, second bot refused while busy, interrupt | pass (7/7) |
| Routine set to run on the cloud, same key hand-off | run completes on its own Box, answer lands in the thread | pass |
| Local Claude flows (`scripts/e2e-server.mjs`) | plain turn, permission allow and deny, interrupt | pass |

Unit tests: 1,231 pass across the touched files (`server/index.test.ts`, box tests, driver tests). One test in `server/index.test.ts` ("team import is additive-only") fails the same way on unchanged `main`, so it is unrelated.

## Commits

1. Run cloud turns on the Box's own agent; drop the local computer relay
2. Every cloud turn runs on the Box: shared team computers, interrupts, and the Box's own logins
3. Stops and steers reach the engine that started the turn
4. Boxes keep `noEnv`; OpenMausBot hands each Box the keys it already holds


milind-soni:
This is awesome!

PR #1228, opened 2026-09-14
fix(sidebar): show a teammate wait instead of a work spinner

### LINKED ISSUE #1223: Sidebar spinner conflates a thread that is working with a thread merely waiting on a dispatched teammate
## Summary

The sidebar thread spinner renders one state, "Working", for two very different situations:

1. The thread's own turn is actively running (including long stretches of silent local tool work with no streamed text).
2. The thread's own turn has finished, but the thread handed work to a teammate through a direct coordination handoff and is merely waiting for the result.

In case 2 the user sees a "Working" spinner for the entire duration of the teammate's run, which can be many minutes, with nothing else changing in the UI. Users read this as stuck or hung.

## Evidence on current main (5039833)

- \`server/index.ts\` \`wireTask\` coerces the task to \`busy: true, activity: "working"\` whenever \`activeCoordinationForThread(threadId)\` is true and the task is not already busy (lines 1528-1541). \`activeCoordinationForThread\` is \`roomHandoffs.activeDirect\` (line 2520), which is true while any handoff node for the thread is non-terminal. After the source turn settles, \`sourceSettled\` leaves the node in \`waiting\` while the child works, so the source thread is repainted as busy/working the whole time even though its own turn completed.
- \`src/components/SidebarThreadRow.tsx\` distinguishes \`waiting-on-you\` (lines 105, 143) but maps both \`activity === "working"\` and \`busy\` to the same spinner via \`isWorking\` (line 37).
- \`src/components/Sidebar.tsx\` does the same at lines 110-111.
- For case 1, a turn doing several minutes of local tool calls (for example reading large local files) can stream no text for minutes; the thread row shows an unchanged spinner with no progress signal. I verified on 0.1.78 that such a turn was genuinely running the entire time per the server event log, while the user perceived it as stuck.

The internal threads API already models this better than the UI: it returns distinct running, queued, waiting-on-you, and idle states, and deliberately treats coordinated work as busy for close_thread semantics. The human-facing sidebar collapses that nuance.

## Proposal

Split the human-facing thread state so each situation is legible:

- working: this thread's own turn is running. Keep the spinner.
- waiting-on-teammate: this thread has an outstanding direct handoff. Render like waiting-on-you (clock or hourglass, not a spinner), ideally with the teammate name.
- queued: already exists.
- waiting-on-you: already exists.

Concretely, stop coercing busy in wireTask; instead surface the outstanding-handoff fact (for example activity "waiting-on-teammate" or a coordination.outstanding count) and let the client render it distinctly. Server-side gating that treats dispatched work as busy (queueing, close_thread working checks) can keep using activeDirect directly.

For silent local work, consider surfacing the current activity label (the client already derives "Reading a file" / "Running a command" labels from in-flight tool items in chat) in the thread row tooltip or next to the spinner, so a live turn does not look frozen.

Happy to PR this if the state split approach sounds right; it touches the wire format, sidebar rendering, and i18n, so I did not want to guess at the naming first.

bradhallett:
## Problem

A thread that had dispatched a teammate was repainted `busy: true, activity: "working"` on the wire for the entire teammate run, so the sidebar spun its Working spinner over a parent thread whose own turn was already done (#1223). The person could not tell "this thread is working" from "this thread is waiting on a teammate".

## Fix

**Server.** While a direct coordination handoff is live, the wire keeps painting the handoff `busy: true, activity: "working"` and adds `waitingOnTeammate: true` on the task (and on the bot for its active thread). The paint is load-bearing: `wait_for_conversation` (`control-omb wait`) settles when busy clears, and without the paint wait returned while the teammate chain was still running (the coordination e2e suites caught exactly that on this branch). Existing clients keep seeing what they always saw, busy until the teammate settles, while new clients can tell the two states apart from the flag. Server-side gating (queueing, `close_thread`) still uses `activeCoordinationForThread` directly; nothing behavioral changed there. The task mapping lives in `server/wire.ts` with its own tests.

**Sidebar.** A thread waiting on a teammate renders a quiet wait: a clock icon and "Waiting on teammate" label in the thread row (never the spinner), an accent dot on the bot avatar, and matching text in the bot row, attention list, and thread pickers. The flag outranks the busy paint in every status derivation, so a wait never renders as work. It still demands attention (stays visible past the six recent rows) and ranks between working and queued.

**Live activity label.** While a thread really is working, its row now shows the live verb the chat pane already derives from the visible tail (for example "Reading a file") instead of a generic Working.

**i18n.** Two new keys, `task.waitingOnTeammate` and `sidebar.preview.waitingOnTeammate`, in all ten locales. The translations are hand-authored; happy to regenerate them if you prefer the generator's wording.

The state names (`waitingOnTeammate`, "Waiting on teammate") are a maintainer call; I will rename on request.

## Scope

Groups keep their existing `working` semantics (`publicGroupState` still counts an active handoff as the group working), since a group encompasses its teammates; only per-thread/per-bot state changed.

## Tests

- `server/wire.test.ts`: the wire mapping. Regression for #1223: an idle task under active coordination comes back `busy: true`, `activity: "working"`, `waitingOnTeammate: true`, so wait clients keep blocking until the teammate settles. A busy task stays working and unflagged; a task outside coordination is untouched with provider bookkeeping stripped.
- `SidebarThreadRow.test.ts`: a waiting row renders the label over the busy paint and never `animate-spin`; a teammate wait ranks between working and queued even when painted busy; an older waiting thread stays visible past the six recent rows; a working row shows the live activity label.
- `SidebarBotListItem.test.ts`: a bot-level wait shows `teammate-wait-dot` with no working dot and no typing dots, even over the busy paint.

Targeted `vitest run`, `tsc -b` plus `tsc -p tsconfig.server.json`, `oxlint --deny-warnings .`, and `generate-locale.mjs --check` all pass locally.

Fixes #1223

## Summary by CodeRabbit

- **New Features**
  - Added a clear “Waiting on a teammate” status for bots and tasks.
  - Sidebar and task picker now show teammate-waiting indicators, labels, and accessible status text.
  - Waiting status takes precedence over active-work indicators while remaining visible in attention lists.

- **Localization**
  - Added teammate-waiting translations across supported languages.

- **Tests**
  - Added coverage for teammate-waiting indicators, labels, visibility, ordering, and working-state behavior.


bradhallett:
CI note for 5a347e61: the Kotlin tests + Android build leg failed on PairingClientTest.hostedGatewayRetriesTailnetButNeverLan with a java.util.ConcurrentModificationException at PairingClientTest.kt:143 (525 of 526 tests passed). The only change in 5a347e61 versus the previous head is two lines in src/components/Sidebar.tsx, and the same job passed on the previous head, so this looks like an unrelated flake. A re-run of that one job should clear it.

bradhallett:
CI note: the `Kotlin tests + Android build` failure on 5a347e6 is unrelated to this change. `PairingClientTest.hostedGatewayRetriesTailnetButNeverLan` failed with a `ConcurrentModificationException` at PairingClientTest.kt:143, while this diff touches only `src/components/Sidebar.tsx`. The same check passed on the previous head of this branch (85b1d91) and on main, and the test stub records requests into plain lists from the HTTP client threads while assertions iterate them, so it is timing sensitive. A rerun should clear it; I cannot rerun jobs on this repo.

The windows-latest red is the inherited main failure (codex.test.ts wording assertions at lines 1338 and 1353) and clears once #1222 merges.

bradhallett:

The `Kotlin tests + Android build` leg fails on the pre-existing `PairingClientTest.hostedGateway` flake already noted on the earlier commits of this PR; the desktop sidebar diff does not touch it.

milind-soni:
Thanks — the server half, i18n and tests are clean, and the idea is right (#1223). One blocker before merge:

**The bot-level `waitingOnTeammate` flag is never cleared on the client.** `wireBot` omits the key once coordination settles, and `botPatched` merges `{ ...b, ...action.bot }` over the previous bot — so a bot that was once waiting keeps `waitingOnTeammate: true` forever: the row keeps saying "Waiting on a teammate…" and the working spinner/mascot stay suppressed after the wait ends.

- [src/state/store.tsx:1422](https://github.com/milind-soni/OpenMausBot/blob/HEAD/src/state/store.tsx#L1422) — `botPatched` merge; the sibling omitted-key fields (`browserProfile`, `computer`, `section`) are already carried explicitly there.

Fix: add `waitingOnTeammate: action.bot.waitingOnTeammate,` to the `updateBot(...)` merge (and the `taskSwitched` bot spread at ~1418), plus a reducer test: a bot with `waitingOnTeammate: true` receives a complete `bot` frame without the key and comes back `undefined` — for both a `busy:true` working frame and a `busy:false` idle frame. Then this is good to go.

bradhallett:
Ack. Queued the client fix: clear bot-level waitingOnTeammate when wireBot omits the key, handling it like the sibling omitted-key fields in the botPatched merge at store.tsx:1422, so rows stop showing a stale wait.

CI note: on run 35113438354 the only failing leg is macos at Verify native approval mode transitions (smoke-approval-modes.cjs:228, list_bots reads 3 vs expected 2), identical to upstream main's own CI run 35108771265 at b94f4a61 where ubuntu and windows pass. Main-carried, no PR change owed. Fork Vercel red is the known deploy authorization noise.

bradhallett:
Re: your blocker on the bot-level waitingOnTeammate flag - fixed in 4d53fbb0. botPatched now carries waitingOnTeammate explicitly next to computer/section/browserProfile, so a frame that omits the key clears the stale flag instead of inheriting the old value; the taskSwitched and switched-thread paths pass it through the same way. Reducer coverage is in the teammate-wait block of src/state/store.test.ts: a bot seeded with waitingOnTeammate: true receives a complete frame without the key and comes back undefined, for both a busy working frame and an idle frame, plus a still-carrying-it guard. The code quoted predates that commit - ready for another look.

bradhallett:
The blocker from the last review is fixed in 4d53fbb0. I also updated the branch from `main` (merge 50e8cf90): the wire view now emits both main's `waitingForTeammates` and this branch's `waitingOnTeammate` quiet-wait paint, with the `wireTaskFor` tests updated to match. Could you re-review when CI is green?

PR #1233, opened 2026-09-15
feat(mobile): float attention threads to the top within a bot

bradhallett:
## Gap

Mobile thread lists keep stored order, so threads that need attention sit wherever they landed. Desktop floats them: `orderedSidebarThreads` / `attentionRank` in `src/components/SidebarThreadRow.tsx` stable-sorts each bot's threads by attention rank and never filters (#1199).

## Change

Both platforms now order threads by the same attention tiers: waiting-on-you, then busy/working, then queued, unread, current, idle.

- iOS `ThreadNavigation.swift`: `attentionRank` plus `threadsInAttentionOrder` applied before grouping, with a stored-index tiebreak to keep the sort stable; search results keep relevance order.
- Android `ThreadNavigation.kt` and `TaskRules.kt`: the same tiering through a stable sort; search keeps relevance order; the task sheet sorts within its open and closed halves.

Ordering only repositions. Tests on both platforms assert the full list survives, and the tiers match desktop one for one.

## Tests

- `swift build`, `swift test` (417/417, 18 ThreadNavigationTests)
- `./gradlew --no-daemon :core:test :app:testDebugUnitTest :app:assembleDebug :app:assemblePreview` (all green)

Rides the mobile parity umbrella #1197 and the desktop pattern in #1199.

## Summary by CodeRabbit

- **New Features**
  - Threads are now prioritized by attention level, bringing tasks waiting for action, in progress, queued, or unread ahead of idle history.
  - The current thread is surfaced above idle threads.
  - Ordering is applied consistently across Android and iOS navigation views.

- **Bug Fixes**
  - Search results continue to preserve relevance ordering.
  - Threads with equal attention levels retain their existing order.


bradhallett:

PR #1236, opened 2026-09-15
feat(threads): opt-in parking for direct messages behind delegated work (#1194)

### LINKED ISSUE #1194: Direct conversations: a new user message cancels the bot's in-flight delegated work; please make room-style parking available
## What happens today

When a person sends a new message in a **direct conversation** while the bot has delegated work in flight, the server cancels all of that thread's pending handoffs before the new turn starts:

server/index.ts:4837

    if (!opts?.coordination && !opts?.cardContinuation) roomHandoffs.cancelDirect(threadId, "Superseded by a new message in this conversation");

`cancelDirect` (server/room-handoffs.ts:149) cancels every non-terminal handoff node for that thread. Room coordination turns and approval-card continuations are exempt; a plain human message is not.

## Why it hurts

A casual follow-up ("how's it going?") silently kills long multi-step delegated work. There is no UI indication that anything was cancelled, the agent's continuation never sees the result, and the work has to be re-issued from scratch. The only workaround is to know in advance that delegation-heavy work belongs in a room.

## What rooms already do right

In rooms, a member's in-flight turn is never touched: the round parks on the busy member with "X is finishing another conversation — will reply here when free" (server/index.ts:2538, pinned by server/room-chat-wait.e2e.test.ts). The parking machinery already exists.

## Proposal

Make the direct-conversation behavior configurable, reusing the room parking semantics:

- a per-thread or per-bot setting such as "queue my messages behind running work" (the current cancel behavior can stay the default), or
- a workspace-level toggle for direct conversations.

If cancelling stays the default, two smaller fixes would still help a lot:

- surface the cancellation in the UI ("this message stopped pending work"), and
- deliver the cancellation reason into the bot's next turn so it can re-issue the work instead of losing it silently.

## Environment

Desktop app 0.1.78; behavior verified against main at commit 83678c99.

bradhallett:
## What

Adds a per-bot `parkDirectMessages` opt-in for direct conversations. While it is on, a plain user message that arrives while this bot still has delegated assignments out does not steer the running work: it waits in the existing composer queue (the same pending chip a busy thread shows, cancellable and durable) and runs as its own follow-up turn once every outstanding assignment has settled and the resumed coordination has finished.

The default is unchanged: messages keep steering immediately, exactly as #1171 established. That PR already removed the cancel-on-new-message behavior this issue reported, so this change only adds the parking mode the issue proposed on top of it.

## How

- `store.ts`: `parkDirectMessages?: boolean` on the bot record; unset keeps steer-immediately.
- `index.ts`: a `parksBehindCoordination` guard joins both the condition that queues an incoming direct message and the condition that holds a queued message back in the drain. It parks only while a direct coordination for that thread is still alive, so the parked turn can never jump the coordination's own resume turn.
- Releasing the parked words keys on handoff node changes, not only `turn.completed`: the tree's final settle emits no later turn event the queue could drain on (the resume turn's completion lands while the source node is still non-terminal), so the handoff `changed` hook now also drains queued sends for direct threads.
- `PATCH /api/bots/:id` accepts the boolean.

## Tests

- Default path (pinned): a message sent while a teammate works still runs immediately as a steering turn; the assignment is kept and named in that turn's prompt.
- Opt-in path (new): with `parkDirectMessages` on, the message is queued (`queued: true`), never reaches the transcript early, the assignment is not cancelled, and after the teammate result resumes the coordination and the tree settles, the parked words run as their own turn without the outstanding-assignments prompt.
- The full direct-coordination and room-chat-wait e2e files pass, along with typecheck and lint.

## Deferred

A settings-dialog toggle in the desktop app; the flag is settable today via `PATCH /api/bots/:id`.

Fixes #1194.

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to queue direct messages while delegated teammate work is in progress.
  * Queued messages run as a follow-up turn after coordination settles, while the default steering behavior remains unchanged.
  * Invalid values for the new setting are rejected.

* **Documentation**
  * Documented the optional direct-message parking behavior and its interaction with Stop.


bradhallett:

PR #1240, opened 2026-09-15
feat(threads): cross-bot attention inbox on mobile

bradhallett:
## What

Mobile parity for the cross-bot attention inbox (#1206) on iOS and Android, part of the mobile parity umbrella #1197.

- New Needs attention section above the roster, aggregating per-bot attention rows from the tasks array already in GET /api/bots.
- The whole fleet is flattened and ordered once, never per-bot, mirroring desktop SidebarBotActivity.tsx plus TaskPicker.tsx including 7fedb278: an earlier bot's unread reply cannot outrank a later bot's waiting approval.
- Tier ladder identical to desktop attentionRank with an empty activeId: waiting > working/busy > queued > unread.
- Hidden bots and the excluded bot stay out; routine runs are dropped; flags are per task, not per-bot aggregates.
- Rows jump straight to the thread; accessibility labels on both platforms.

## Evidence

- Android commit cf167508, iOS commit 63f5542e, both DCO signed.
- Gates: `swift build`, `swift test` 418/418, and the Gradle gate all green.
- Cross-bot ordering tests on both platforms, hidden and excluded bot tests, a queued-from-client-map test on Android, and iOS parses activity `queued` defensively per the wire contract.

## Notes

- Rides #1197 (umbrella) and #1206 (desktop inbox).

## Summary by CodeRabbit

- **New Features**
  - Added a cross-bot “Needs attention” section to the Android and iOS companion apps.
  - Highlights waiting, working, busy, queued, or unread threads, prioritized by urgency.
  - Displays the bot, thread title, and current status; selecting an item opens the relevant conversation.
  - Excludes hidden bots, routine runs, and the currently selected bot.
  - Includes fallback attention items for bots without detailed task lists.

- **Tests**
  - Added coverage for prioritization, filtering, queued items, legacy task handling, and status displays.


bradhallett:

milind-soni:
The attention rule, the hidden/excluded-bot and routine-run filtering, and the ordering are correct on both platforms and covered by tests — nice parity with desktop. One blocker on iOS:

**Tapping an attention row does nothing.** `openAttention` appends the raw `Bot` to the `NavigationPath`, but the stack's only destination is `Chat.self`, so no navigation happens (it compiles because `Bot` is `Hashable`). CodeRabbit flagged this and it's still open at 9c4856d3.

- [ios/App/ChatListView.swift:565-568](https://github.com/milind-soni/OpenMausBot/blob/HEAD/ios/App/ChatListView.swift#L565)

Fix: `guard let bot = session.state.bot(forThread: entry.task.threadId) else { return }` then `path.append(Chat.bot(bot))`, matching lines 170 and 536 and the single `.navigationDestination(for: Chat.self)` at line 116. Optionally pass `activity:` through for parity with desktop `SidebarBotActivity.tsx:11-14` and Android `AttentionInbox.kt:76-80`.

bradhallett:
Ack. Queued the iOS fix: resolve the bot for the thread in openAttention and append Chat to the path so the row navigates; the open CodeRabbit finding gets answered with the fix.

CI note: on run 35113449014 the only failing leg is macos at Verify native approval mode transitions (smoke-approval-modes.cjs:228, list_bots reads 3 vs expected 2), identical to upstream main's own CI run 35108771265 at b94f4a61 where ubuntu and windows pass. Main-carried, no PR change owed. Fork Vercel red is the known deploy authorization noise.

bradhallett:
CI note for run 35142055579: the red windows leg fails exactly two tests in server/drivers/codex.test.ts ("SyntaxError: Unexpected end of JSON input" plus a 20s test timeout), with 6990+ tests passing. The same two-test signature hit runs on three branches in this window that do not touch the codex driver, so it is either a runner flake or carried by current main; the decisive main run is still queued. No code change owed on this branch.

bradhallett:
Fixed in 365c444 — openAttention now resolves the thread's bot and appends Chat.bot(bot) to the path (a missing bot stays a safe no-op; covered by two new CompanionCore tests), and the fix now sits on top of the refreshed head.

bradhallett:
CI note for run 35159188430 (365c444): the only red legs are typecheck+test ubuntu and macos, each failing the same two browser-mount codex integration tests (default and absolute CLI variants, expected ... to include 'done from fake codex') that main 53dd47fc own run 35148261823 fails on the same two legs, with windows and the Swift/iOS leg green on both runs - main-carried, note only. This diff touches only ios/ files and the legs that exercise it (Swift tests + iOS build + thread UI, windows typecheck+test) are green here; older main 5dbce1ff run 35139548557 was fully green, so the family is intermittent. Fork Vercel red is systemic and ignored.

bradhallett:
CI note: run 35219149301 (27166c00) failed only windows vitest shard 4/4, on server/delta-context.e2e.test.ts (expected 'failed' to be 'settled' in the fake-CLI fixture). This diff is mobile-only with no server-file overlap, and the same suite passed on #1384, #1282, and main (35218327969) in the same window. Classified as a flake. Pushed empty retrigger bca148a7; fresh run 35230369908.

PR #1244, opened 2026-09-15
feat(harness): phase 0 part 3 — typed turns, command filters, headless bench driver

aivsomkar:
Phase 0 of the harness upgrade, part 3 of 3 (stacked on #1243). Decisions become validated objects, noisy commands can be bounded, and there is a headless driver that benchmarks and the harness's own regression runs can call.

## What changes

- **Typed turns** (`server/typed-turns.ts`, `SendTurnInput.outputSchema`, `turn.completed.structured/structuredError`): the harness appends one fixed instruction naming a JSON schema, extracts the last fenced JSON block from the reply and validates it in code. Same path on every engine; a driver with native constrained decoding may pass its object in and it is re-validated. First consumer: the goal room's coordinator decision is schema-first, with the prose envelope kept as fallback, and a run without a valid decision fails closed with the reason on the card. Native accelerators are deferred (finding F4).
- **PreToolUse command filters** (`server/hooks/filters.ts`, per-bot `commandFilters`): known noisy shell commands (test runners, installs, unbounded `git log`) rewritten to a bounded form through Claude Code's `updatedInput`. Never blocks, never rewrites unknown commands. Off by default; every rewrite counted in `/api/metrics` (finding F5).
- **Headless bench driver** (`server/bench.ts`, `POST /api/bench/run`, `GET /api/bench/runs/:id`, `scripts/bench/run.ts`, `pnpm bench:run`): one task on one bot as an unattended launch inside a steps/tokens/minutes budget; result and trajectory readable by a script; books under its own ledger trigger `bench`. `docs/bench/2026-09-baseline.md` records the fixture rows that were run and states that live Terminal-Bench numbers are not run yet.
- Verification recipes: `docs/verification/digests.md`, `hooks.md`, `bench.md`; the plan is marked built with findings F1–F5.

## Test plan

- [x] `typed-turns.test.ts`, `group-goal-run.test.ts`, `hooks/filters.test.ts`, `bench.test.ts`, `metrics.test.ts`, `drivers/claude.test.ts`
- [x] `typed-turns.e2e.test.ts`: decision block on fake Claude (split across two assistant items, never shown in the room), ACP echo, and fail-closed on Codex and pi
- [x] `hooks.e2e.test.ts`: filter rewrites `pnpm test` only for the flagged bot and is counted
- [x] `bench.e2e.test.ts`: settled runs on fake Claude, Codex, ACP and pi; step and time budgets; bad requests; the script end to end
- [x] full vitest suite green on this tip (542 files, 6,728 tests)
- [ ] not done: live Terminal-Bench run (needs a signed-in engine and a paid account; recipe in the baseline doc)

## Platforms
|---|---|---|
| macOS     | yes | in this PR; automated suites only, NOT exercised in a local app build |
| Windows   | yes | in this PR (shared server code, no platform gate); NOT built on Windows. `filters.ts` appends `2>&1 \| tail -n N`, which assumes a POSIX shell behind Claude Code's Bash tool (Claude Code uses Git Bash on Windows) |
| iOS       | n/a | no phone-facing payload changed; `commandFilters` has no settings UI anywhere yet (Phase 1 UI pass) |
| Android   | n/a | same |
| Companion | n/a | bench routes are desktop/script-only; nothing a phone calls changed |


milind-soni (changes requested):
The shell output filters change command success semantics. Reproduced from this head: with an npm stub returning 7, `npm test` exits 7 but the generated `npm test 2>&1 | tail -n 200` exits 0. This can hide failing tests/installations from the model and downstream evidence. Preserve the original command exit status (including Windows/Git Bash support), or bound captured output without rewriting commands. Add an executable regression for nonzero status, not just string matching.

PR #1242, opened 2026-09-15
feat(harness): phase 0 part 1 — command receipts, work digests, Claude Code hooks

aivsomkar:
Phase 0 of the harness upgrade (`docs/plans/2026-09-14-phase-0-foundation.md`), part 1 of 3. The harness starts seeing what a bot *did*, not only what it said, on every engine.

## What changes

- **Command receipts** (`server/commands.ts`): new mutations on the transcript DB run once, keyed by a command id, so a re-delivered event never double-writes.
- **Work digest per turn** (`server/digest.ts`, message kind `digest`): after every settled turn, one row names the tools called, the files added or changed in a checkpointed folder, memory saved, and how good the evidence is (`full` / `preview` / `none`). Digests ride along in context rebuilds and room context, and are searchable. Renderer shows a small chip.
- **Claude Code hooks** (`server/hooks/omb-hook.ts`, `POST /api/internal/hook`): registered through the private `--settings` file the driver already writes. `PostToolUse` delivers the full tool result (spilled to `DATA_DIR/tool-results`, 0600, attached to the activity row); `PreCompact` / `SessionStart(compact)` record a compaction as chips and hand the last two digests back as plain-text context. The helper always exits 0 within 4 s and speaks only to loopback. `OMB_HOOKS=0` disables it.
- Every engine gets the digest; Claude hooks are an accelerator on top, never the baseline (Codex, pi and the ACP family carry tool results through their own protocol; the HTTP family and box agent report `none`).

## Test plan

- [x] `server/commands.test.ts`, `digest.test.ts`, `hooks/omb-hook.test.ts`, `message-db.test.ts`
- [x] `server/digest.e2e.test.ts`: one digest per turn on fake Claude, Codex, pi and ACP; a gated ACP turn edits two files, the digest names them, and a second engine sees the digest in its replay
- [x] `server/hooks.e2e.test.ts`: full results spilled and attached, forged bearer refused, compaction chips and digest context, `OMB_HOOKS=0` registers nothing
- [x] full vitest suite green on the three-PR chain (this branch + #1243 + #1244)
- Recipes: `docs/verification/digests.md`, `docs/verification/hooks.md` (added in part 3)

## Platforms
| Platform | Applies? | Status |
|---|---|---|
| macOS     | yes | in this PR; automated suites only, NOT exercised in a local app build |
[...]

PR #1245, opened 2026-09-15
feat(ios): digest and compaction chips (phase 0 receipts)

aivsomkar:
iOS half of the phone follow-up named in #1242 and #1243 (phase 0 parts 1 and 2), stacked on #1243, which introduces the two message kinds this renders.

## What changes

- `Message.Kind` gains `digest` and `compaction`; `Message` gains `digest: WorkDigest?` and `compaction: Compaction?` with the fields the chip counts from. Keys the phone does not read stay unread, so a newer digest never fails the decode.
- Chat: a quiet `ReceiptChip` capsule under the reply, "4 tool calls · 2 files" or "Context compacted · 12,345 tokens summarised", tap for the full text in the selection sheet.
- Hidden together with tool activity (`isActivityReceipt`), never folded into a run of it. Roster preview and the updates line use the chip text; the voice never reads a receipt aloud.

## Test plan

- [x] `swift test` (CompanionCore): 417 passed, including the new `DecodingTests` (digest counts, digest without files, compaction record) and `ChatPreferencesTests` (hidden drops the receipts, reduced keeps a digest as its own row)
- [x] App target simulator compile gate (`xcodebuild … -sdk iphonesimulator CODE_SIGNING_ALLOWED=NO build`): BUILD SUCCEEDED
- [ ] not run: on-device look (`ios/TESTING.md` stage 4)

## Platforms
|---|---|---|
| macOS     | n/a | phone-only change |
| Windows   | n/a | phone-only change |
| iOS       | yes | in this PR |
| Android   | yes | #1246 (twin) |
| Companion | n/a | no route changed; the phone reads the kinds off the existing messages route |


DaNewChamp:
Confirmed the user-visible issue this PR addresses against a stock v0.1.85 hosted workspace: the native iPhone chat shows raw `[digest] · tools: … · reply: …` messages after assistant replies. Approval-heavy workflows make the duplicated-looking bubbles especially noticeable.

The source cause is still present on current `main` (`3eb9046`): `Message.Kind` decodes `digest` as `.unknown`, so it falls back to an ordinary text bubble. The personal Activity → Hidden preference filters only `.activity`, so it does not hide these receipts. The web renderer already handles digest rows separately.

This PR's explicit receipt kinds, compact rendering, and `isActivityReceipt` filtering address that mismatch. I reviewed the included Hidden/Reduced regression tests and the successful **Swift tests + iOS build** CI check; I have not run an on-device build of this branch. The overall PR checks are not all green, so this is not a blanket verification claim.

Adding this reproduction here rather than opening a duplicate fix or maintaining a private iPhone build. An official release containing this change would let users hide the noise without deleting the server's useful digest/context records. Is the iOS fix still intended to land through the stacked base PR, or could it be landed separately?

No backend workaround or private conversation/screenshot is included in this report.

milind-soni:
Included the useful compaction-receipt UI in #1755, now merged into main. Thank you for the implementation. We retained the current policy of hiding routine work-digest messages, and adapted this patch without importing the still-unreviewed #1243 foundation branch. The combined iOS/Android builds and regression suites passed. Closing this PR as incorporated in adapted form; #1243 remains separate.

PR #1246, opened 2026-09-15
feat(android): digest and compaction chips (phase 0 receipts)

aivsomkar:
Android twin of #1245, stacked on #1243 (phase 0 part 2), which introduces the two message kinds this renders.

## What changes

- `Message.Kind` gains `DIGEST` and `COMPACTION`; `Message` gains `digest: WorkDigest?` and `compaction: Compaction?` with the fields the chip counts from. Unknown keys stay ignored, so a newer digest never fails the decode.
- Chat: a quiet `ReceiptChip` under the reply, "4 tool calls · 2 files" or "Context compacted · 12,345 tokens summarised", tap to expand the full text. Copy gives the full text.
- Hidden together with tool activity (`isActivityReceipt`), never folded into a run of it. Roster preview and the updates line use the chip text.

## Test plan

- [x] `DecodingTest`: digest counts, digest without files, compaction record (chip text)
- [x] `ChatPreferencesTest`: hidden drops the receipts, reduced keeps a digest as its own row and breaks the run
- [ ] NOT compiled on the authoring Mac (no Java runtime there); this PR's CI runs `:core:test` and `:app:testDebugUnitTest` and builds both APKs. Please treat a red CI as mine to fix.
- [ ] not run: on-device look (`android/README.md`)

## Platforms
|---|---|---|
| macOS     | n/a | phone-only change |
| Windows   | n/a | phone-only change |
| iOS       | yes | #1245 (twin) |
| Android   | yes | in this PR; NOT compiled locally, CI compiles it |
| Companion | n/a | no route changed; the phone reads the kinds off the existing messages route |


milind-soni:
Included the useful compaction-receipt UI in #1755, now merged into main. Thank you for the implementation. We retained the current policy of hiding routine work-digest messages, and adapted this patch without importing the still-unreviewed #1243 foundation branch. The combined iOS/Android builds and regression suites passed. Closing this PR as incorporated in adapted form; #1243 remains separate.

PR #1247, opened 2026-09-15
feat(mobile): voice engine picker with Chatterbox

bradhallett:
## Problem

The companion apps could see which voice engine the paired computer uses but could not choose it, and Chatterbox was missing entirely. The server already ships a Chatterbox provider (server/tts/chatterbox.ts) and the desktop picker offers it (src/components/VoiceSettings.tsx), so a phone user who set Chatterbox on the computer had no way to change the engine or finish setup from the phone.

## What this does

Two commits on main, one per platform:

- iOS (5ec5508f): VoiceProvider gains chatterbox with the exact wire string, the bot settings sheet picks between all three engines over the ordinary config write, and a Chatterbox address and model id commit together in one write. Built-in Mac voices stay selectable only when the server reports a darwin host, and the not-configured copy names the engine it belongs to.
- Android (2d9f8ba6): the same enum, picker, and wiring. The engine write is PUT /api/config carrying only {"tts":{"provider":"chatterbox"}} on both platforms, and decoding keeps the server's own fallback: a missing or unknown engine means ElevenLabs.

Deviations, declared: the desktop "Voice output on this Mac" local-speech toggle is Electron-only and is not ported. Key material never returns to the phone, so ElevenLabs key entry stays on the computer, and Android also defers the Chatterbox address to the computer (iOS can save address and model, which the server round-trips; the key it does not). Android offers the picker on every platform because the phone cannot know the host OS; the configured status explains an engine that cannot run there.

## Tests

- iOS: swift build, swift test (415/415), simulator xcodebuild build.
- Android: ./gradlew --no-daemon :core:test :app:testDebugUnitTest :app:assembleDebug :app:assemblePreview.

Part of the mobile parity program: #1233, #1234, #1240, #1241 are the first four gaps.

## Summary by CodeRabbit

- **New Features**
  - Added voice engine selection across Android and iOS.
  - Added support for the Chatterbox voice engine alongside ElevenLabs and built-in voices.
  - Added Chatterbox server address and model configuration on iOS.
  - Added saving and switching controls with progress and error feedback.
  - Added platform-aware availability for built-in voices.

- **Bug Fixes**
  - Improved handling of missing or unrecognized voice engine settings by safely falling back to the default engine.


bradhallett:

PR #1253, opened 2026-09-15
feat(mobile): show queued sends from client queue state

bradhallett:
## Problem

Desktop shows messages the harness holds while a thread is busy: the queue row sits by the composer with a cancel control, a closed thread holding a send floats back into the list, and Updates counts it. Mobile showed none of that. A held send vanished until the thread finished, a closed thread with a queued message stayed folded away, and the pill ignored it.

## What this does

Two commits on main, one per platform: 829dedbe (iOS) and c2ac9545 (Android).

- Queued is client state, never task activity. Both platforms mirror the desktop queue from three sources (send receipts, bot.queued frames, and the fleet snapshot) into a pendingQueued store, and thread rows, the closed-thread fold, search, roster, and the Updates pill derive Queued from that state. The dead queued activity string is gone from parsing; running and waiting stay untouched, pending the gap 2 wire cleanup.
- Bounded drained-queueId tombstones (64) win the race where a drain beats the POST response, so a vanished queue row retires instead of resurrecting. Landed lines retire their rows on the message frame, and whole-transcript reconcile catches the rest.
- Held sends list above the composer with per-row delete through the DELETE queue route; the bot route carries threadId in the body, matching desktop store.tsx. A 404 reading "no such queued message" counts as drained; any other 404 surfaces as an update-your-computer error instead of silently dropping a still-queued message.
- bot.queued frames parse defensively: malformed entries drop and the frame never fails. An unreadable send-response body degrades to not-queued rather than failing a succeeded send.

Deviations from desktop: the send response's message is not appended locally; the stream delivers the real line, consistent with the app's existing design. The Steer affordance in the queued composer stays out of scope. Queued rows are derived at view time via queuedThreadIds, the desktop's own pattern in Sidebar.tsx, rather than stored on wire-decoded structs. Android passes queuedThreadIds as a parameter where iOS reads session state directly. One label nuance: a thread that is busy with non-working activity and also holding a send reads Working on desktop and Queued on both mobiles; the row surfaces and floats either way.

## Tests

- iOS: swift build, swift test (429/429), simulator build. New QueuedSendTests cover the state machine, the tombstone race and its bound, defensive decode, and the closed-thread float; QueuedSendClientTests pin the cancel wire contract.
- Android: :core:test, :app:testDebugUnitTest, :assembleDebug, :assemblePreview, 87/87 tasks green on the first run. New tests pin the held-send float, the queuedThreadIds derivation, and that the dead activity string stays dead.

Neither half ran on a physical phone; the composer's held-send list and cancel flow rest on unit tests and review.

Part of #1197 (inbox-style thread management) and the final lane of the mobile parity program after #1233, #1234, #1240, #1241, #1247, #1248.

## Summary by CodeRabbit

* **New Features**
  * Queued messages now appear across thread lists, tasks, search results, and status updates.
  * Displays whether one or multiple messages are waiting.
  * Added cancellation for queued messages from conversations.
  * Closed threads with queued messages remain accessible until processing completes.
  * New bots open immediately after creation.

* **Bug Fixes**
  * Queued status now reflects actual pending messages.
  * Improved synchronization as queued messages are delivered or removed.
  * Improved conversation ordering and archived-thread handling.


bradhallett:

bradhallett:
4213ada5 restores the creatingBot in-flight guard: the New bot button disables while createBot runs, mirroring the existing creatingThreads gate.

milind-soni:
The iOS half is complete and matches the server contract (send receipt + `bot.queued` frame + fleet `botQueuedMessages`, wholesale replace with drained tombstones, per-row cancel with the harness's 404 wording) — 40 tests pass locally. Holding on the Android half, which isn't yet what the PR describes:

**Android has only one queue source (its own send receipts).** There's no `bot.queued` frame handling (it falls to `Frame.Unknown(kind)`) and no fleet `botQueuedMessages` snapshot, and `hydrate` never resets `pendingQueued` — so roster rows, the closed-thread float and the Updates pill derive "Queued" from a store that can't see queues created elsewhere and can't forget drained ones across a reconnect.

- [android/core/.../Store.kt:47](https://github.com/milind-soni/OpenMausBot/blob/HEAD/android/core/src/main/kotlin/com/openmausbot/companion/core/Store.kt#L47) `queuedThreadIds` (new) read by ThreadNavigation, TaskRules, BotThreadRow, Updates.kt:69
- [android/core/.../Frames.kt:132](https://github.com/milind-soni/OpenMausBot/blob/HEAD/android/core/src/main/kotlin/com/openmausbot/companion/core/Frames.kt#L132) `bot.queued` → `Frame.Unknown`
- Models.kt:426 `data class Fleet(val bots, val groups)` — no `botQueuedMessages`

Ask: give Android the same two server-owned sources iOS gets here — parse `bot.queued` into a `Frame.BotQueued(queues)` applied with wholesale-replace + tombstone-vanished-ids (mirroring iOS `CompanionState.replaceBotQueues`), and decode `botQueuedMessages` in `Fleet` and re-seed `pendingQueued` in `hydrate`. Happy to merge once both platforms agree.

bradhallett:
Ack. Android half queued: handle the bot.queued frame, take the fleet botQueuedMessages snapshot with drained tombstones, and reset pendingQueued on hydrate, matching the iOS contract you verified.

CI note: on run 35113457535 the only failing leg is macos at Verify native approval mode transitions (smoke-approval-modes.cjs:228, list_bots reads 3 vs expected 2), identical to upstream main's own CI run 35108771265 at b94f4a61 where ubuntu and windows pass. The windows leg on this run was cancelled, not failed. Main-carried, no PR change owed. Fork Vercel red is the known deploy authorization noise.

bradhallett:
Replying to https://github.com/milind-soni/OpenMausBot/pull/1253#issuecomment-5700571529

Pushed `8816505e`: the Android half now consumes the server\x27s own queue state (`bot.queued` frames, fleet snapshot re-seed on hydrate, drained queues forgotten across reconnect), and the branch is refreshed onto current main with both review gates green.

bradhallett:
CI classification for run 35164675148 (head 8816505e):

- ubuntu + macos: both fail exactly the two `server/browser-codex-path.integration.test.ts` cases with `expected to include 'done from fake codex'` — the same failures main itself has on its latest full run at 53dd47fc (run 35148261823), so main-carried.
- Swift: fails only at the `Verify thread navigation on disposable iPhone and iPad simulators` step (`ThreadNavigationUITests.swift:142`); the same step failed this hour on #1384's run, whose diff contains no iOS code, and it was classified as a simulator flake on #1320 (run 35142066576).
- windows: `server/comms.test.ts > legacy routine comms e2e (fake ACP fleet) > keeps a Chief available while its delegated teammate is still working` — this PR's diff is mobile-only (android/ios), so the server code under test is identical to main, whose latest full run passed the windows leg (run 35148261823). Runner flake, no action.
- Fork Vercel red is the standing authorization-only status.

Holding any retrigger while main's own run is red, since a rerun cannot go green until main is fixed.

PR #1254, opened 2026-09-15
fix(comms): teach coordinate_bots argument errors and accept aliases

### LINKED ISSUE #1239: coordinate_bots: accept camelCase arguments or name the rejected field in the usage error
## What happens

The agents MCP bridge declares `coordinate_bots` with snake_case-only arguments — `server/drivers/agents-proxy.ts` **L409–416**: `inputSchema` with `additionalProperties: false` and `required: ["bot_ids", "message", "request_key"]` (`group_id`, `bot_ids`, `message`, `request_key`, `rework`, `label`), with the handler reading exactly those keys at **L992–996**.

camelCase payloads — which MCP clients commonly send (`botIds`, `requestKey`) — fail validation with a generic usage error that does not name the offending field, so the caller cannot tell which key was wrong. Observed when a coordination return had every payload rejected this way and no work ran.

## Ask

Either:

- accept both shapes (map camelCase to snake_case on the way in), or
- return a field-specific error naming the expected key, e.g. `expected "bot_ids", got "botIds"`.

The sibling tools that share the pattern (`create_room`, `manage_room` with `member_bot_ids`) would benefit from the same treatment.

bradhallett:
## Summary

`coordinate_bots` documents snake_case arguments (`bot_ids`, `message`, `request_key`, plus optional `group_id`, `rework`, `label`), but the harness route it posts to is camelCase. A caller that reached for the camelCase spelling (`botIds`, `requestKey`, `groupId`) was refused with a generic validation error that named none of the expected fields, leaving nothing to self-correct from (#1239).

## Change

- `callTool`'s `coordinate_bots` branch maps the camelCase aliases onto the canonical snake_case keys before validating, so a call with either spelling works. When both spellings arrive, the documented snake_case key wins; the alias only fills a missing canonical key.
- An unusable call (missing or wrongly shaped `bot_ids`/`message`/`request_key`) is refused in the proxy with a usage error that names every expected field - required and optional - together with the keys the call actually carried. No request reaches the harness in that case.

## Alias decision

The aliases were kept deliberately narrow: only the three fields whose wire name differs from the tool schema, mapped in one place immediately before validation. The proxy does not grow a general camelCase convention and the advertised input schema is unchanged - snake_case remains the documented contract; the aliases only stop a reasonable call from failing on spelling.

## Testing

- a new `coordinate_bots arguments (room turn)` suite drives a room-turn proxy child end to end against a stub harness: camelCase aliases map to the canonical fields on the wire; when both spellings are present the snake_case values win for `bot_ids`, `request_key` and `group_id`; and an unusable call returns the teaching error (all six field names plus the received keys) without hitting the harness route
- full `server/drivers/agents-proxy.test.ts` (109 tests), `pnpm typecheck`, and `pnpm lint` pass

Fixes #1239.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the `coordinate_bots` tool to accept both camelCase and snake_case argument formats.
  - When both formats are provided, the documented snake_case values take precedence.
  - Added validation for bot IDs, messages, and request keys, with clearer errors for invalid input.


bradhallett:

---

macos leg: the failure is `server/team-setup.e2e.test.ts > Clive reviews multi-provider teams once, continues after each decision...`, asserting that a denied team-setup request still listed one proposed bot. This PR touches only `server/drivers/agents-proxy.ts` and its test; the team-setup request flow is untouched, and the same suite passed on the ubuntu leg of this run. It reads as a timing flake in the e2e denial race. I do not have rights to re-run jobs on this repo; I will retrigger it with the next push to this branch, or on request.

PR #1267, opened 2026-09-15
fix(android): open the thread picker on a multiple-threads send failure

### LINKED ISSUE #1132: Android companion app can't message a bot with multiple threads (no thread picker)
## Summary

Since #981 (independent bot threads, merged 2026-09-08), the server rejects any client message to a bot with more than one thread unless the client explicitly picks a thread:

```
server/index.ts, requirePinnedClientThread():
"This bot has multiple threads. Update this client and choose a thread before sending this action."
```

The Android companion app has no UI to select a thread, so it can never satisfy this — once a bot has 2+ threads, the Android app can no longer send it any message at all. It just shows the generic "Something went wrong" dialog with that error text.

## Why this matters

The multi-thread feature is genuinely useful (e.g. keeping a saved thread with search results/links you want to keep, separate from ongoing chat) but it directly breaks the Android app's core use case: messaging a bot on the go. Concretely — asking a health-assistant bot "when was my last appointment?" from your phone is now impossible the moment that bot has a second thread for any reason (including one created by a one-hop delegation from another bot, which happens without the user creating it directly).

Right now the only workaround is to either delete threads back down to one (destroying whatever was in the extra thread) or switch to desktop/Mac to message that bot, which defeats the point of having a phone companion app at all.

## Suggested fix

Add a thread picker to the Android companion app (even a minimal one — a dropdown/list of the bot's open threads) so it can pick a thread and satisfy the same 409 gate the desktop/web/Mac clients presumably already handle. Alternatively, allow the companion API to default to the bot's most-recently-active thread instead of hard-rejecting when no threadId is given, at least for simple one-off asks.

## Environment

- OpenMausBot Android companion app, paired to a Windows-hosted server (desktop-win-pc)
- Reproduced by triggering a one-hop `delegate_bot` task on a bot (Inga), which created a second thread for her, then trying to message her directly from Android — every message attempt fails with the 409 above.

buttonsjasper360-lang:
## Summary
- Fixes #1132 — sending a message from the Android app to a bot with more than one thread and no pinned thread failed with a generic "Something went wrong" dialog instead of surfacing the existing thread picker.
- Root cause: `server/index.ts`'s `requirePinnedClientThread()` returns 409 in this case, but the Android app's send paths never checked `session.actionError` for this specific error, so it fell through to the generic error dialog rather than opening `TaskSheet` (the thread picker already used elsewhere in the app).
- Both send paths in `ChatScreen.kt` now detect the multiple-threads error and open the thread picker; the plain-text path also restores the drafted message into the composer so picking a thread doesn't require retyping.

## Changes
- `Session.kt`: added `MULTIPLE_THREADS_MESSAGE_PREFIX` constant.
- `ChatPolicy.kt`: added pure predicate `isMultipleThreadsError(message: String?)`.
- `ChatScreen.kt`: both send paths open `TaskSheet` on this error instead of the generic dialog.
- `ChatPolicyTest.kt`: new `MultipleThreadsErrorTest`, 4 cases.

## Test plan
- [x] Full JVM suite green (`core` 530/530, `app` 898/898, 0 failures/errors)
- [x] Debug APK built and installed on a physical device (Samsung S26+)
- [x] Reproduced with a real 2-thread bot (Nightingale) — confirmed the thread picker opens on send and lets you pick/swap between threads instead of the generic error dialog

## Summary by CodeRabbit

* **Bug Fixes**
  * When a message cannot be sent because the destination has multiple threads, the app now opens the thread picker instead of showing a generic error.
  * Draft text is restored, and attachments remain available, so users can retry after selecting a thread.
  * Previous send errors are cleared after a successful retry, preventing outdated error messages from appearing.


buttonsjasper360-lang:
Pushed `e6852395` addressing the CodeRabbit finding above, with a more targeted fix than the suggested one.

Root cause: `send(text, to)` (`Session.kt:1065`, the plain-text path this PR added a check after) never cleared `actionError` on entry, unlike its attachment-carrying sibling `send(text, attachments, to)` a few lines below, which does. Its only caller is `ChatScreen.kt:701` — the new multiple-threads check added here — so a stale error from an earlier failed send on the same chat could be misread as this call's own result. Concretely: send fails with "multiple threads" → picker opens → user picks a thread → resend succeeds → but the stale error was still sitting in `actionError`, so the picker would reopen even though the message sent fine.

Fixed by clearing `actionError` at the top of `send(text, to)`, matching the existing pattern already used by `send(text, attachments, to)` in the same file — minimal, single call site affected, no change to `perform()`'s shared behavior used by the other ~30 call sites.

Added a regression test (`SessionP1Test.kt`: `plainTextSendClearsAStaleErrorLeftByAnEarlierFailedSend`) reproducing the exact sequence. Full JVM suite green.

buttonsjasper360-lang:
Closing this — the underlying bug (Android companion app can't message a bot with 2+ threads, #1132) is now fixed natively in the official Android 1.3.0 release (a different implementation than this PR: it sends `threadId` from the phone and adds a real thread switcher in the chat header, rather than intercepting the send failure). Tested live on-device: create/delete/start threads, message a bot in a new thread, works with no picker needed. This PR's approach is now redundant.

Issue #1280, opened 2026-09-15
Thread data retention: event logs grow unbounded; single-thread delete leaves them on disk


bradhallett:
## Problem

Thread event logs grow without bound. Nothing in the server applies any retention, TTL, cap, or sweep to threads, messages, or event logs — existing sweeps cover attachments, room-handoff node eviction, and cloud Box TTLs only. On one multi-bot install, three days of activity accumulated ~4.4 GB of per-thread logs (events/ 1.8 GB, native/ 2.6 GB, messages.db 61 MB) across 176 threads, while the thread metadata itself was 204 KB. The weight is event logs, not thread count.

## Bug: single-thread delete leaves event logs on disk

`DELETE /api/bots/:id/tasks/:threadId` → `Store.deleteTask` → `deleteThreadRecord` removes the messages.db rows and the legacy messages JSON, but never unlinks the thread's `events/<threadId>.ndjson` and `native/<threadId>.ndjson`. Group delete and bot delete both unlink those files (the `[EVENTS_DIR, NATIVE_DIR]` loop in each handler). Single-thread delete should match, or the logs a delete is meant to remove survive it.

## Proposals

1. **Fix the delete gap** above — small, bug-level.
2. **Retention sweep for event logs**: a server-side sweep that compacts or removes `events/` and `native/` NDJSON for threads closed or archived older than X days, reusing the attachments sweep pattern. Suggest X = 30 days, default off.
3. **Optional auto-archive**: archive closed threads after X days (30 suggested, default off), as a global default with a per-bot override, following the existing `threads:` / `rooms:` zod config pattern (`threads.maxConcurrentPerBot` today). Archive rather than delete because it is reversible and preserves resume cursors, last instance state, and cwd (#1194 context). Closed threads already fold by default (#1108), and archive landed with #1201 / #1241.

## Safety

A sweep must never touch threads that are busy, unread, snoozed with a pending wake (#1205), or carrying open handoffs. Auto-delete stays out of scope.

Citations: `Store.deleteTask` / `deleteThreadRecord` (server/store.ts); thread-delete, group-delete, and bot-delete handlers (server/index.ts); attachments sweep (server/attachments.ts); room-handoff eviction (server/room-handoffs.ts); Box TTL (server/box.ts); config schema (server/config.ts). Issues: #1108, #1194, #1201, #1205, #1241.

bradhallett:
Status roll-up for triage — two of the three asks here are fully landed, the third is in review:

[...]

PR #1284, opened 2026-09-15
fix(server): delete thread event logs with the thread

bradhallett:
Implements the delete-gap part of #1280.

## Problem

`DELETE /api/bots/:id/tasks/:threadId` -> `Store.deleteTask` -> `deleteThreadRecord` removed the messages.db rows and the legacy JSON, but left `events/<threadId>.ndjson` and `native/<threadId>.ndjson` on disk. Group delete and bot delete unlinked those files in their route handlers; single-thread delete was the gap, so the logs a delete is meant to remove survived it.

Bot delete had a second, quieter gap: its handler only unlinked the *active* thread's files, so the event logs of every non-active task thread were orphaned when a bot was deleted.

## Change

- `Store.deleteThreadRecord` now unlinks both NDJSON files alongside the transcript rows. Task, group, and bot deletion all funnel through it, so a thread's logs cannot outlive it on any delete path.
- `deleteGroupTask` also funnels through `deleteThreadRecord`, so deleting a group's task threads is covered by the same cleanup.
- The now-redundant handler loops in the bot-delete and group-delete routes are removed.
- The refuse-to-delete-running behavior is untouched: the route still 409s on a busy or routine-active thread before reaching the store.

## Tests

- New cases in `server/tasks.test.ts`: deleting a task removes both NDJSON files while a surviving sibling task keeps its own; deleting a bot removes every member thread's logs (active thread plus tasks).
- Store-side delete-path suites pass (`tasks`, `group-tasks`, `independent-task-store`, `store`, `delegations`, `memory-store`), plus typecheck and lint.

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleting tasks, bots, or groups now consistently removes their associated event history while preserving logs for remaining threads.
  - Improved cleanup of conversation data during resource deletion, preventing stale event history from persisting after deletion.


bradhallett:
CI note for 649bb40f: all three typecheck+test legs are red, and every failing test comes from files this branch does not touch (run 35034086740; macos job 104599052346, windows job 104599052379, ubuntu job 104599052429). Everything else on this head is green: typecheck and lint pass on all three legs, and the Swift, Kotlin, control-plane, renderer smoke, package+smoke, open-source build, and cloudflared checks all pass. This branch's own suites (tasks/store/message-db) also pass locally, 124/124.

Each failure is carried by main, matching main's own CI (run 35034070097) failure for failure:

- All three legs: src/components/BotProfileAvatarCard.test.ts fails with ReferenceError: window is not defined through useOrganizationBranding (src/lib/use-organization-branding.ts:7, from #1315). Main fails the same test on all three of its legs (macos job 104603474028, ubuntu job 104603474021, windows job 104603473957).
- Ubuntu also fails server/index.test.ts:4441, "team import is additive-only: smuggled grants, claimed ids, and re-imports never touch existing records" (expected [ ...(2) ] to deeply equal [ Array(1) ]). Main's ubuntu leg fails the same assertion at the same line. The test passes locally on both this head and plain 0101f1c2.
- Windows also fails server/box-delete-journal.test.ts with expected 438 to be 384: the test asserts POSIX mode 0o600 (384) through statSync, and Windows reports 0o666 (438). Main's windows leg fails the same assertion. The file passes 9/9 locally.

This branch deletes thread event logs with the thread (server store code) and touches none of the failing files, so no tree on current main can produce a green typecheck+test leg right now.

PR #1286, opened 2026-09-15
feat(server): optional auto-archive for long-closed threads

bradhallett:
Implements the optional auto-archive part of #1280.

## Problem

Closed threads fold out of the default sidebar but stay in "all threads" forever; nothing ever archives them, so the archived state the sidebar already understands (#1201 / #1241) is only ever reached by hand.

## Change

- Global setting `threads.autoArchiveDays` (zod, 1..3650, **absent = off, the default**; 30 days is the suggested value), following the existing `threads:` config pattern.
- Per-bot override `autoArchiveDays` on the broad bot PATCH (`PATCH /api/bots/:id`): absent inherits the global window, **0 opts this bot out**, null clears back to inherit. It deliberately stays out of `parseBotProfilePatch` so the paired-client profile boundary stays narrow.
- New `server/thread-auto-archive.ts`: a pure window resolver plus selector. A thread is selected only when it has been closed (`closedBy.at`) **strictly longer** than its effective window, is not already archived (idempotent), and is not busy, unread, or carrying an open direct handoff.
- Wired in `server/index.ts`: runs at boot and daily (unref'd timer); archives via the existing `store.patchTask(..., { archivedAt })` path. Archive, not delete — it is reversible and keeps resume cursors, last instance state, cwd, and handoff state (#1194).

## Scope notes

- Group (room) tasks carry no close stamp in the data model today, so auto-archive covers bot threads; rooms join once they grow one.
- A snooze-with-pending-wake exclusion (`#1205`) joins naturally if that PR lands.
- No dedicated UI yet: the setting is the config file plus the bot PATCH field.

## Tests

- New `server/thread-auto-archive.test.ts`: window-resolution precedence (inherit / override / 0-opt-out / opt-in-when-global-off), selection (past-window picked, recent and unclosed kept, exact-cutoff kept, busy/unread/handoff excluded, never re-archived), and a store-level case proving the archive rides `patchTask` while `resumeCursors`, `lastInstanceId`, `cwd`, and `closedBy` survive.
- Extended `server/config.test.ts`: accessor null by default, valid value round-trips, schema rejects 0 / negatives / non-integers / 3660 / non-numbers.
- Gates: new + config suites 101 green; `tasks`/`store`/`bot-profile` suites 115 green; typecheck, lint, and a real-server boot smoke all clean.

## Summary by CodeRabbit

- **New Features**
  - Added optional automatic archiving for closed threads after a configurable number of days, with global and per-bot settings, inheritance, and opt-out support.
  - Archives only eligible threads and runs at startup and every 24 hours.
  - Added per-thread computer or browser surface selection.
  - Added model variant selection with validation and effort compatibility checks.
  - Added support for URL-based MCP servers with headers.
- **Bug Fixes**
  - Prevented surface changes while a thread is busy.

## Rebase and CI hardening

Rebased onto current main (the branch previously sat 4 drift commits behind with merge bubbles). It now resolves against main's newer `threads` config surface (event-log retention fields coexist with `autoArchiveDays`) and declares the per-bot override on `WireBot` so the store's wire-projection exactness check holds. It also carries, verbatim from the stack, the bounded self-recovery retry window (#1516 / #1537), the Electron relaunch session-bus fix (#1545 / #1518), and the iOS preview fixture seed (#1546 / #1519) so this standalone branch is deterministic against the base flakes its CI hits.


bradhallett:
CI note: the typecheck+test legs on this head are expected to fail one main-carried test, `server/mcp-selection.e2e.test.ts` ("imports atomically…", `expected 201 to be 400`). #1297 (d67fc3a8) made http(s) URL MCP servers importable without updating that old probe, so every run composed against current main fails it — main's own run [35021469154](https://github.com/milind-soni/OpenMausBot/actions/runs/35021469154) shows the same failure on all three OS legs. One-line fix proposed in #1303; nothing in this branch touches it.

bradhallett:
CI note for 67d05d3f: only the macOS typecheck+test leg is red, failing server/tts/tts.test.ts > Fish Audio > "lists only trained TTS voices using Fish model ids". That is byte-for-byte main's own current macOS failure (main run 35028474977, job 104587937305, same test, same deep-equal assertion), introduced by #1301. Ubuntu and Windows are fully green on this head, and this branch touches no TTS code.

bradhallett:
CI note: on run 35112076566 the only failing leg is macos at Verify native approval mode transitions (smoke-approval-modes.cjs:228, list_bots reads 3 vs expected 2), identical to upstream main's own CI run 35108771265 at b94f4a61 where ubuntu and windows pass; every other check is green. Main-carried, no PR change owed. Fork Vercel red is the known deploy authorization noise.

bradhallett:
Updated this branch from `main` (merge fb0abc6c). Both conflicts keep both sides: main's deterministic browser-runtime test fixture, and `autoArchiveDays` alongside main's new `context` config block. CodeRabbit follow-ups are in. Could you re-review when CI is green?

bradhallett:
CI note for run 35627345666: the windows-latest shard 1/4 failure is a single test — `server/context-compaction.e2e.test.ts > respects the selected account's native compaction setting ('off')` — which died on `fetch failed → ECONNRESET` (socket reset, no assertion). The file's other six tests passed, including the sibling `'undefined'` / `'100000'` / `'auto'` variants, the same file is green on ubuntu shard 1/4, and none of this PR's changed files touch the compaction path. For contrast, the delegation failure main hit on this shard at e87f4880 (`server/independent-threads-api.test.ts`) passes here — this branch carries the fix from #1613 (8b7333c1, cherry-picked as 06d8ee39).

Calling it a runner/network flake. I don't have rerun rights on this repo, so could a maintainer re-run the failed jobs? (macOS shards 1–3 are still queued behind runner availability.)

PR #1289, opened 2026-09-15
fix(desktop): keep the chat usable at every window size, and fold the header into one menu

aivsomkar:
## Summary

Desktop sizing pass. The app opens at 1100×780 and can shrink to 840×620, and until now several parts of the chat column broke as soon as a side panel took space or the window got small. Every break had the same shape: a flex row or column whose neighbours could not shrink, so the one flexible element got crushed. Verified over CDP screenshots on a dev instance at 1100, 974 and 840 wide with the settings panel, the inspector, the computer panel and both open, then on an OMB2 side-by-side build.

### Composer
The row (attach, Full access, Auto, editor, mic) let only the editor shrink, so it collapsed to a sliver and its placeholder stacked one letter per line while auto-grow made the box tall. Below a 30rem composer width the row wraps: editor on its own line, chips and mic beneath. Container query on the composer box; `MentionTextarea` gains `wrapperClassName`.

### Chat and group headers
The chip group is `shrink-0`, so beside a panel the name truncated to nothing and the rename pencil landed under the find button. Below 30rem the header wraps into a name line and a chip line. The query lives on a child row: a container query never matches the container itself (cost one round).

### Header controls, simplified (asked for during review)
Threads live in the sidebar now, so the thread chip is gone. Find (⌘F), **Share** → "Copy as Markdown" / "Download as .md", Usage (with the cost as its trailing figure) and Inspector fold into one "more" menu that opens on hover; Model, Call and Computer stay as buttons. Four controls instead of eight. `SidebarPopoverMenu` gains `placement="below"` and a `heading` item field.

### Sidebar beside a panel
Bot settings deliberately keep the inspector or computer panel open (their own controls open settings), and the full sidebar plus a panel left the default window a ~330px chat. While a side panel is open and the window is narrower than 1280px, the sidebar folds to its avatar rail (`collapseToIcons`); the saved density is untouched and returns when the panel closes. New `useMediaQuery` hook.

### Panels
- Beside the chat from 768px up instead of covering the whole window below 1024px ("why does the whole thing hide behind this?"). With the rail that leaves a 350px chat at the 840px minimum.
- Two panels at once: until the window is 1536px wide, settings floats over the other panel as a sheet, and that panel is still there when the sheet closes. The global `:focus-visible` ring is suppressed on the sheet container.

### Also
Composer chip group and header control group may wrap instead of overflowing.

## Platforms
|---|---|---|
| macOS     | yes | in this PR; verified on screenshots at 1100/974/840 and on an OMB2 build |
| Windows   | yes | in this PR (renderer code, no platform gate); NOT built on Windows — the wrapped header keeps `controlsShiftStyle` for the caption buttons, but the second chip line under a frameless title bar is unverified |
| iOS       | no  | n/a: desktop renderer only |
| Android   | no  | n/a: desktop renderer only |
| Companion | no  | n/a: no route changed |

## Test plan

- [x] `vitest run` on ChatView.controls, GroupView, Sidebar (simple-mode, i18n), SidebarPopoverMenu, SidebarMoreMenu, ComposerQueuedMessages, EnginesSettings — all passing; new cases pin the composer and header wrap classes
- [x] `tsc -b` clean; `node scripts/generate-locale.mjs --check` valid
- [x] Renderer build: compiled CSS contains the `@container composer` and `chathead` rules with their utilities
- [ ] Windows: open a bot's settings at the default window size and check the header's second line clears the caption buttons

## Summary by CodeRabbit

* **New Features**
  * Chat and room headers now adapt more cleanly to narrow screens, wrapping controls when needed.
  * Added a consolidated menu for chat actions, including transcript copying, Markdown download, and usage details.
  * Sidebars and settings panels automatically adjust between docked, collapsed, and overlay layouts based on available space.
  * Sidebar menus support labeled items and configurable popup placement.

* **Bug Fixes**
  * Improved responsive layout for the message composer and inspector panel across medium-sized screens.


weskcode:
I ran into the same overlap in the desktop app and reproduced it on main at 4e190733 with the isolated fixture (`control-omb ui launch`), at the default 1100×780 window:

- With the inspector open, the chat column is about 320px wide. The header control row is `shrink-0`, so the find button sits on top of the avatar button and the avatar can't be clicked.
- With bot settings and the inspector both open, the chat column is 0px wide. The header controls and composer paint over the settings panel, and the composer textarea is 8px wide.

This PR looks like it fixes both. Are you still planning to rebase it? If a smaller first PR would be easier to review, I can help split the header and composer wrapping away from the sidebar and panel changes.

milind-soni:
Holding this redesign out of the current integration batch. The current main branch supports Icons-only, closed-sidebar, and Show threads-off modes; this patch removes the header TaskPicker while assuming the sidebar always exposes the histories. Please retain the accessible history/activity route in those modes and split the composer wrapping fix from the header/panel redesign. A focused renderer pass should cover narrow widths, sidebar hidden/icons, keyboard access, and Windows caption controls before merging the broader layout changes.

PR #1288, opened 2026-09-15
feat(board): phase 2 part 1 — the task object: owner, due date, a money cap that pauses, results as digests

aivsomkar:
Phase 2, part 1 of the harness programme (`docs/plans/2026-09-15-phase-2.md`). Stacked on #1281 (Phase 1). Spec: `docs/superpowers/specs/2026-09-15-phase-2-task-object-design.md`.

## In plain terms

The durable task board from #1185 (a task that survives a restart, is claimed by a bot, heartbeats while it runs, and is reclaimed when its bot dies; off behind `features.board`) is brought onto the chain unchanged, and the task row learns four things a person needs to read a board and a bot needs to be stopped by:

- **owner** (a bot id or `person`) and **due date**;
- **a money cap** (decision 11): the harness books every settled turn's cost from the usage row it already writes, posts one comment at 70%, and at the cap moves the task to `blocked` with *paused, needs a budget increase*. A paused task is never dispatched again (no attempt is spent). Only a person raising the cap through `PATCH /api/tasks/:id` resumes it; a bot may set a cap when it files a task but cannot raise one. `board.defaultBudgetUsd` caps unattended work filed without a cap. A turn that reports no price books nothing and is counted (#59's rule).
- **the result is the turn's digest**: what was done (tools, files, memory, reply), not only what was said.

`task_create` accepts `owner`, `due_at`, `budget_usd`; `task_list` shows owner, due date, spent-of-cap and the result.

## Measured

End to end on the fake Claude engine (`server/task-budget.e2e.test.ts`, real HTTP harness, real dispatcher ticks): a task capped at $0.014 runs once and sits at 70% with the warning; runs again and pauses at $0.02 with the stated reason and comment; a cap below the spend is refused with 400; a raised cap resumes it and a third turn runs; the result carries the digest line; the ledger shows three priced turns. The scorecard T1–T10 and the recall set are unchanged by construction: the board is off by default and nothing on the turn path changed.

## Every engine

Harness-side, full on every engine: spend comes from the usage row every driver produces; a driver that reports no cost books nothing and the row says so. Digests exist on every engine since Phase 0.

## PRs checked (AGENTS.md rule 2)

- **#1185 (open, this account):** the board core, imported as is (twelve commits plus two follow-ups; the only changes are the conflict resolutions against today's main, keeping both feature gates). Its rule that completion is observed by the harness, never self-reported, is what the budget follows. #1185 should be closed in favour of this PR once the chain merges.
[...]

PR #1293, opened 2026-09-15
feat(harness): phase 2 part 2 — bounded tools that spill, and an engine fallback ladder

aivsomkar:
Phase 2, part 2 of the harness programme (`docs/plans/2026-09-15-phase-2.md`). Stacked on #1288 (part 1). Spec: `docs/superpowers/specs/2026-09-15-phase-2-reliability-design.md`.

## In plain terms

**Tools that fail honestly.** Every call a harness tool makes is bounded at 60 seconds; a read that fails on the network is tried once more; an error tells the model which tool, how long, whether it was retried, and what to do next, instead of a bare "fetch failed". A result over 24,000 characters is cut to its first 16,000 and the whole text is kept by the harness (redacted, 0600, per conversation) and read back in slices with `tool_result_read`, so a long listing no longer floods the context or gets lost.

**Engines that fall back (decision 12).** A bot may name an alternate engine (`fallback.alternate` on `PATCH /api/bots/:id`). When a direct turn fails on a terminal provider error (quota, auth, overloaded, rate limits past the driver's own retries, server errors) the harness switches the task to the alternate, records `task.fallback = { from, to, reason, at }`, says so in the thread ("continuing on Grok after quota on claude"), and re-sends the exact message that failed once the failed turn has settled; the fresh-session replay carries the history. Two things stop the silent rung and leave it to the person: the failed turn had a side effect, or there is no alternate. Then, for a quota error, the card from #1184 (imported here) offers the other engines; otherwise the thread shows the error as today. A task falls back silently at most once.

## Measured

- `server/fallback.e2e.test.ts` (real harness, fake Claude failing the way an out-of-quota CLI does, fake ACP as the alternate): the bot with an alternate gets its reply from the alternate, the task carries the record, the chip is there, the ledger's row names the alternate's driver; the bot without one gets the card and no reply.
- Proxy suite: the 60 s bound and retry rule, the teaching errors, the cap, the spill and the slices.
- T1–T10 and the recall set unchanged by construction: no scorecard bot names an alternate and the bounds do not fire on a healthy harness.

## Every engine

Full on every engine: every engine mounts the same proxy, and the alternate can be any engine the workspace has; the failed message is re-dispatched with the same fresh-session replay an engine switch uses.

## PRs checked (AGENTS.md rule 2)

- **#1184 (open, this account):** the quota card is decision 12's "approval before degrading"; its two card commits are imported, its replay-tool-history commits are not (Phase 0's digests carry that). #1184 should close when this merges.
- **#422 (open, external):** an output bound on the browser tool; this cap is on the harness's own tools and keeps the whole result readable.
- **#59 (open, external):** per-task wall-clock and retry budgets; these bounds are per call, in the proxy; they compose.

## Platforms

| Platform | Impact | Verified |
|---|---|---|
| macOS | server (proxy, fold, routes); the card renders through the existing option card | suites, e2e, typecheck, lint |
| Windows | server; POSIX-gated e2e like `branching.test.ts` | CI |
| iOS / Android | none | — |

## Checks

`pnpm typecheck`, `pnpm lint`, `pnpm i18n:check` clean; the suites the wiring touches green (the two that failed in one batch run pass alone and in the full suite: a shared temp folder, not the code). Full suite on the Phase 2 tip: see #1294 (parts 3–4), which carries the run for the whole branch.


milind-soni:

PR #1299, opened 2026-09-15
fix(docs): cite browser-live tests with their server/ paths

bradhallett:
## What

The chat-to-action acceptance section added in #1296 (23581b4e) cites `group-local-vm.e2e.test.ts`, `vps-routing.test.ts` and `index.test.ts` as bare filenames in `docs/verification/browser-live.md`. Those files live under `server/`, and `scripts/testing/verification-docs.test.ts` joins every cited test path at the repo root, so the "cite test files that exist" check now reports all three as missing.

## Impact

Every `typecheck + test` matrix run started after that merge fails on this one doc check, on all three OS legs, regardless of the PR's own changes (any merge against current main reproduces it; the doc and the test are identical on main and in the merges). This one-line prefix restores the check.

## Verification

- Before: `vitest run scripts/testing/verification-docs.test.ts` fails 1/7 with the three citations listed as missing.
- After: 7/7 pass. One line changed, no code touched.

## Summary by CodeRabbit

- **Documentation**
  - Updated browser live verification instructions to use the correct test paths.


bradhallett:
Closing: #1298 landed this same citation fix on main (4174e028) while this PR was in flight. Verified the verification-docs suite passes at current main, so no further change is needed here.

milind-soni:
Haha youre fast!

PR #1306, opened 2026-09-15
fix(test): give the MCP import test an address that is actually invalid

milind-soni:
**Third red on main**, found by running the full suite locally during release pre-flight. CI kept getting cancelled by rapid merges, so this one never surfaced there.

`server/mcp-selection.e2e.test.ts` → `expected 201 to be 400`.

## Cause

**#1297** added remote MCP servers by URL. The entry this test used as its *bad* one — `{ url: "https://invalid.example/mcp" }` — is now a perfectly valid remote server. Unreachable is not the same as invalid; the address is well-formed https.

So the batch import succeeded, and the test's real subject — **one invalid entry rejects the whole batch and writes nothing** — stopped being exercised.

## Fix

Swap in an address that genuinely cannot parse. The validator itself is correct and unchanged: `urlError` requires a parseable `http(s)` URL ([mcp-registry.ts:73-80](server/mcp-registry.ts#L73-L80)).

## Verification

Mutation-checked the assertion rather than just the pass: making `urlError` accept any address reproduces `expected 201 to be 400`; restored, 2/2 pass.

One honest limitation: this case does **not** pin the `http(s)`-protocol rule. An unparseable address fails at `new URL()` first, so deleting the protocol check alone stays invisible here — I verified that directly. Worth a separate case if you want that rule guarded.

typecheck and lint clean.

---

This is the third merged PR in a day to leave main red on a required check (#1296 docs, #1297 features + this, #1269 SSH paths). None of them can be caught by the bump PR, since `prepare-release.yml` creates it with `GITHUB_TOKEN` and GitHub fires no workflows for it.

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end coverage to verify that an unparseable server address causes the entire import batch to be rejected with a 400 response.
  * Clarified test coverage distinguishing invalid addresses from valid HTTPS remote servers.


milind-soni:
Superseded by #1303, which landed the same fix first — and a better one.

Both swap the batch-import probe for a genuinely invalid address. #1303 uses `ftp://invalid.example/mcp`, which trips the **protocol** check; mine used `"not a url"`, which fails earlier at `new URL()`.

That difference matters, and it's the exact limitation I flagged in this PR's own description. Mutation-checked against current main just now:

- delete the `http(s)` protocol check → main's `ftp://` fixture **fails** (`expected 201 to be 400`) ✅
- same mutation against my `"not a url"` fixture → **passes**, the regression stays invisible ❌

So #1303's version pins a rule mine left unguarded. Closing rather than resolving the conflict, since merging this would trade coverage down.

Main passes `mcp-selection.e2e.test.ts` 2/2, so the third release blocker is cleared.

Issue #1305, opened 2026-09-15
Routines silently defer for hours when the target bot is busy


bradhallett:
## Problem

Bot-target routines queue on schedule but only dispatch when the target bot is not busy. Busy includes any thread in working, waiting-on-you, or no-signal (`ACTIVITY_BUSY` in `refreshBotActivity`), so a coordinator bot holding one long working turn silently defers its routines for hours.

Observed on 0.1.80 (2026-09-15):

- A routine scheduled for 08:00 queued at 08:00:01 and only dispatched at 17:06 — 9h06m late. The target bot had held a working turn continuously from 07:52 to 17:06, back-to-back across two threads.
- A second routine scheduled for 09:00 queued at 09:00:02 and dispatched at 17:06 the same day.
- The day before, a weekly review ran 45 minutes late the same way.

During the whole wait:

- The `routine.run` message in the source thread is created as queued at schedule time and only patched again at finish (`syncRoutineRunToSource`), so the conversation shows nothing unusual for hours.
- Deferral is unbounded. Once the scheduled time is more than `CATCH_UP_MS` (12 hours) in the past, the run is instead marked missed and never executes — also silently.

## Proposal

1. Surface deferral in the run state: while a queued run is being skipped because its target is busy, show a "deferred: target busy" state on the routine run card and in the source-thread message, instead of leaving it looking freshly queued.
2. Raise one user notification when a scheduled run has been deferred beyond a threshold (30 minutes seems a reasonable default), so an unattended user learns the schedule is parked rather than silently late.

## Open questions

- Should waiting-on-you count as busy for routine dispatch? A bot blocked on user input arguably has room to start a detached routine task.
- Should deferred runs queue onto the bot the way user messages do (dispatch as soon as the current turn ends), rather than waiting for a scheduler tick to find the whole bot idle?

A small PR follows implementing the surfacing and notification only — no dispatch-semantics change.

bradhallett:
Closing as fixed by #1309 and #1318.

#1309 delivered the surfacing half: a queued run behind a busy target shows "deferred: target busy" on the run card and in the source-thread message, with one restart-safe deferral notice after 30 minutes.

#1318 delivered the semantics change: routine and webhook dispatch now admits by thread-slot capacity the same way startTurn does, so a bot with a free slot runs scheduled work immediately instead of waiting to be fully idle, with active group turns staying exclusive. That also answers both open questions from the filing in code.

PR #1309, opened 2026-09-15
feat(routines): surface busy-target deferral on queued runs

bradhallett:
Refs #1305 — implements the surfacing part only; dispatch semantics are deliberately unchanged (open questions in the issue need maintainer input).

## Problem

A routine queued behind a busy target looks freshly queued for as long as the deferral lasts, which can be hours (live evidence in #1305). The `routine.run` source message is written as queued at schedule time and only patched again at finish, so nothing surfaces the wait, and deferral is unbounded.

## Change

- `RoutineRun` gains `deferredAt` (first busy-skip tick) and `deferredNoticeAt` (notice raised). Both persist with the run, so a restart neither loses the wait nor re-notifies. Status stays `queued`; the lifecycle is unchanged and the run dispatches normally the moment the target frees.
- The run card and source-thread fallback text read "deferred: target busy" while a queued run is deferred (`routineRunCard` passes `deferredAt` only while queued; `RoutineRunCard` renders the label).
- One `routine-deferred` notification (new `NotifyKind`, title "<bot>'s routine is waiting") fires when the deferral passes `ROUTINE_DEFERRAL_NOTICE_MS` (30 minutes), once per run, routed to the routine's source thread the same way `routine-failed` is.

Deliberately out of scope: whether waiting-on-you should block dispatch, and whether deferred runs should queue onto the bot like user messages — both raised in #1305.

## Tests

- `server/routines.test.ts`: deferral stamp on the first busy skip (durable across reload, published through run-changed), no notice before 30 minutes, exactly one notice after, quiet when the target frees early, promptly dispatched runs never stamped, deferred run still dispatches when the target frees; the destination-rollback tests were updated for the one extra publish.
- `src/components/RoutineRunCard.test.ts`: deferred label for queued + `deferredAt`, plain Queued otherwise.
- `server/notify.test.ts`: title for the new kind.
- Gates: `pnpm typecheck` and `pnpm lint` clean; targeted vitest suites green on the rebased head (routines 113, notify 10, RoutineRunCard 12, index routine-filtered 7).

## Pre-existing failures on main

Two CI notes, both reproducible on clean `origin/main` (`e7299e1d`) and unrelated to this change:

- The fork's Vercel check is red for every fork PR (systemic to this repo's fork setup); GitHub Actions is the meaningful signal.
- `server/index.test.ts` has one failing test ("reconciles a committed crash-stale bot reference before ACK and profile-id reuse") that fails identically in isolation on clean main — browser-cleanups reconciliation, no routine involvement.

## Summary by CodeRabbit

* **New Features**
  * Queued routines now indicate when they are deferred because the target is busy.
  * Deferred routines send a notification after extended waiting, helping explain why execution has not started.
  * The desktop app can now keep the computer awake while a routine is running or due soon.

* **Bug Fixes**
  * Deferred routines remain queued until their target becomes available, then dispatch normally.
  * Repeated deferral notifications are prevented for the same routine run.


bradhallett:
CI note: the macos-latest failure is not from this patch.

The leg fails at scripts/smoke-approval-modes.cjs:228 with the assertion Repeated reads complete without another prompt (3 !== 2), after the full vitest run passes (552 files passed, 15 skipped).

PR #1303, which only changes an MCP selection probe, hit the byte-identical assertion on its macos-latest leg at 22:02 UTC tonight, also with a fully green suite (run 35026184125). The failure tracks the time window, not the diff: this patch touches server routine and notify wiring plus one card component, nothing in the approval or read-prompt path.

Matrix at posting: ubuntu-latest green, windows-latest still running. Happy to retrigger macos if that helps.

PR #1324, opened 2026-09-16
fix(electron): stale Windows data-dir lease after same-boot PID reuse

### LINKED ISSUE #1229: Windows: stale data-dir lease permanently blocks startup after PID reuse
## Summary

On Windows, a stale `openmausbot-server.lease` in `<dataDir>/.openmausbot-server-child/` permanently prevents the app from starting if the recorded PID has been recycled by an unrelated process. Every launch exits silently (code 0) — or shows "previous server process <pid> is still shutting down" — even though no OpenMausBot process is running.

## Root cause

`assertNoLiveDelegatedChild` → `ownerIsAlive(child)` → `!ownerPredatesThisBoot(child) && processIsAlive(child.pid)`.

On Windows `readBootSession()` returns `null` (no `/proc` boot_id, no macOS `kern.bootsessionuuid`), so `ownerPredatesThisBoot` falls back to the `uptime` field, which only detects owners from *before* the current boot. It cannot detect **PID reuse within the same boot session**.

`processIsAlive` then calls `process.kill(pid, 0)`, which returns true because Windows reassigned that PID to another running process (in my case `clipdiary.exe` grabbed PID 28152 seconds after the server child died).

Result: the lease is considered live forever → `DataDirLeaseError` → app quits. The user has to manually delete the lease file to recover.

## How I hit it

1. Auto-update to 0.1.78 ran the silent installer twice (log: `Install on explicit quitAndInstall` at 15:28Z and again at 15:50Z — the pending `update-info.json` apparently wasn't cleared after the first install, possible second bug).
2. Second install killed the desktop while the server child was still up → `server child did not stop before desktop exit; retaining the data-directory lease`.
3. The child died at 15:50:22Z. PID 28152 was immediately recycled by an unrelated process.
4. All subsequent launches died instantly: lease owner PID "alive" → "still shutting down".

I also accumulated 19 `openmausbot-server.lease.reap-*` files, suggesting the child dies without releasing fairly often on this machine — so the exposure window is real.

## Environment

- OpenMausBot 0.1.78, Windows 11, installed per-user under `%LOCALAPPDATA%\Programs\openmausbot`

## Suggested fixes

Any of:

1. Verify owner *identity*, not just liveness: record the owning process's start time (e.g., via `NtQueryInformationProcess`/`wmic process get CreationDate`/PowerShell `Get-Process .StartTime`) or image path, and compare on read. A recycled PID fails the check → normal stale-lease recovery runs.
2. On Windows, use a named mutex held by the child instead of (or in addition to) PID-liveness.
3. Ship a boot-session equivalent on Windows — e.g., `GetTickCount64`-based boot timestamp from `NtQuerySystemInformation(SystemTimeOfDayInformation)` or the `System Up Time` perf counter, stored in the lease — so a PID from a different boot is ignored even when alive.
4. At minimum: bound the "still shutting down" grace period (lease `createdAt` older than N minutes + liveness → treat as stale and recover, since a delegated child should never outlive its desktop parent for long).

## Workaround for affected users

```
del "%USERPROFILE%\.openmausbot\.openmausbot-server-child\openmausbot-server.lease*"
del "%USERPROFILE%\.openmausbot\openmausbot-server.lease"
```

then relaunch.

santhiprakash:
## Problem

On Windows, a stale `openmausbot-server.lease` can record a PID that has been reused by an unrelated process within the same boot session. `process.kill(pid, 0)` reports the recycled PID as alive, so `acquireDataDirLease` treats the lease as live and refuses to start, leaving the user stuck until they manually delete the lease.

## Root cause

`processIsAlive` only signals whether *any* process holds the PID. `ownerPredatesThisBoot` falls back to `uptime`, which only detects owners from before the current boot. PR #937 added boot identity for Linux/macOS but explicitly did not solve same-boot PID reuse on Windows.

## Fix

In `electron/data-dir-lease.mjs`, `ownerIsAlive` now also calls `ownerIdentityMatches` on Windows. That helper queries the process creation time via `wmic process where ProcessId=<pid> get CreationDate` and compares it with the lease's `createdAt`. The original process was created before it wrote the lease, so a process created after the lease cannot be the original owner.

If `wmic` is unavailable or its output cannot be parsed, the check conservatively returns `true` and the lease remains live (the previous, safer behavior). The change is Windows-only and does not alter the lease record shape or Linux/macOS behavior.

## Verification

- `pnpm typecheck` passes
- `node --test electron/data-dir-lease.node-test.mjs` passes (19/19 passing on this host; the new Windows-only test is skipped on Linux)
- `pnpm exec vitest run server/data-dir-lease.test.ts` passes (5/5)
- `pnpm exec oxlint electron/data-dir-lease.mjs` passes (0 warnings, 0 errors)
- `pnpm check:electron` passes (118 modules syntax-checked)
- `pnpm test:electron` passes (350 tests, 0 failures, 5 skipped)
- A simulated Windows environment (faked `wmic` + `process.platform = \"win32\")` confirms the pre-fix code refuses to acquire a live-but-unrelated PID, while the fixed code correctly retires it.

## Notes / risks

- This uses `wmic`, which Microsoft is deprecating. It is still present on all supported Windows versions. If a maintainer prefers a PowerShell/`Get-CimInstance` fallback, happy to add it.
- The new unit test is gated by `process.platform === \"win32\"` because it needs real `wmic` output. It will exercise the fix on Windows CI.
- Older lease records without creation-time data continue to work: missing/unreadable identity data falls back to treating the owner as live.

Fixes #1229.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows lease ownership detection by validating process identity and creation time.
  * Reduced the likelihood of stale locks blocking access to application data directories.
  * Added safe fallback behavior when process creation details cannot be retrieved.
* **Tests**
  * Added regression coverage for reused process IDs and unavailable Windows system utilities.


milind-soni:
Thanks for taking #1229 on — the direction (owner identity beyond the PID) is right, and the change is safe in the conservative direction. Two things to change before it can land:

1. **`wmic` isn't on the OS where #1229 was reported.** It was removed from Windows 11 24H2/25H2 and is absent on the `windows-2025` CI runner, so `processCreationTimeMs` returns null → owner treated alive → the bug persists exactly where it was seen, and the new test skips there (no coverage). Please query creation time through something that exists on Windows 11 24H2+/Server 2025 — e.g. PowerShell `Get-CimInstance Win32_Process -Filter "ProcessId=<pid>" | Select CreationDate`, or `(Get-Process -Id <pid>).StartTime.ToFileTimeUtc()` — keeping the conservative null→alive fallback. [electron/data-dir-lease.mjs:193](https://github.com/milind-soni/OpenMausBot/blob/HEAD/electron/data-dir-lease.mjs#L193), :224.
2. **`created < owner.createdAt` is a wall-clock ordering, which this module's own contract forbids** (lines 39-41: verdicts must be exact, nothing derived from the wall clock, because a wrong "stale" lets two instances share one data directory). A backward clock step between process start and lease write — w32time at boot, a VM snapshot restore — marks a *live* owner stale. Please make the identity exact: record the owner's creation time in the lease when it's written and compare equality with what the live process reports. [electron/data-dir-lease.mjs:225](https://github.com/milind-soni/OpenMausBot/blob/HEAD/electron/data-dir-lease.mjs#L225).

Smaller: the two synchronous `wmic` calls (5 s timeout each) on every Windows server-child start are a startup-latency regression on machines that still have wmic (:402, :414); a single query, or an async path, would avoid that.

PR #1325, opened 2026-09-16
feat(board): a cap nobody has to guess, and a pause that is one tap (phase 2 budget defaults)

aivsomkar:
Follow-up to #1288 (the task object), from Omkar's objection that a money cap a person must keep raising is a roadblock and a cap chosen blind is the wrong shape. Stacked on #1294. Addendum in `docs/superpowers/specs/2026-09-15-phase-2-task-object-design.md`.

## In plain terms

- **Nobody types a number.** A task filed without a cap gets one at dispatch from its bot's own history: three times the median spend of that bot's finished tasks, never under $1. A workspace default (`board.defaultBudgetUsd`) still wins when set. Chats you are watching are still never capped; the monthly cap stays the safety net.
- **A pause is one tap.** When a task pauses at its cap, its run thread gets one card with three answers: *Allow $X more* (X is the cap, never under a dollar), *Finish without a cap*, *Stop this task*. Raising, removing or stopping is one tap on the card; no form.
- **Warn late, pause rarely.** The 70% warning stays as the thing a person reads; with the history-based cap sitting well above a typical task, pausing means something went wrong.

## Measured

End to end on the fake engine (`server/task-budget.e2e.test.ts`): the pause raises the card in the run thread with the three labels; answering "Allow $1.00 more" through `POST /api/threads/:id/respond` moves the task back to `ready` at $1.02; a task filed with no cap runs with a $1 cap taken from history (the bot's finished tasks cost $0.01, so the floor applies). Unit: the median with an outlier (0.4, 0.5, 0.6, 9 → $1.65), per-bot isolation, the floor; the card's three outcomes and its once-per-task rule.

## PRs checked

#1184's option card shape (`fixedOptions`) is reused; #59's rule (never enforce from a guess) holds: the automatic cap is computed from booked spend only.

## Found by hand (2026-09-16, first in-app test)

Two defects Omkar's first test surfaced, fixed here with tests:

- **The board's tools vanished under deferred loading** (part 3 default). With `task_create` hidden, the bot offered a connected task app instead. The board's two tools are now core while the board is on, and the system prompt names the board as the place for work filed for later. `coreToolNames` test.
- **A bot could not file a task for itself.** The internal route reused the peer-reach rule, which excludes the caller by design, and `list_bots` never lists the caller, so "assign it to yourself" came back "different section". The route now accepts the sender's own id or `"me"`; the tool says so. `task-routes.test.ts` calls in as the bot through the test-only capability key, which the fixture launcher now forwards.

- **An assigned task waited in silence.** The board only runs a task for a bot on "Approve for me" (Auto), by design; a bot on Ask had its task sit at `ready` with no message anywhere. The dispatcher now exposes the reason in words (`hold`); `task_create` says it at once and `task_list` repeats it. Unit + route tests. The board view showing the same text is a UI follow-up.

## Tried by hand (rule 7)

Fresh bot, nothing changed (Claude, approval unset), all Phase 2 flags on together, live engine, standalone harness on a throwaway folder. All seven steps passed; the bot's actual replies are in `docs/bench/scorecard/2026-09-16-phase2-live.md` under "First-run pass". Not tried live: fallback on quota, the 24k spill, routine overlap (each has fake-engine tests).

## Platforms

Server only; the card renders through the existing option card on every client.

## Checks

`pnpm typecheck`, `pnpm lint` clean; board suites, the budget e2e, dispatch and card tests green.


milind-soni:

PR #1332, opened 2026-09-16
feat(board): a board screen in the app (phase 3 part 2)

aivsomkar:
Phase 3 part 2 (`docs/plans/2026-09-16-phase-3.md`), stacked on the gates PR. Omkar's addition to the phase on Sep 16: the board had no screen — a person could only ask a bot what was on it.

## In plain terms

- **A "Task board" entry in the sidebar**, offered only where the server has the board on (the server now publishes `features.board`).
- **The board by status**: to do, ready, running, paused, review, done. Each card shows the assignee, owner, due date, spend against its cap, *why it waits* (the hold reason from #1325), the gate line from part 1, and the result behind a disclosure.
- **Actions go through the existing routes only**: mark ready, assign, allow more money or finish without a cap on a paused task, mark done, back to ready, archive, open the run thread; a small form files a new task. The page never becomes a second scheduler.
- Refreshes every 10 s and on demand.

## Measured

`src/lib/task-board-view.test.ts`: the six columns in the board's order with archived work hidden, newest first; the raise step (the cap again, never under a dollar — the pause card's rule); the spend label; the card tone (paused, blocked or a failed gate = danger, waiting = warning, review = accent). `src/lib/feature-flags.test.ts`: the entry shows only when the server says so. `server/task-routes.test.ts`: the public list carries `hold`.

## Tried by hand

OMB3 built from this branch on the test folder: the sidebar's Tools menu lists "Task board"; the page shows Momo's "Say hello" task under Review with owner, $0.17 of $1.00, the result and its three actions; empty columns render with a zero count. Screenshot in the conversation record. Known small thing: at a narrow window the six columns scroll sideways.

## Found by hand (Omkar, first look)

Six fixed-width columns put the Review column off the right edge; the card could not be moved anywhere and the page had no way back. Fixed in the second commit: the columns are a grid that fits the window (two, three or six across), a card drags onto the columns the board would accept from a person (to do → ready, paused → ready, review → done or ready; targets light up while dragging; `dropTargets` is a tested helper), and the header has a close button back to the chat.

**Second look (Omkar):** at a medium window the page still ran past the right edge. Measured in a browser: the page was 849px wide in an 1100px window because the subtitle never wrapped and set the page's minimum width. Third commit: the page has a zero minimum width and clips, the subtitle wraps and hides on small windows, the columns are as many as fit the page's own width at 200px each (six wide, three medium, two narrow), card text wraps. Verified at 600, 800, 1100 and 1700px: the page never exceeds the window and no card leaves its column.

## PRs checked

#1202 (open, external): a task board with dated cards on its own JSON store and routes — superseded on storage (the board is `tasks.db`, there must not be two); its shape lessons taken (sidebar entry, cards by owner, Start through the thread routes). This page is ~450 lines on `/api/tasks` rather than 8.6k lines and a parallel store.

## Platforms

|---|---|---|
| macOS     | yes | in this PR, tested locally (OMB3) |
| Windows   | yes | in this PR (shared web UI + server; no platform gate); NOT built on Windows |
| iOS       | yes | follow-up: a board screen in `ios/App` reading `/api/tasks` |
| Android   | yes | follow-up: the same in `android/app/.../ui` |
| Companion | no  | n/a: the phones do not call the board routes yet; the allowlist entry comes with the phone follow-ups |

- [ ] follow-up: iOS board screen
- [ ] follow-up: Android board screen

## Checks

`pnpm typecheck`, `pnpm lint`, `pnpm i18n:check` clean; client and board suites green; full vitest on the branch tip 7,201 passed (one unrelated flake passes alone).


milind-soni:

PR #1340, opened 2026-09-16
feat(graphs): routine runs as checkpointed graphs — check, judge, ship after the turn; resume on restart (phase 3 part 4)

aivsomkar:
Phase 3 part 4 (`docs/plans/2026-09-16-phase-3.md`), stacked on the verifier PR. Spec: `docs/superpowers/specs/2026-09-16-phase-3-graph-runner-design.md`.

## In plain terms

- **A routine run was one turn and one status.** Now it is a graph — intake → work → check → judge → ship — with a checkpoint per node in `graphs.db`. The turn is the `work` node and runs exactly as before. After it, the harness runs the folder's gates (part 1), asks the verifier with the routine's instructions as the acceptance text (part 3), and ships: the run completes with the verdict on its own field, or fails with the verdict as the reason. The routine's instructions stay the source; the graph is derived from them.
- **A crash no longer loses the run.** A run found running after a restart whose turn is checkpointed resumes at its next node instead of failing as "restarted while running". A finished node's output is reused, never recomputed, so the resume makes no new turn. That reuse is the recording driver the roadmap asked for, at the harness level: a run whose nodes are all recorded replays with zero model calls.
- `verify.routines` (default on) switches it off; room-goal runs are unchanged in v1.

## Engines

Full on every engine: the graph is harness-side; the judge uses the bot's own engine through the same one-shot path as the board verifier.

## Measured

`server/graph-runner.test.ts` (nodes, checkpoints and reuse, failure, completion, the resumable set, reopen); `server/routine-graph.e2e.test.ts` (fake engine): a run completes with `verdict` "Verified: complete" and its own report untouched, nodes intake/work done, check skipped (no folder), judge/ship done; a run whose prompt carries `[[fake:incomplete]]` fails with the verdict as its error; a run put back to running with its judge node interrupted, then a server restart, completes from the checkpoint and the fake engine's prompt log gains no turn. The existing routine suites (continuity, results, delegation, cron, startup) stay green; the fake's verifier dump goes beside the turn's dump so fixtures keep reading the turn's prompt.

## Tried by hand (rule 7)

One fresh bot on Auto, live Claude (`docs/bench/scorecard/2026-09-16-phase3-live.md`, "Part 4"): a routine asked for the day of the week completed with `Verified: complete (confidence 0.80)`; a routine asked for the contents of a file that does not exist ended `failed` with `Verified: not complete (confidence 0.95) — Bot did not output the exact contents …` as its reason, after the bot rightly refused. Not tried live: the restart resume (simulated in the e2e).

## PRs checked

#60 (open, external): turn-level checkpoints in `bots.json` — a different grain (this is per node, per run); #1309/#1304 (merged on main) touch `routines.ts` and are reconciled at the chain's re-stack; #372's lesson: a completed node carries evidence (a gate result, a verdict), never only a model's word.

## Platforms

|---|---|---|
| macOS     | yes | in this PR, tested locally (standalone harness, live engine) |
| Windows   | yes | in this PR (shared server code, SQLite via node:sqlite as the board already uses); NOT built on Windows |
| iOS       | no  | n/a: server-side; the phones read the run's `verdict` and `error` through the existing routes |
| Android   | no  | n/a: same |

## Checks

`pnpm typecheck`, `pnpm lint` clean; full vitest on the Phase 3 tip 7,224 passed.


milind-soni (changes requested):
Blocking correctness issue in server/index.ts near the graph ship step: `const ok = verdict ? verdict.isComplete : true` returns success when deterministic gates have failed but the judge is unavailable, budget-blocked, or throws. The judge branches explicitly skip in these cases, leaving verdict null. Failed deterministic checks must independently prevent completion; unavailable verification must not become positive evidence. Please add regressions for a failed gate with (1) an engine without generateText, (2) exhausted verifier budget and (3) a verifier error. Also distinguish an unverified outcome from verified success. This is a source-level review finding, not a live-provider reproduction.

milind-soni:

PR #1341, opened 2026-09-16
bench(bakeoff): the engine bake-off on the board, and a Codex one-shot so its work is judged (phase 3 part 5)

aivsomkar:
Phase 3 part 5 (`docs/plans/2026-09-16-phase-3.md`), stacked on the graph-runner PR. Spec: `docs/superpowers/specs/2026-09-16-phase-3-bakeoff-design.md`.

## In plain terms

- **The engine bake-off.** `scripts/bench/bakeoff.ts` files the same three tasks (write a file with exact contents, a multiplication answered with the number only, a count of the board through a tool) on bots that differ only in engine, one bot per trial so every trial starts cold, and scores each run on the task's own check, the verifier's verdict and their agreement, attempts, tokens, cost, wall time, cards raised and policy denials, averaged over trials, as one markdown table.
- **Codex gets a one-shot call.** The first live run showed every Codex task unjudged: the Codex driver had no one-shot, so the verifier skipped it (F20). `codex exec` — prompt on stdin, read-only, ephemeral, last message to a file, 180 s bound — is now the driver's `generateText`; the fake Codex CLI answers `exec` the same way.
- Deferred with reasons in the spec: the planner/executor split and best-of-N with a judge (they need a hard task subset with ground truth); a screen for the table.

## Measured

Live, Claude vs Codex, 3 tasks × 2 trials (`docs/bench/scorecard/2026-09-16-bakeoff-live.md`): both engines 100% correct by the deterministic check; the judge agreed with the check on every Claude run and on 4 of 6 Codex runs — it refused Codex's right board count twice because it cannot see tool output (F21, follow-up: give it the last tool results' tails). Codex reports no price, so compare tokens: ~59k input per task on both. Smoke record on the fake engine: `2026-09-16-bakeoff-smoke.md`.

## Tests

`server/drivers/codex.test.ts`: the one-shot through `codex exec` (verdict JSON for a verifier prompt, fixed text otherwise); the Codex suites green. Two feature-shape assertions in `server/index.test.ts` fixed on this tip (`board: false`).

**Follow-on effect:** Codex now also drafts the compaction summary with a model call (it had none before). `server/context-compaction.e2e.test.ts` expected zero harness calls on Codex; the expectation is updated on the Phase 4 tip (`phase-4/4-routine-notes`), where the full suite otherwise passes.

## Platforms

|---|---|---|
| macOS     | yes | in this PR, run live |
| Windows   | yes | in this PR (the prompt goes through stdin for Windows' argv limit); NOT built on Windows |
| iOS       | no  | n/a: a bench script and a server driver call |
| Android   | no  | n/a: same |

## Checks

`pnpm typecheck`, `pnpm lint` clean; full vitest on the Phase 3 tip before this commit 7,224 passed.


milind-soni:

PR #1354, opened 2026-09-16
feat: shared wire model for the server-client contract

bradhallett:
## What
- shared/wire.ts becomes the single source of truth for the wire contract: WireTask, WireBot, WireGroup, and a 21-kind ServerFrame union derived from the client handleFrame switch in src/state/store.tsx.
- The server projections toWireTask, wireBot, and publicGroupState now return those shared types. Wire output is unchanged byte for byte. Compile-time AssertExact guards pin each projection to the client view, and wire.test.ts covers them.
- SECRET_FLAG redaction moves to shared/redact.ts with the regex byte-identical to the old server copy; taskTimeline imports it from shared. Token masking stays pinned by redact.test.ts and task-timeline.test.ts.
- No cross-layer imports: no file under src/ or shared/ imports server/.

## Why
The server and the client each kept their own copy of the wire vocabulary, so every frame or field change had to be mirrored by hand. One shared model makes the contract explicit and lets TypeScript catch drift at compile time instead of at runtime.

## Documented casts
Two casts remain, both at the dispatch boundary and both documented inline:
- message frame payload cast to the client Message type, because on the wire from.color is an honest string while the client view expects MausColor
- the config frame read as unknown as ConfigStatusFrame until model extraction lands

## Out of scope, deliberately
ConfigStatus frame typing, model extraction, and server-side adoption of the shared routine and webhook types. Each is a separate change and none is left half done here.

## How tested
At 51bece04 on main 13ca6ea9:
- tsc -b clean, tsc -p tsconfig.server.json --noEmit clean, oxlint --deny-warnings clean
- server touched suites 252/252 across 10 files, including webhook-ingress and routines-startup
- client touched suites 202/202 across 8 files
- byte fidelity checked with an AssertExact compile probe: clean tree passes, a dummy BotRecord field fails at the projection guard

## Size
33 files, +1201/-998, move heavy: most of the delta is existing types and projections moving to shared/ plus import updates. The change splits badly because the types, projections, and redaction move together.

## CI notes
- Fork Vercel checks report Authorization required to deploy; that is systemic on fork PRs.
- Main run 35100164327 is currently red on verification-docs (computer-wait.md assertion) and managed-desktop-cleanup (async is not defined). This diff touches neither area.

## Summary by CodeRabbit

- **New Features**
  - Added shared support for routine scheduling, routine runs, webhook triggers, delivery attempts, and ingress status.
  - Added standardized inspector data for runtime and native activity.

- **Improvements**
  - Unified client-visible data formats across tasks, bots, groups, messages, notifications, and runtime events.
  - Improved consistency of secret redaction across application surfaces.
  - Strengthened type safety for API responses and live event updates.

- **Tests**
  - Added validation ensuring private server data is excluded from client-facing task and bot responses.


bradhallett:
Classified the macos leg red before retriggering. The single failure is the timing-sensitive assertion at server/team-setup.e2e.test.ts:89 (a bot from a denied team-setup request still present). Evidence it is a flake, not this diff: the same test passes on the ubuntu and windows legs of this very run, passes when run locally at 51bece04 on macOS, and passed on the two most recent macos legs of other branches (0feb4b28 for #1351, b5a94622 for #1353). Pushed an empty retrigger commit with this note; no code changed.

PR #1357, opened 2026-09-16
fix(ios): prevent kana input from triggering send in Release builds

### LINKED ISSUE #1334: iOS Release: Japanese kana input disappears or sends unexpectedly
## Confirmed symptom

On a physical iPhone 16 Pro running iOS 26.6, TestFlight OpenMausMobile 1.0.0 (14) intermittently loses Japanese kana/flick input in the chat composer. A flicked character can briefly appear underlined and then disappear. The reporter observes this in both bot conversations and channel/group conversations. They report that the previous app version worked; its build number is unconfirmed.

There is a more serious variant: typing `あ` → `い` → `う` without touching Send or Confirm can send the draft as a message. The reporter observed `あいう` sent in TestFlight. They also reproduced an unintended send in a separately signed `OMB IME Probe` app paired only to an isolated fake-engine bot. The private TestFlight recording contains real conversation content and has not been uploaded.

Only this device, keyboard and OS are confirmed to exhibit the user-facing symptom. Effects on other iOS releases or Chinese/Korean IMEs are not established.

## Reproduction

1. Open a bot or channel conversation in TestFlight build 14 and focus the composer.
2. Use the stock Japanese kana keyboard to flick `か` down to `こ`, or type `あ` → `い` → `う` without pressing Send/Confirm.

Expected: marked text remains in the composer until the user explicitly confirms or sends it.

Actual: a marked character can disappear; during multi-character input, the draft can be sent without a send action. The keyboard and insertion caret may remain visible, so a lasting focus loss is not supported by the recording.

## Root cause and isolation

The composer uses `ComposerKeyboard.shouldSubmit(previousText:proposedText:)` to distinguish a software Return newline from ordinary edits. Its newline predicate passed the bound method `CharacterSet.newlines.contains` directly to `allSatisfy`. In an optimized Release build, this misclassifies ordinary inserted characters as Return and can fail to recognize an actual newline.

This was reproduced independently of TestFlight build 14's unpublished source:

- Current public `origin/main` (`13ca6ea9`) was built as a separate-bundle-ID **Release** probe, paired only to an isolated fake-engine server. On the affected iPhone, three kana input actions caused an unintended send. The reporter reproduced the send manually in the same probe. Single-character and offline Debug checks had previously passed, so those negative results did not cover this path.
- Instrumentation on that probe recorded `draft=""`, proposed text `"あ"` (U+3042), a one-character insertion, and a direct Unicode newline check of `false`, while `ComposerKeyboard.shouldSubmit("", "あ")` returned `true`. The instrumentation was removed after diagnosis and is not part of the proposed fix.
- macOS `swift test -c release --filter ComposerKeyboardTests` failed 4 of 5 cases before the fix: Japanese and ordinary character insertions were treated as submission, while Return was missed. Replacing the bound method reference with an explicit closure made all 5 cases pass. The new Japanese regression test also passed.
- On the same iPhone and isolated conversation, a Release probe containing only that one-line production fix retained three kana input actions in the draft for 1.5 seconds and sent no new message. The reporter then manually typed `あいう` in the fixed probe and confirmed normal behavior. The existing TestFlight app and its data were never modified.

The previous explanation in this issue that `ComposerKeyboard.shouldSubmit("", "こ")` *must* return `false` was based on source inspection and Debug behavior. The Release runtime results above correct it. Build 14's exact source is still unknown, but is no longer required to reproduce the fault in public source.

## Fix acceptance

- Japanese kana text remains in the composer until an explicit send; no automatic send occurs while typing. The software Send/Return key still submits; pasted multiline text and ordinary edits do not.
- Run the classifier regression tests in optimized Release configuration in CI, since Debug-only tests missed this fault.
- Do not include private conversations in public evidence or modify the reporter's installed TestFlight app/data.

shuhei-suzuki:
Closes #1334.

## What changed

- Keep the newline predicate in `ComposerKeyboard` as an explicit closure. Passing the bound `CharacterSet.newlines.contains` method directly to `allSatisfy` misclassifies ordinary characters in optimized Release builds.
- Add Japanese kana / Return regression assertions.
- Run this classifier suite in Release configuration in CI; the existing Debug suite did not catch the failure.

## Evidence

- **Before:** `swift test -c release --package-path ios --filter ComposerKeyboardTests` failed 4/5 cases: `あ` and ordinary `!` insertions were treated as submit, while Return was missed. An isolated Release probe on the reporter's iPhone 16 Pro / iOS 26.6 sent a kana draft without touching Send.
- **After:** the same Release suite passes 5/5 cases. The separately signed Release probe kept three kana input actions in the composer for 1.5 seconds; the isolated server received no new message. The reporter manually entered `あいう` in the fixed probe and confirmed that it behaved normally.
- The probe was paired only to a disposable fake-engine server. The installed TestFlight app and the reporter's conversations were not modified. The probe's temporary bundle-ID/signing changes and diagnostic instrumentation are **not** in this PR.

This changes no visual layout. The TestFlight build 14 source commit remains unknown, but current public `main` Release source independently reproduces the failure and the one-line fix resolves that reproduction.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where typing Japanese kana characters could incorrectly trigger message submission in optimized builds.
  - Japanese text now remains in the composer as characters are entered and submits only when a line break is added.

- **Tests**
  - Added coverage to verify Japanese kana input behaves correctly before and after pressing Return.


milind-soni:
Superseded by #1335 — closing rather than resolving the conflict, because the conflict is `DU`: main **deleted** the file this patches.

`3913155c` (#1335, "keyboard Return inserts a newline, only the arrow button sends") removed `ComposerKeyboard.swift` and its tests entirely. `shouldSubmit` no longer exists anywhere in `ios/`, so there is nothing left to rebase onto. #1335 also addresses the kana problem more fundamentally than this could: Return cannot trigger send at all now, so no IME classifier is in the path.

**On the optimizer claim, which I took seriously.** The real substance here was that `allSatisfy(CharacterSet.newlines.contains)` misclassifies as a bound method reference under optimization. That pattern still exists on main in three security-relevant places, all rejecting control characters:

- `ios/Sources/CompanionCore/Client.swift:1257`
- `ios/Sources/CompanionCore/MessageAttachments.swift:140`
- `ios/Sources/CompanionCore/MessageAttachments.swift:155`

If the claim held, those sanitizers would silently pass control characters in Release builds — worth porting the fix for. So I tested it before doing that:

| build | bound reference | closure |
|---|---|---|
| `-Onone` | true | true |
| `-O` | true | true |

Identical on Swift 6.3.3 / arm64-apple-macosx — the same toolchain family CI uses on `macos-latest`. I could not reproduce the misclassification, so I have not ported the change to those three call sites.

That is not a flat contradiction: my test was a standalone snippet, and a module context, a different Swift version, or a real device build could behave differently. **If you reproduced this on an actual iOS Release build, please reopen with that detail** — the three call sites above would then genuinely need fixing, and the release-mode test step this PR adds would be worth having on its own merits.

Thanks for finding the pattern either way.

PR #1358, opened 2026-09-16
fix(smoke): count the fake's own read lines, not the notes that quote them

milind-soni:
## Why main is red

Every `typecheck + test` leg that gets past vitest has been failing in `pnpm exec electron scripts/smoke-approval-modes.cjs`:

```
AssertionError [ERR_ASSERTION]: Repeated reads complete without another prompt
3 !== 2
```

The native-Auto Grok scenario counts `list_bots:` in the reply to prove two reads completed without a second prompt. Since #1302, every finished turn appends a daily-log line to `memory/log/<date>.md` quoting the reply's first line — here `list_bots: Reachable teammates…`. The fake's middle read is `session_search "approval fixture"`, and from the second (resumed) turn on it finds that log line as a memory hit and the first turn's reply as a conversation hit, so a snippet hands `list_bots:` back into the reply. Three matches, still one prompt. Whether a leg goes red depends on whether the memory index has caught up by the second turn, which is why macOS hit it while ubuntu/windows passed on the same commit (b94f4a61) and why the reverse happened on #1282.

## Fix

Count only lines that start with the marker. The fake writes each completed read as its own line; every echoed copy sits mid-line inside a one-line snippet (both snippet paths collapse whitespace).

## Verification

Reproduced locally under real Electron with a debug dump of the reply text: turn 0 → 2 matches, turn 1 → 3, the third inside `- [memory file memory/log/2026-09-16.md] - 21:46 · from chat "[Approval] [fixture] grok-4.6 auto" · list_bots: …`. With the line-anchored count the full smoke passes (`Approval smoke passed; HTTP elevation rejected, private grant and resumed mode transitions verified.`).

Main's last green run was 82d277ae (Sep 15 16:13Z). First red run failed on other, since-fixed tests; this assertion is what remains on b94f4a61 and on every open PR rebased onto main (#1205, #1234, #1282, #1286, #1321).

## Summary by CodeRabbit

* **Tests**
  * Improved smoke-test validation for Auto mode output by matching expected entries only at the start of lines.
  * Prevented quoted references in later responses from being counted as additional reads.


milind-soni:
Superseded by #1370, which landed the same line-anchored read count (plus the Qwen/Gemini matrix updates) while this was in CI. Closing to avoid a duplicate change.

PR #1375, opened 2026-09-16
fix(index): openai-compat and minimax already replay natively

banana-monkey-tree:
## What changed

`replaysNatively` (passed into `buildTurnContext`) now checks a small exported list of driver
kinds — `grok`, `openai-compat`, `minimax` — instead of only `driverKind === "grok"`. The list
lives in `turn-context.ts` next to the field it feeds, with a comment explaining why: all three
are built on the same shared `createOpenAIChatRuntime`, whose `messagesFor()` already maps
`turn.transcript` into chat messages on every turn, so they all already "replay natively" and none
of them need `buildTurnContext` to also inline the transcript into the turn text.

## Why

`openai-compat` and `minimax` were missing from this check, so on a rewound, fresh, or
externally-updated turn, `buildTurnContext` inlined the whole transcript into the outgoing turn
text *in addition to* the transcript these drivers already send structurally — the same
conversation history landed in the request twice. `grok` was already correctly excluded from this
double-send; `openai-compat` and `minimax` share the exact same runtime and had the exact same bug.

## How it was verified

  files, 7116/7117 tests passing (one net-new test), both before and after. The 4 failures are
  pre-existing on `main` and unrelated to this change (a browser-bundle path test, a
  terminal-escape-sequence rendering test, and a team-import ordering test) — confirmed by
  running the identical suite on `main` at the same commit.
- New test in `server/drivers/openai-compat.test.ts`: builds a rewound turn with a sentinel string
  in the transcript, sends it through the real driver, and counts the sentinel's occurrences in the
  actual outgoing request body. Fails on `main` (2 occurrences) and passes on this branch (1).

## Screenshots (UI changes)

N/A — server only.

## Checklist

- [x] No `dist-server/` edits

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed duplicate transcript handling when continuing rewound conversations.
  - Improved replay behavior for Grok, OpenAI-compatible, and Minimax integrations by ensuring conversation history is included consistently.
  - Clarified replayed delegated responses by distinguishing their source and safely formatting their message content, preventing confusion with user-authored text.


banana-monkey-tree:
Same as on #1377/#1378: the two failing `server/browser-codex-path.integration.test.ts` cases ("keeps the browser mounted and completes a turn with the default/absolute Codex CLI") aren't from this branch — they also fail on `main` at the same commit, and on unrelated PRs from other contributors (e.g. #1384). This branch only changes which driver kinds are listed as replaying natively. Happy to rebase whenever a fix for those lands.

banana-monkey-tree:
Merged current `main` (`0c327bca`) into this branch and revalidated it end to end, since a fair bit has landed since I opened it.

**The defect still reproduces on `0c327bca`.** `server/index.ts` still passes `replaysNatively: instance.driverKind === "grok"` to `buildTurnContext`, while both OpenAI-compatible and MiniMax instantiate `createOpenAIChatRuntime`, whose `messagesFor` already turns each transcript entry into a chat message before appending the turn text. So those two drivers receive history twice on the inline-replay path.

Verified through the repo's isolated fixture server against a loopback provider rather than by reading: create a bot on OpenAI-compatible, send a sentinel, send a second message, edit the second message. The resulting provider request contains the sentinel **twice** on `main` and **once** on this branch, with the first user message and assistant reply still present as separate chat messages in both.

**Conflict resolution**, for the record:

- `server/drivers/openai-compat.test.ts` — your streaming-usage test and this branch's replay test were added at the same place. Both kept whole; they assert different things and neither is weakened.
- `server/index.ts` — kept `peerMessageText`, `decideContext`, and the stale-handoff, external-update and recovery bookkeeping, and applied the native-replay membership check inside `decideContext` in place of the Grok-only condition, rather than restoring this branch's older context block.

Final diff against `main` is three files, +61/−2.

**On CI, correcting my earlier comment.** I previously said the only failures were two `browser-codex-path` cases. That was too narrow a claim to have made. Running the full suite on this branch and on a clean `0c327bca` export, I saw eleven failures on the branch run and three on the baseline — so I checked attribution rather than assuming, by replacing this branch's production files with `git show 0c327bca:PATH` **in the same checkout** and re-running. The same ten `harness HTTP API` failures persisted with upstream's own code in place, so they are checkout-dependent and not caused by this change. Two further baseline-only failures came from the export sitting under `/tmp`, which changes which rejection the DEB hook reaches. Both `browser-codex-path` cases now pass locally on both revisions.

`pnpm typecheck`, `broker:test`, `test:electron` and `test:packaged-server` pass on both. All of `openai-compat`, `minimax`, `openai-chat`, `turn-context`, `openai-tools.e2e` and `delta-context.e2e` pass on both.

What I can't speak for: Windows, macOS and the Swift/iOS job, none of which I can run — this was Linux arm64. The previous run's iOS job was cancelled on the 35-minute limit rather than failing. Fresh CI on this head is the real check.

milind-soni:
Reviewed integration at 211e20a1: Ported the native-replay flag to the new dispatch-time closure, preserving main compaction behavior. Added an isolated HTTP rewind regression that captures an outgoing OpenAI-compatible request and verifies history appears once. 46 driver/context tests plus the new end-to-end test, typecheck and lint pass. Normal auto-merge is enabled, subject to required GitHub checks; no check bypass.

PR #1376, opened 2026-09-16
fix(codex): settle a requested stop as interrupted

banana-monkey-tree:
## What changed

`CodexDriver`'s `child.on("close", ...)` handler now checks `stopRequested` before falling
into its exit-classification logic. When a turn is stopped via `interruptTurn()`, the app-server
child is killed and this handler fires before the app-server itself ever sent `turn/completed`.
Previously that always fell through to a `runtime.error` event and `stopReason:
"exit_before_result"` — the turn was misreported as an unexpected crash even though the stop was
requested by the driver itself. It now settles immediately as `stopReason: "interrupted"`, with
no `runtime.error` event.

Nothing else in the close handler changes — the retry-eligibility path (already gated on
`!stopRequested`) and the exit-classification logic used for genuine crashes are untouched.

## Why

`stop()` kills the child process and sets `stopRequested = true`, but does not itself call
`settle()`. The close handler that eventually runs had no way to distinguish "the app-server
process died because we asked it to stop" from "the app-server process crashed on its own" —
both look identical from the handler's point of view (child exited, turn not yet settled), so
every requested stop was reported as a failure.

## How it was verified

  files, 7116 tests passing, both before and after. The 4 failures are pre-existing on `main` and
  unrelated to this change (a browser-bundle path test, a terminal-escape-sequence rendering
  test, and a team-import ordering test) — confirmed by running the identical suite on `main` at
  the same commit.
- The new test in `server/drivers/codex.test.ts` (extending an existing "rejects a second turn
  while one is in flight" case) fails on `main` and passes on this branch.

## Screenshots (UI changes)

N/A — server only.

## Checklist

- [x] No `dist-server/` edits

## Summary by CodeRabbit

- **Bug Fixes**
  - Interrupted turns now finish with an explicit “interrupted” status.
  - Prevented unnecessary runtime errors and misleading exit-related errors when stopping a turn.
  - Improved handling when a new turn is requested while another turn is still in progress.


banana-monkey-tree:
Superseded by #1323, which landed the identical stopRequested guard (plus a more thorough graceful-interrupt path and test coverage) as part of the steer feature. Closing — no separate fix needed.

PR #1377, opened 2026-09-16
fix(openai-chat): parse the leftover buffer on stream EOF

banana-monkey-tree:
## What changed

The OpenAI-compatible chat streaming runtime (`createOpenAIChatRuntime`, shared by the
OpenAI-compatible, MiniMax and Grok drivers) now parses whatever is left in its buffer when the
HTTP stream closes, instead of only ever parsing complete newline-terminated `data: ...` lines.

The per-line parsing logic (updating text, reasoning, tool calls, `finishReason` and `usage`) is
pulled out into a small `consumeDataLine()` closure that both the normal per-line loop and the
end-of-stream path call. No behavior changes for a normally-terminated stream — this only adds
handling for the one case that previously fell through the cracks.

## Why

Several OpenAI-compatible servers close the connection right after their last frame, with no
trailing newline and no `data: [DONE]` sentinel. The reader loop only updated its parsing state
(`finishReason`, `usage`, streamed text) inside the newline-delimited loop, so a final frame that
arrived as the literal last bytes before the socket closed was silently dropped — either the whole
turn failed with "Stream ended before completion" (if no earlier frame had set `finishReason`), or,
if it had, the turn completed successfully but with `usage: null`, permanently losing the
turn's token accounting.

## How it was verified

  files, 7116/7119 tests passing (three net-new tests), both before and after. The 4 failures are
  pre-existing on `main` and unrelated to this change (a browser-bundle path test, a
  terminal-escape-sequence rendering test, and a team-import ordering test) — confirmed by
  running the identical suite on `main` at the same commit.
- Three new cases in `server/drivers/openai-chat.test.ts`, covering the tail-parsing paths that
  weren't exercised before: a final finish_reason+usage frame with no trailing newline before the
  socket closes; the same final frame split across two separate stream chunks (part of it in one
  `read()`, the rest — including the missing newline — in the next); and an empty final flush
  (stream ends cleanly right after a newline-terminated frame, nothing left to parse) to confirm
  the new EOF handling doesn't misfire when there's nothing left in the buffer. All three fail on
  `main` and pass on this branch; the existing `[DONE]`-terminated and MiniMax
  `[DONE]`-less-with-trailing-newline cases are untouched and still pass, as do `minimax.test.ts`
  and `openai-compat.test.ts`, which share this runtime.

## Screenshots (UI changes)

N/A — server only.

## Checklist

- [x] No `dist-server/` edits

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for streaming chat responses that end without a trailing newline.
  * Ensured final usage details and completion status are captured when responses are split across chunks or streams close unexpectedly.
  * Prevented false errors and empty updates when streams end cleanly after a newline.
  * Improved handling of provider errors, malformed responses, tool calls, and reasoning or content updates.
* **New Features**
  * Improved Markdown handling for links and images that use Windows-style file paths.


banana-monkey-tree:
The two failing `server/browser-codex-path.integration.test.ts` cases ("keeps the browser mounted and completes a turn with the default/absolute Codex CLI") also fail on `main` at the same commit with none of this branch's changes — I re-ran the suite against a clean `upstream/main` worktree at the same SHA to check. They're unrelated to this PR, which only touches the OpenAI-compatible streaming path. Happy to rebase once a fix for those lands.

milind-soni:
Holding this one — CI caught a real regression, and it's the truncated-tail case.

**`vitest (windows-latest, shard 4/4)`**
```
FAIL server/delta-context.e2e.test.ts > offers the results again when the person
     stops the return turn before the provider acts on it
AssertionError: expected 'failed' to be 'settled'
```

**Why this PR causes it**

When a person stops a turn mid-stream, the socket closes with a *partial* frame in `buffer` — no trailing newline. On main that partial never enters the parse path at all: the `while ((newline = buffer.indexOf("\n")) !== -1)` loop simply doesn't fire for it, the leftover is dropped, and the turn settles.

With this PR the EOF branch now does:

```ts
const line = buffer.trim();
if (line && line !== "data: [DONE]") consumeDataLine(line);
```

so the truncated tail *is* parsed, `JSON.parse` throws, and the shared closure sets `malformedFrame = true`. That flag then reaches:

```ts
toolCalls: calls.finish(finishReason, malformedFrame)
```

and the turn hard-fails instead of settling. A stop goes from a clean settle to a non-retryable failure.

**Suggested fix**

The flush is the right idea — the bug is treating a truncated tail as a protocol violation. At EOF an incomplete frame is *expected* (a stop, an abort, a dropped connection); only a malformed frame that was properly newline-terminated means the provider actually sent something bad. So keep the flush but don't let it set `malformedFrame` — e.g. give `consumeDataLine` a second parameter:

```ts
const consumeDataLine = (line: string, atEof = false): boolean => {
  ...
  } catch {
    if (!atEof) malformedFrame = true;   // a truncated tail at EOF is not a bad frame
    return false;
  }
```

and call `consumeDataLine(line, true)` from the `done` branch. That keeps all three of your new tests passing (their final frames are complete, just unterminated) while leaving a stopped turn settling as it does today.

Worth a test for the truncated-tail case specifically — that branch is what `delta-context.e2e.test.ts` is hitting, and it currently has no direct coverage in `openai-chat.test.ts`.

The refactor itself reads cleanly and the three new tests are good; it's just this one path. Happy to push the change if you'd like.

PR #1378, opened 2026-09-16
fix(openai-compat): request streaming usage totals

banana-monkey-tree:
## What changed

One line: the OpenAI-compatible driver's `requestBody` now sends `stream_options: {
include_usage: true }` on a streamed turn, mirroring what the MiniMax driver already does in the
same file family.

## Why

Several OpenAI-compatible servers only include a token-usage object in their streaming response
when the client explicitly asks for it via `stream_options.include_usage`. Without it, a streamed
turn through this driver gets no usage accounting at all — `usage` stays `null` for the whole
turn, even though the server would have reported it if asked.

## How it was verified

I tested this against a real local OpenAI-compatible server (an MLX-based inference server
exposing an OpenAI-compatible API on localhost, small local model, no third-party network calls):

| request | result |
|---|---|
| `stream: true` + `stream_options: {include_usage: true}` | 200, final chunk carries a `usage` object (`prompt_tokens`, `completion_tokens`, `total_tokens`) |
| `stream: true`, no `stream_options` | 200, zero usage frames in the whole stream |
| `stream: true` + an unrelated unknown field | 200 (this server ignores unrecognized fields) |

So the omission is not cosmetic — it's the difference between getting token accounting on a
streamed turn and getting none at all — and the field is accepted and honored by at least one real
implementation, matching the OpenAI streaming spec and this repo's own `minimax.ts`.

**What I could not verify:** whether a *strict* OpenAI-compatible endpoint — one that rejects
unrecognized request fields outright — would break on this. I did not have access to one to test
against (untested: vLLM, llama.cpp's server, Ollama, LM Studio, Groq, OpenRouter, Together). If
that's a real concern for endpoints your users hit, a safer variant would be to send the field and
retry once without it if the request is rejected because of it — happy to send that instead if
you'd rather not risk it on the always-on version, since you have visibility into which endpoints
people actually use this driver against.

`pnpm typecheck`, `pnpm lint`, and the full `vitest run` all pass; the 4 pre-existing failures on
`main` at this commit (a browser-bundle path test, a terminal-escape-sequence rendering test, and
a team-import ordering test) are unaffected and unrelated to this change.

## Screenshots (UI changes)

N/A — server only.

## Checklist

- [x] No `dist-server/` edits

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved streaming requests for OpenAI-compatible providers by including usage details when streaming is enabled.
  - Non-streaming requests remain unchanged.


banana-monkey-tree:
The two failing `server/browser-codex-path.integration.test.ts` cases ("keeps the browser mounted and completes a turn with the default/absolute Codex CLI") also fail on `main` at the same commit with none of this branch's changes — I re-ran the suite against a clean `upstream/main` worktree at the same SHA to check. They're unrelated to this PR, which only touches the OpenAI-compatible streaming path. Happy to rebase once a fix for those lands.

banana-monkey-tree:
Follow-up evidence, from running this against a real OpenAI-compatible server (a local MLX-based one) on a released build rather than a fixture:

Two bots on the `openai-compat` driver, one real turn each. Both `turn.completed` events carried **no `usage` key at all**, and each bot's broadcast usage stayed `{"input":0,"output":0,"costUsd":null,"turns":1}` after a real prompt and a real generated answer.

Worth noting the driver already sets `includeUsageInCompleted: true` at the runtime-config level — it wants the numbers, it just never asks the server for them on a streamed turn. `stream_options` appears exactly once in the shipped bundle, in the MiniMax driver.

So the user-visible effect is that per-turn accounting silently reads zero for these engines, rather than being merely absent internally.

PR #1387, opened 2026-09-17
fix(chat): show an edited message immediately on every platform

aivsomkar:
## Summary

Editing a sent message left the old question and its old answer on screen until the new reply finished, which could take a while.

**Cause.** The server forked the conversation immediately but only broadcast the new message. Every client (desktop, iOS, Android) adopts a new message as the visible branch only when it follows the message already on screen. An edit is a *sibling* of the original, so the check failed and the new branch stayed hidden until a later snapshot, usually when the turn ended.

**Fix, part 1: server tells clients to switch.** `branchMessage` now also emits the `thread` leaf-move frame (the same frame version switching already uses). This alone fixes every client, including already-installed phone apps.

**Fix, part 2: instant swap on submit.**
- **Desktop (macOS + Windows):** `editMessage` adds an optimistic sibling that becomes the leaf, so the old answer disappears on Enter. It reconciles onto the server fork by `sendId` (same mechanism as normal sends). A failed edit restores the previous branch, not just the parent.
- **iOS + Android:** the phones deliberately never fold unconfirmed state, so they get a `PendingEdit` display overlay in `CompanionState.visibleTranscript`. It swaps the text in and hides the old answer while the request runs. The fork returned by the edit response is adopted immediately, without moving the leaf back if a reply already arrived. Reactions and a second edit are disabled on the stand-in row.

**Retry safety.** The edit route accepts an optional `sendId`, stored on the fork. A retry with the same id returns the existing fork (checked before the busy guard, since that fork's own turn is what is busy) and a reused id with different text is a 409.

## Test plan

- [x] `vitest run src` (1766 passed), server `store`, `branching` e2e, `message-db`, `team-backup` tests
- [x] New server tests: fork emits message then thread frame; retried edit returns the same fork; reused sendId is refused
- [x] New desktop reducer tests: instant swap, reconcile + late POST does not rewind, failure restores old branch
- [x] iOS `swift test` StoreTests/ApprovalClientTests, including 4 new pending-edit/adopt tests; iOS app simulator build succeeds
- [x] Android `:core:test :app:testDebugUnitTest :app:assembleDebug`, including 4 new store tests
- [x] `tsc -b`, `tsc -p tsconfig.server.json`, oxlint on changed files
- [x] Live desktop check on an isolated server with a fake engine that holds the reply open: 150ms after pressing Enter the edited question shows with 2/2 versions, the old question and answer are gone, the bot is still thinking; the released reply then lands under the edit
- [ ] Hands-on edit on a real iPhone and Android phone

## Platforms
|---|---|---|
| macOS     | yes | in this PR, verified in the dev UI against an isolated server (not packaged as OMB2) |
| Windows   | yes | in this PR (shared server + desktop UI code, no platform gate); NOT built on Windows |
| iOS       | yes | in this PR; unit tests + simulator build, not run on a device |
| Android   | yes | in this PR; unit tests + debug APK build, not run on a device |
| Companion | n/a: no new route; the edit route is already allowlisted and only gains an optional `sendId` body field |

## Summary by CodeRabbit

* **New Features**
  * Edited messages now appear immediately while updated responses generate across web, Android, and iOS.
  * Completed edits automatically switch conversations to the new response branch.
  * Retrying the same edit reuses the existing result instead of creating duplicates.

* **Bug Fixes**
  * Failed edits restore the original message and response.
  * Pending edit placeholders can’t be edited again or receive reactions.
  * Late responses no longer hide replies that have already arrived.
  * Optimistic placeholders are protected from version switching and regeneration actions.


milind-soni:
Heads-up on overlap: part 1 here (the `thread` frame from `branchMessage`) is the same change as #1386, which was opened a few minutes earlier with the same store tests and is set to auto-merge. Once it lands, this PR will conflict in `server/store.ts` and `server/store.test.ts`. If the optimistic client swap (part 2) is still wanted, rebasing this onto main and dropping the server half would leave that as a clean, separately reviewable change — that call is Milind's; noting it here so nobody merges the same lines twice.

milind-soni:
I rebased this onto main and resolved the conflicts (main landed the server half as #1386, so `store.ts` was comment wording only; `store.test.ts` had each side pinning a different thread frame exactly and the other loosely, so I kept the exact form of both; `ios/Store.swift` was additive on both sides and kept both). Build and `store.test.ts` are green — 97 tests, and removing the leaf-move emit fails 2 of them, so that half is solid.

**Holding on the Android half, which cancels its own edit request.** I traced all four links in the current code:

1. `MessageRow.kt:119` — `val scope = rememberCoroutineScope()`, owned by the row for the message being edited.
2. `MessageRow.kt:328` — `scope.launch { session.edit(message, bot, text) }`.
3. `Session.kt:2040` — `edit()` writes `pendingEdits` into `_state` **before** it awaits `perform { … }`.
4. `Store.kt` `visibleTranscript()` — returns `branch.subList(0, index) + standIn`, i.e. the source message is **removed** and replaced by a row whose id is `pending.placeholderId`.

`ChatScreen.kt:1050` keys the list `key = { messages[it].id }`. So the moment step 3 lands, the row from step 1 leaves composition, its remembered scope is cancelled, and the coroutine suspended inside the HTTP call is cancelled with it — `invokeOnCancellation { call.cancel() }` aborts the OkHttp call, often before the request is even written. The `finally` then clears `pendingEdits`, the stand-in flickers for about a frame, and the old question and old answer come back. `adoptEdit` can never run on Android.

This is new to this PR: on main, `edit` only mutates state *after* the await, so the row survived until the work finished. iOS is unaffected because it launches an unstructured `Task {}` that isn't tied to view lifetime.

**Suggested fix:** don't run the request on a scope the optimistic update destroys. `Session` already holds a long-lived `scope` (Session.kt:33) — running the network part on it, rather than on the caller's row-scoped one, matches what iOS effectively does and keeps the optimistic swap. Hoisting the `launch` to a screen-level scope would work too.

I haven't pushed a fix for this: it's your design area, and I have no JDK here, so I couldn't run the Kotlin tests to verify a lifecycle change rather than just argue it. Happy to take a pass if you'd like, but it needs a real device or emulator run either way — note the Kotlin suite passes today, so CI will not catch this.

milind-soni:
Triage result: holding this merge after reproducing a mobile branch regression. Store.adoptEdit unconditionally promotes the returned fork when it is outside the current ancestry. Sequence: edit response is delayed, the server selects a newer branch, then the old edit response arrives; iOS/Android switch back to the old fork. A review-only Swift regression reproduces it (32 existing Store tests pass; the new case fails); 222 existing desktop/server tests also pass. Please fence edit adoption against newer authoritative branch/selection changes. Also keep the edit request identity through an ambiguous transport failure: generating a new UUID on every retry defeats the server receipt and can run the edit twice.

PR #1388, opened 2026-09-17
fix(test): keep the Codex fake dependency-free — inline its atomic dump write

milind-soni:
## Why main is red

Since #1372, `server/browser-codex-path.integration.test.ts` fails both cases on macOS and ubuntu (main run at 53dd47fc, and every PR rebased since). The turn ends with:

```
error: codex exited 1 before turn/completed: … ModuleJob.link (node:internal/modules/esm/module_job:277:33)
```

#1372 made `server/testing/fake-codex-app-server.ts` import `writeFileAtomic` from `../atomic.ts`. The browser PATH test strips that fake to a plain `.mjs` in a temp bin dir and runs it as `codex` (that is the point of the test: an extensionless executable on a minimal PATH), so the relative import cannot resolve and the fake dies at link time. The fake's own header already says it must stay dependency-free.

## Fix

Inline the atomic write in the fake (sibling temp file, `renameSync` over the target, brief retry on the Windows `EPERM`/`EACCES`/`EBUSY` rename refusals, same codes and delays as `atomic.ts`), and document why the file cannot import from the repo.

## Verification

- Bisect: parent of #1372 (76e4a2a0) passes the browser PATH test; #1372 (764a8da2) fails it; #1368 fails it (inherits).
- With this change: `server/browser-codex-path.integration.test.ts` and `server/drivers/codex.test.ts` pass together, 123 tests. `pnpm typecheck`, `pnpm lint` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when saving diagnostic dumps.
  - Temporary filesystem failures are now retried automatically, reducing failed writes.
  - Failed save attempts clean up temporary data before reporting the error.
  - Dump retrieval now waits for partially written files to become valid before reporting success or failure.


milind-soni:
CI on 8024a43a: macOS and ubuntu green (573 files; the browser-codex-path cases this PR fixes now pass). Windows attempt 1 was cancelled at the job time cap (Run tests 05:49→06:21). Attempt 2 (job 105099162307) failed a single unrelated case in comms.test.ts (legacy routine comms e2e › does not inject the agents integration into a depth-1 turn) with 543 passed; that suite uses the fake ACP fleet, which this PR does not touch. Attempt 3 is running now. #1385, #1386 and #1352 are waiting on this to rebase.

PR #1395, opened 2026-09-17
ci: shard the suite, one job per smoke, merge-queue gate, fake guard, pre-push checks

milind-soni:
## Why

Main was red for roughly 36 of the last 48 hours, and most of that time the failure a PR showed was not its own. Four causes, one fix each.

| Cause this week | Fix |
|---|---|
| A PR green alone turned main red when combined with what had just landed (#1356 vs #1316's tests; #1372) | `merge_group` trigger + queue-safe concurrency, so the merge queue can be enabled in the branch ruleset and every PR is tested on the real merge result before landing |
| Serial suite takes 17–19 min on one runner; slow runners hit the cap and read as red (three Windows "failures" with zero failing tests) | vitest in **three shards per OS** (~7 min each), with the `github-actions` reporter so failing assertions are annotations on the PR |
| The approval smoke was broken for a day and invisible, hidden behind any vitest failure in the same job | Every smoke is its own job: `packaged server smoke (os)`, `Windows CUA host smoke`, `Electron smokes (macOS)`. Typecheck, lint, locale check, Electron syntax check and UI build run once as `typecheck + lint` |
| The Codex fake imported `../atomic.ts` and died at link time where a test copies it out of the repo (#1372 → #1388) | `server/testing/fakes-self-contained.test.ts`: subprocess fakes import nothing from outside `server/testing`; a copied-out fake imports nothing relative |

Plus a **pre-push hook** (lint, typecheck, `i18n:check`; about a minute) that `pnpm install` wires up via `core.hooksPath`. It never fails an install (no git, CI, read-only checkout all no-op), `git push --no-verify` skips once, `OMB_SKIP_HOOKS=1` skips for a session.

## Required checks stay put

The branch ruleset requires `typecheck + test (macos-latest|ubuntu-latest|windows-latest)`. A `gate` job per OS keeps those exact names and aggregates every job above, so nothing has to change in settings for this to merge. When you update the ruleset to require the gate directly (or the individual jobs), the matrix can collapse to one job.

## What you need to do once this lands

1. Branch ruleset → enable **merge queue** for `main`. The trigger and concurrency are already in place.
2. Optionally re-point required checks at the new job names.

## Verification

- Workflow parses; jobs: `static, vitest, packaged-server, windows-cua, electron-smokes, gate` plus the untouched `cloudflared…, foss, control-plane, ui-smoke, package-linux, ios, android`. This PR's own run is the first real exercise of the layout.
- `pnpm exec vitest run --shard=2/3 --reporter=default --reporter=github-actions …` runs locally (vitest 4.1.10).
- Guard test: 7 passed; mutation check — re-adding `import "../atomic.ts"` to the Codex fake fails it with `fake-codex-app-server.ts imports ../atomic.ts from outside server/testing …`.
- Installer: in a throwaway repo, `CI=1` leaves `core.hooksPath` unset; plain run sets it; second run is a no-op. `sh -n` on the hook passes; `.gitattributes` pins it to LF.
- Gate script: passes on all-success; exits 1 naming the jobs on failure/skipped.
- `pnpm lint`, `pnpm typecheck` clean.

## Summary by CodeRabbit

- **New Features**
  - Added automatic Git hook setup after installation.
  - Added pre-push checks for linting, type validation, and translation consistency.
  - Expanded automated validation across supported platforms, including smoke tests.

- **Documentation**
  - Expanded contributor guidance with CI workflow details and instructions for reproducing four-way test shards locally.

- **Bug Fixes**
  - Improved hook compatibility across operating systems by enforcing consistent line endings.


milind-soni:
Two findings from the first runs of the new layout, both folded in:

- **Windows shard 3 failed one test out of 2,608**: `delta-context.e2e › gives a delegated return … while that turn is being set up` with "this test needs git on PATH". The fixture writes a node-shebang `git` and looks for a file named exactly `git` on PATH, so it is POSIX-only by construction (Windows has `git.exe`). It was added today (ec4cd6ed) and is the reason main's own Windows leg is red right now (run 35202286505). It is now `skipIf(win32)` with the reason beside it, like every other process fixture in the suite.
- **Shard balance**: vitest splits by file count, so the shard with the heavy e2e files ran 13 min (macOS/Ubuntu) and 19 min (Windows) against 4–7 for the others. Moved to four shards; caps 20/25.

Also fixed on the way: the phone-relay smoke forks `dist-server/index.js`, which the old single job got from the packaged-server step; the smokes job now builds it itself.

Timings from run 2 (three shards): typecheck + lint 1m · packaged server smoke 0–1m each · Electron smokes 1m · Windows CUA smoke 1m · shards 4–13m (mac/ubuntu), 8–19m (Windows). The old single leg was 19–35m.

PR #1508, opened 2026-09-17
fix(checkpoints): find a folder's checkpoints however Windows spells it

kevin9327:
## What changed

`server/checkpoints.ts` resolves a bot's working folder with `realpathSync.native` instead of `realpathSync` before it derives the shadow repo's key and hands the folder to git. The three entry points (`snapshot`, `listCheckpoints`, `restore`) now share one `canonicalWorktree()` helper.

Base commit: `897b21bbf3192b693088eed1b6d7608a4e0d7baa`.

## Why

A shadow repo lives at `DATA_DIR/checkpoints/<botId>/<sha256(cwd)[..16]>`, so the key is only as stable as the path string. `realpathSync` follows symlinks, but on Windows it hands back whatever spelling the caller passed in:

```
rs(C:\...\Temp\x\MyProject)   -> C:\...\Temp\x\MyProject
rs(c:\...\temp\x\myproject)   -> c:\...\temp\x\myproject      <- same folder, other key
rsn(c:\...\temp\x\myproject)  -> C:\...\Temp\x\MyProject
```

`c:\work\app`, `C:\Work\App` and the 8.3 alias are one directory on disk, so a folder that was typed one way and later picked another way lost its history: `listCheckpoints` returned `[]` and a restore answered *"no checkpoints exist for this folder"*, while a fresh shadow repo quietly started beside the old one.

`refusalReason()` a few lines above already resolves with `realpathSync.native` for exactly this reason (#1381), and `turn-resources.ts` does the same for workspace claims — this brings the storage key in line with the guard that protects it.

## How it was verified

Windows 11, Node 24, pnpm 10.33.0.

New test in `server/checkpoints.test.ts` (`it.runIf(process.platform === "win32")`, matching the existing Windows-spelling test in that file): snapshot a folder, then list / snapshot again / restore through `cwd.toUpperCase()`.

Fails on unmodified `main`:

```
FAIL  server/checkpoints.test.ts > snapshot > keeps one history however Windows spells the folder
AssertionError: expected [] to deeply equal [ { …(3) } ]
- [ { "at": 1789683399000, "hash": "7cb974b2…", "label": "turn 11111111" } ]
+ []
 ❯ server/checkpoints.test.ts:111:49
```

Passes with the change. Every changed line runs in that test (`snapshot`, `listCheckpoints` and `restore` all go through `canonicalWorktree`).

Suites touching this module, against a baseline on unmodified `main`:

| | baseline | with change |
| --- | --- | --- |
| `vitest run server/checkpoints.test.ts server/delta-context.e2e.test.ts` | 48 passed, 2 skipped | 49 passed, 2 skipped |

`pnpm typecheck` and `pnpm lint` both clean, same as the baseline. The full `pnpm test` (~30 min) was not run here, so that checkbox is left unticked.

## Screenshots (UI changes)

None — server-side only.

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — `pnpm typecheck`, `pnpm lint` and the affected vitest suites pass; the full `pnpm test` was not run

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved checkpoint consistency when accessing the same Windows workspace through different path spellings.
  - Checkpoint listings and snapshots now remain aligned across equivalent paths.
  - Restoring a checkpoint through an alternate path spelling correctly rolls back the workspace.
  - Prevented unchanged workspaces from being associated with duplicate checkpoint histories.


milind-soni (changes requested):
The canonical-path fix makes sense for newly created checkpoints, but the checkpoint directory is already keyed by a hash of the old realpathSync result. Changing that input without a legacy lookup/migration can hide existing checkpoint history for Windows paths whose casing or 8.3 spelling changes under realpathSync.native. The new test only creates history with the new algorithm. Please add an upgrade regression that seeds a pre-change shadow directory and verifies list/restore still finds it through the canonical path; retain or migrate that legacy history without overwriting a different checkpoint repository.

milind-soni:
Triage result: holding this merge. The switch to realpathSync.native changes the hashed shadow-repository path on Windows, but listCheckpoints has no lookup/migration for the previous path. Upgrading would make existing checkpoint history appear lost. The platform-independent tests pass (12; 2 Windows-only skipped), but they only cover new-format storage. Please preserve or migrate the legacy directory/key and add an upgrade regression with a checkpoint created under the previous key. No user data was touched during review.

PR #1510, opened 2026-09-17
fix(recall): a day named as session_search's until is the whole day

kevin9327:
## What changed

`server/recent-work.ts` gains `parseUntil()`, and the `/api/internal/session-search` route uses it for `until` instead of `parseSince`. A value that names a day now closes the window on that day's last instant; everything else is untouched.

Base commit: `897b21bbf3192b693088eed1b6d7608a4e0d7baa`.

## Why

`since` and `until` were both read by `parseSince`, which resolves a day name or a bare date to that day's **first** instant. The range clause is `m.at >= since AND m.at <= until`, so `until: "2026-09-16"` cuts the window off at 00:00:00.000 that morning — the day it names is almost entirely outside it.

The obvious way for a bot to ask what happened on one day is therefore an empty window:

```
session_search {"since": "today", "until": "today"}
→ "Nothing of yours is there since today until today — no message in any of your conversations in that window."
```

Nothing distinguishes that from a day on which nothing happened, so the bot reports a full day of work as no work. `since: "2026-09-14", until: "2026-09-16"` has the same shape: the 16th is silently missing.

Only the three forms that name a day are stretched — `today`, `yesterday`, `YYYY-MM-DD`. A span (`24h`, `3d`, `2w`), an epoch, and a date with a clock time (`2026-09-15 09:00`) still mean the exact instant they name, so `until` as a cursor keeps working. `since` is not touched: the first instant of a day is already what it should mean.

I left the tool's `until` description (`"Only messages up to this time; same forms as since."`) as it is, since it stays true — happy to spell the day rule out there too if you'd prefer.

## How it was verified

Windows 11, Node 24, pnpm 10.33.0.

`server/index.test.ts` — the existing `session_search by time` route test now also asks for one named day at both ends. Fails on unmodified `main` (test changes only):

```
FAIL  server/index.test.ts > bot memory API > session_search by time lists the bot's recent messages, its rooms included, and only its own
AssertionError: expected false to be true // Object.is equality
 ❯ server/index.test.ts:9188:134
    9187|       const oneDay = (await (await search({ since: "today", until: "today" })).json()) …
    9188|       expect(oneDay.hits.map(…).some((text) => text.includes("Please reconcile the September invoices"))).toBe(true);
```

`server/recent-work.test.ts` — two unit tests for `parseUntil`: the day forms close on 23:59:59.999 local, and spans / epochs / explicit clock times are returned unchanged. Both changed lines (`parseUntil`'s body and the route's call) run in those tests.

Suites touching this code, against a baseline on unmodified `main`:

| | baseline | with change |
| --- | --- | --- |
| `vitest run server/recent-work.test.ts server/index.test.ts` | 233 passed, 1 skipped | 235 passed, 1 skipped |

`pnpm typecheck` and `pnpm lint` both clean, same as the baseline. The full `pnpm test` (~30 min) was not run here, so that checkbox is left unticked.

## Screenshots (UI changes)

None — server-side only.

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — `pnpm typecheck`, `pnpm lint` and the affected vitest suites pass; the full `pnpm test` was not run

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session search date filtering so named days such as “today,” “yesterday,” and specific calendar dates include the entire day.
  - Corrected upper-bound date handling to include events through 11:59:59.999 PM on the selected day.
  - Preserved precise behavior for explicit times, date ranges, and epoch values.


milind-soni (approved):
Reviewed the focused diff and updated with current main to include the existing browser/iOS fixture repairs. Independently ran isolated HTTP regressions on the updated head: whole-day recall search plus date-boundary tests pass. No live user data touched. Merge remains conditional on required CI.

milind-soni:
The refreshed head e76e6602 is locally qualified: 5 focused recall tests pass, including a real isolated HTTP search. Normal auto-merge is enabled subject to required CI.

PR #1514, opened 2026-09-18
fix(comms): name the person who sent a message, not the workspace profile

milind-soni:
Came out of "where did Milind come from?" — a Chief relayed a teammate question to another bot as *"Milind ne pucha hai"*, naming someone who hadn't sent it. The receiving bot then answered the wrong question, and the Chief relayed that back.

## The cause

**The server authenticates per person but attributed per workspace.**

A session carries real identity — `email`, and the device `label` chosen while pairing ([sessions.ts:88-97](server/sessions.ts#L88-L97)) — and a workspace can hold many people (`signIn.admins` / `signIn.members`, plus every paired device).

But `WireMessage` had no human sender. The only `from` is `{ botId, name, color }`, for bots ([wire.ts:304](shared/wire.ts#L304)). So every user turn rendered as `${cfg.profile.name}:` ([index.ts:7749](server/index.ts#L7749)), and rooms stated it outright:

> Room members: …, and **&lt;name&gt;** (the human).

Everyone collapsed into whoever Settings named. Bots address the wrong person, relay their questions under someone else's name, and write memories about work that person never did. On a shared or hosted instance that's a privacy problem as much as a cosmetic one — and it's invisible to the people involved, because the bot is faithfully repeating what the harness told it.

Worth noting the relay wrapper itself was always correct — `peerMessageText` names the *sending bot* and says "not from your user". The wrong name came from the transcript the Chief was reading.

## The change

User messages now record who sent them; the transcript and the exported markdown render that instead of the profile name.

**Loopback stays unstamped on purpose.** It's the desktop owner by design, so a single-user setup reads exactly as it does today — as does every message written before this existed. The field is optional and `Message = WireMessage` has no projection guard, so this is backwards compatible both ways.

Covered: the direct send, a room send, and **steering into a running turn**. That last one appends on its own path and was the gap this PR's first draft missed — the test caught it, because the owner's message came back `steered: true` and my "owner is unstamped" assertion had been passing for the wrong reason.

## Not covered (deliberately)

Queued sends that drain later still fall back to the profile name. Their queue records are durable and live in another module, so threading identity through them is a follow-up rather than a quiet half-change.

The room roster line still names the profile owner. Listing actual people there wants the same follow-up.

## Verification

- `tsc -b`, `tsc -p tsconfig.server.json`, `oxlint --deny-warnings` clean.
- New e2e test in the real isolated-server harness: pairs a device the way a phone does, sends as that person, steers as that person, and sends as the owner against a *second* bot so the owner's message cannot be folded into the first turn.
- Mutation-checked three ways — never stamping, stamping the owner too, and dropping the steered stamp each fail the test.

## Summary by CodeRabbit

* **New Features**
  * Messages sent from paired devices now display the paired person’s name.
  * Sender attribution is preserved while steering active conversations and in transcript or Markdown exports.
  * Desktop owner messages continue to use the existing profile attribution.
  * Messages without paired-device attribution continue to display the profile name.


milind-soni:
Follow-up commit: the new test now cleans up after itself.

It started real fixture turns and left two bots and a paired session alive. This suite shares one isolated server, so that showed up as *other* tests failing much later — the local run went from the usual handful of flakes to 27 failures, in tests nowhere near this change. It now interrupts and deletes both bots and revokes the session in a `finally`.

Worth recording what the local numbers actually say, since `server/index.test.ts` is unstable on my machine and I don't want the CI result read against a bad baseline:

| run | failures |
|---|---|
| clean `main`, no changes | **6** |
| this branch, my test skipped | 8 |
| this branch, test without cleanup | **27** |
| this branch, test with cleanup | 8 |

Each run fails a different, barely-overlapping set — including on unmodified `main` — so the 6–8 band is environmental here (this machine has been running heavy suites all session), while the 27 was a genuine regression from the missing cleanup, and it's gone.

The targeted verification is the part I'd trust: the new test passes on its own, and all three mutations still fail it (never stamping, stamping the owner too, dropping the steered stamp). CI is the authority on the rest.

PR #1515, opened 2026-09-18
test(server): poll the fake CLI dump until its JSON is whole

bradhallett:
## Problem

Four tests in server/index.test.ts waited for the fake CLI dump with an existsSync poll and then read it with JSON.parse(readFileSync(...)). On a loaded CI runner the file can be observed before its JSON is fully written, so the parse throws a syntax error and the test fails — most recently the standing-instructions (soul) test at index.test.ts:6464.

## Change

Convert the four remaining sites to the file's existing readJsonFileWhenReady helper (bounded parse-poll, 15s, matching the adjacent already-converted sites):

- mounts the verification skill into a real turn
- injects the bot's standing instructions (soul) into a real turn
- re-enters setup mode for a configured bot on /setup
- never lets an out-of-band SOUL.md edit reach the prompt

The existence-only polls that never read the file are unchanged.

## Verification

- pnpm typecheck
- the four touched tests pass; the rest of the suite is unaffected

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability when waiting for generated configuration data.
  - Tests now wait for valid JSON content before continuing, reducing failures caused by partially written files.
  - Extended the wait period for slower environments and preserved validation of skill setup, standing instructions, setup flows, and SOUL.md behavior.


milind-soni (approved):
Reviewed the focused diff and updated with current main to include the existing browser/iOS fixture repairs. Independently ran isolated HTTP regressions on the updated head: all five matched skill/soul/setup prompt integration tests pass. No live user data touched. Merge remains conditional on required CI.

bradhallett:
CI note: the two vitest failures on the updated head (`thread-aware-bots.e2e.test.ts`, capacity 1 and capacity 2) reproduce on `main` itself — run 35598240945 for e87f4880 fails the same two tests with the same assertion (`expected 'Hand the pull requests to QA.' to contain 'Your downstream room requests have se…'`). This branch's diff against `main` is only the `server/index.test.ts` dump-poll hardening, so these failures look inherited from `main`, not caused here.

bradhallett:
Follow-up on the inherited failures documented above: no fix was in progress upstream, so I opened #1613 with a test-only fix.

Root cause is #1610 (80acbc3b): a source waiting on teammate results is now idle (`waitingForTeammates: true`) instead of busy, and accepted handoffs dispatch immediately instead of waiting for the source turn to end. `thread-aware-bots.e2e.test.ts` still assumed `busy === false` meant the resumed review turn had finished (line 407-408, both capacity tests), and the deleted-destination test raced the immediate dispatch when deleting the queued task (the Windows-only 409).

#1613 waits for actual settlement in both capacity tests and holds the recipient's single thread slot so the deleted destination is deterministically queued. The full file passes 10/10 locally; typecheck and lint are clean.

PR #1516, opened 2026-09-18
test(server): retry browser-runtime self-recovery inside a bounded window

bradhallett:
## Problem

The self-recovery test's echo call after a transport timeout can itself blow the 60ms budget on a loaded runner (browser-runtime.ts:123), so the fresh spawn's echo rejects with a timeout and the test fails even though recovery works (CI run 35285379867, shard 2/4).

## Change

Retry the recovery echo inside a bounded 10s window (well inside the 20s global test timeout), tolerating only transient timed-out rejections: each timed-out attempt kills only its own transport and the gate keeps resolving itself. Any other error — for example the interrupted-browser-action regression this test guards against — still fails fast with its diagnostic intact.

## Verification

- pnpm typecheck
- pnpm vitest run server/browser-runtime.test.ts — 22/22 green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when temporary connection timeouts occur, with limited retries for more reliable requests.
  * Request timeouts reported by the engine are now surfaced accurately and handled separately from connection failures.

* **Tests**
  * Expanded coverage for recovery success, timeout handling, and immediate failure behavior.


bradhallett:
@coderabbitai review

milind-soni:
Closing because main now deliberately requires an explicit browser restart after an uncertain accepted daemon action (#1555). This PR retries the previous self-recovery contract, which would regress that behavior. The loaded-runner timeout budget is preserved in the updated #1520; the iOS queued fixture already landed in #1519. No branch deletion.

PR #1518, opened 2026-09-18
fix(ci): give the Electron relaunch test a session bus on headless Linux

### LINKED ISSUE #1512: CI (ubuntu vitest job): Electron relaunch vendor-updater test dies to "Failed to connect to the bus" on headless runners
## What is failing

In the ubuntu-latest vitest job, with the vitest step itself fully green (143 files, 2093 tests), the Electron node-test step fails on run 35285379867 (fork PR head 203d6ade, job "vitest (ubuntu-latest, shard 1/4)"):

```
test at electron/vendor-updater.node-test.mjs:246:1
✖ Electron relaunch waits for deferred cleanup and the replacement acquires the same data directory (12095.32721ms)
  AssertionError [ERR_ASSERTION]: [11289:0917/231237.987738:ERROR:dbus/bus.cc:405] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
  [11289:0917/231238.123511:ERROR:dbus/object_proxy.cc:572] Failed to call method: org.freedesktop.DBus.NameHasOwner: object_path= /org/freedesktop/DBus: unknown error type:
  actual: null,
  expected: 0,
  operator: 'strictEqual',
```

https://github.com/milind-soni/OpenMausBot/actions/runs/35285379867

The relaunched Electron child dies to a signal on a headless runner with no session bus, so the exit-code assertion sees `null` instead of `0`. This looks like a runner-environment gap rather than a product regression.

## Possible fixes

Either export a session bus address for the job (for example via `dbus-launch`), or skip the relaunch exit-code assertion when no session bus exists. Happy to send a PR for the second option if maintainers prefer a repo-side guard.

bradhallett:
Supersedes the earlier skip-based approach on this branch with the root-cause fix (identical content to #1545, the stack copy).

## Problem

Fixes #1512. The Electron relaunch vendor-updater test launches real Electron under xvfb-run, and app.requestSingleInstanceLock() needs a DBus session bus on Linux. The headless ubuntu runners have no bus, so the fixture died to Failed to connect to the bus and the assertion failed on every run.

## Change

The test probes for an inherited DBUS_SESSION_BUS_ADDRESS or a dbus-run-session binary (the pattern scripts/run-linux-package-smoke.mjs already uses). With the tool available it launches under dbus-run-session -- xvfb-run -a; with an inherited bus nothing changes; with neither it skips with a logged reason instead of failing. Assertions and the embedded fixture are untouched.

## Verification

node --test electron/vendor-updater.node-test.mjs: 10/10 on the direct macOS path; ubuntu CI gets its bus where dbus-x11 is already installed for the smoke scripts.

## Summary by CodeRabbit

- **Tests**
  - Improved Linux application relaunch test handling when no session bus is available.
  - Tests now use an available session-bus runner when possible and continue supporting virtual display environments.
  - Tests skip with a clear explanation when the required session-bus setup is unavailable, reducing misleading failures in unsupported environments.


milind-soni (approved):
Reviewed the focused Linux test-harness fix. A disposable session bus is supplied only when the environment lacks one; existing desktop sessions and non-Linux behavior remain unchanged. Linux and Windows shards and packaging checks are green. Remaining required CI must pass before merge.

PR #1524, opened 2026-09-18
fix(acp): read the --version banner from stderr when stdout is empty

anhybear:
## What

Hermes Agent 0.19.1 prints its whole `hermes --version` banner to **stderr** and nothing to stdout. The ACP snapshot probe only read stdout, so a perfectly good Hermes install showed up as **unavailable — "`hermes` CLI not found"** in Engines.

Prefer stdout; fall back to the first stderr line; `null` only when both are empty. The check is extracted into a small exported `versionFromProbe(stdout, stderr)` helper with a unit test.

## How to reproduce

```
$ hermes --version >/dev/null; echo $?   # 0
$ hermes --version 2>/dev/null | wc -c   # 0  ← the banner is on stderr
```

Before: Engines shows Hermes unavailable. After: `Hermes Agent v0.19.1 (…)` and Hermes bots run.

## Checks

- `pnpm exec tsc -p tsconfig.server.json --noEmit` clean
- `vitest run server/drivers/acp` — 49 passed (new `version-probe.test.ts` included)
- oxlint clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version detection for command-line tools that report their version through standard error.
  * Correctly handles CRLF, LF, and CR line endings when reading version information.
  * Prevented valid tools from being incorrectly reported as unavailable when standard output is empty.

* **Tests**
  * Added coverage for version output from standard output, standard error, and empty responses.
  * Added regression coverage for multiple line-ending formats.


milind-soni (approved):
Reviewed stdout precedence, stderr first-line fallback and CRLF/CR handling. All four focused version-probe regressions passed locally. Nonzero CLI exit still reports unavailable. Ready after required CI.

milind-soni:
The refreshed head 6b5e06f0 is locally qualified: 96 ACP/version-probe tests pass, with stderr-only version banners covered. Normal auto-merge is enabled, subject to required CI.

PR #1527, opened 2026-09-18
fix(desktop): hydrate and switch threads by page

nelson872:
## What happens today

Two OpenMausBot instances paired over a Cloudflare tunnel: `GET /api/health` answers `200`, and the desktop client never gets past "Connecting to the bot server…". The companion answers `502 {"error":"the response from OpenMausBot was too large"}` (`companion/src/proxy.ts`, the `MAX_JSON_BODY_BYTES = 32 MB` guard) and `loadAll()` in `src/state/store.tsx` retries forever, because hydration cannot start without `/api/bots`.

The response is that large because the desktop client asks for every message of every thread. On the instance I hit this on, one room held 54,307 messages and the snapshot serialised to ~42 MB.

## Why it started now

`?messages=` and `/api/threads/:id/messages?before=` have been on the server for a while; the desktop client never used either. What changed is how much each message weighs:

- `2ef52bee` ("feat(ui): polish settings, attachment galleries and tool details", first released in v0.1.77) added `toolDetailPreview()`, so an `activity` message now stores `tool.input` and `tool.output` (up to 6,000 characters each, redacted before truncation).
- Measured on that room's database, before and after that release: `activity` rows grew from 0.24–1.66 MB/day to 3.6–5.0 MB/day. Of the newest 5,018 activity rows, `tool.output` alone is 12.57 MB.

That is a UI feature working as designed; it just moved the snapshot past a limit the client was already living dangerously close to. The mobile apps were unaffected — they already page.

## What changed

Desktop client:

- Hydration asks for a bounded page: `GET /api/bots?messages=200`.
- Opening another thread asks for the same page instead of the whole transcript.
- The two `/api/bots` calls that only reconcile bot settings are bounded by the same constant.
- The transcript's "Show earlier" pill, once the locally held page is fully mounted, becomes "Load earlier messages": it fetches the previous page with `/api/threads/:id/messages?before=<oldest held id>` and prepends it. The existing height capture keeps the viewport still, so the row under the cursor does not move.

Server:

- `POST /api/bots/:id/tasks/:threadId` and `POST /api/groups/:id/tasks/:threadId` accept `?messages=N` and answer a bounded page, the same shape `/api/bots?messages=N` already returns. `messages=0` keeps its existing meaning (settings only, no `messages` key) and no parameter still answers with the whole transcript, so no existing client changes behaviour. The parameter is parsed before the switch, so a rejected value cannot leave the thread moved.

No storage, retention or redaction change: this is transport only, and no message is modified or removed anywhere.

## Screenshots (UI changes)

The room holds one page (200 messages); the local window is already fully mounted, so the pill offers the previous page. Oldest row: `probe 250`.

image

After one click: the previous page is prepended, the oldest row is now `probe 50`, and the row the reader was on keeps its place (scrollTop moves by exactly the height that was added).

image

## How it was verified

Measured on an isolated fixture (`node --experimental-strip-types scripts/control-omb.ts launch`, Windows 11, Node 24.19) with a 2,000-message room:

| | before | after |
|---|---|---|
| snapshot response | 5.05 MB | 0.88 MB |
| thread switch response | 4.39 MB | 0.44 MB |
| paging back | not available | 9 pages to the top, all 2,000 recovered |

Driven in a browser against that fixture: the room opens holding 200 messages, "Show earlier" exhausts the held page, the pill becomes "Load earlier messages", and one click takes the transcript from 200 to 400 rows with the oldest row moving from `probe 1800` to `probe 1600`.

- `pnpm typecheck`, `pnpm lint`, `pnpm i18n:check` clean.
- `pnpm exec vitest run src` — 1,789 passing.
- New tests: `server/index.test.ts` ("message pages") covers the bounded switch for a channel and a bot, the unchanged `messages=0` and no-parameter shapes, and the rejected parameter; `src/state/store.test.ts` covers the prepend (overlapping ids kept once), the in-flight flag, and how a payload carrying a transcript answers `hasMore`.

## Known follow-ups, not in this PR

- A room task switch still broadcasts the whole transcript over SSE (`broadcast({ kind: "group", group: fresh })`). The client ignores it for anything but a switch, but it is still sent.
- A bounded page strips `png` from `screen` messages and sets `hasImage`; the desktop transcript does not yet fetch those from `/api/threads/:id/messages/:id/image`. It already dropped older frames' pixels client-side (`MAX_KEPT_SCREEN_FRAMES`), so this is not a regression.

## Checklist

- [ ] `pnpm typecheck` and `pnpm test` pass locally — typecheck, lint, `i18n:check` and `vitest run src` are clean here. The full `server/index.test.ts` suite fails 11 tests on this Windows machine; 10 of the same tests also fail on an unmodified checkout of `main` (Box token, VPS and Chief-room tests), and the eleventh is a 20 s timeout that did not reproduce. Happy to dig into any of those if they are not already known.

## Summary by CodeRabbit

- **New Features**
  - Added pagination for conversation history, allowing users to load older messages on demand.
  - Added “Load earlier messages” controls for chats and group rooms, with loading states and preserved scroll position.
  - Conversation switching now supports limited message history while retaining access to complete transcripts when requested.

- **Bug Fixes**
  - Prevented duplicate requests and stale results when loading older messages.
  - Invalid message-history requests now return an error without switching conversations.
  - Preserved scroll position correctly during conversation and room changes.

- **Tests**
  - Added coverage for pagination, validation, loading behavior, deduplication, and history state handling.


milind-soni (changes requested):
Two issues in the task-switch handlers need fixing before merge: (1) pageSize validation happens after switching and broadcasting, so an invalid messages parameter returns 400 after changing the selected thread. Validate before mutation and test the active thread stays unchanged. (2) both handlers still construct the full transcript with botWithThread/groupWithThread and broadcast it before returning the bounded HTTP page. This still sends the entire history over SSE (and can overwrite the bounded client snapshot), undermining the memory/network benefit. Bound the event payload consistently and add an isolated HTTP+SSE regression proving a paged switch does not emit the full history.

nelson872:
Thanks — both addressed in `cb487cbd`.

**Validation before mutation.** You are right, and the test I shipped with it was worse than useless: it asked for the thread that was already active, so it passed either way. `pageSize` is now parsed before `store.switchTask` / `store.switchGroupTask` in both handlers, and the test parks the channel on its other thread first, then asserts the refused parameter left it there. (The reorder was in my working tree but never made it into the first commit — my mistake, not a disagreement.)

**Switch frame.** Both handlers now build the settings shape once and broadcast it with `messagePage(threadId, DEFAULT_PAGE)` and `hasMore`; only the unparameterised HTTP branch still calls `botWithThread` / `groupWithThread`, so a paged switch no longer materialises the whole transcript at all, let alone sends it. The frame stays what it is — the thread moved — and anything earlier pages back through `/api/threads/:id/messages`. The new regression `keeps the switch frame bounded, so a paged switch never emits the whole thread` seeds a 60-message room, opens `/api/events`, performs a paged switch, and asserts the transcript-carrying frame is at most one page with `hasMore: true`, ends on the newest message, and that no frame in the stream carried the full thread.

One deliberate consequence, in case you would rather scope it differently: the frame is bounded for every switch, not only paged ones, so a client that asks for `?messages=0` also stops receiving a whole transcript over SSE. Android already folds `hasMore` off a bot/room frame (`core/Store.kt`), so that path reads a bounded frame correctly. Happy to key the frame's size off the request instead if you prefer the old shape for unparameterised callers.

**CodeRabbit's scroll capture.** Fixed too: the captured height now carries its `transcriptKey`, is dropped when the key no longer matches, and `transcriptKey` is a dependency of both restore effects, so a switch runs the effect that drops a stale capture.

Locally: `typecheck`, `lint` and `i18n:check` clean, `vitest run src` and the `message pages` suite pass.

milind-soni (changes requested):
The prior blockers are fixed: validation now precedes switching and SSE is bounded. I independently ran all 14 message-page HTTP tests plus 127 store/recovery tests successfully. One remaining paging race needs a regression: loadOlderMessages captures only threadId/before, and olderMessages blindly prepends into any current owner of that thread. If an edit/branch switch rewinds the same thread while the request is in flight, its old-branch page can subsequently reintroduce messages from the abandoned branch (deduplication does not validate ancestry). Fence page responses against the transcript/branch generation that issued them, discard obsolete responses, and test delayed page completion across a same-thread rewind/branch switch. This must preserve ordinary streaming appends and clear loading state on discarded responses.

nelson872:
Fenced in `8aba187c`.

Each thread carries a generation in the store, bumped when its transcript is replaced or its visible branch moves: `threadActive` (an edit, a rewind, another client's switch), `switchBranch`, `taskSwitched`, a room payload that carries a transcript, and a hydrate, which resets the map. `loadOlderMessages` records the value it asked under — it does not move the generation itself, so the value read when the request goes out is the one the answer is judged against — and the reducer drops a page whose generation has moved on, clearing `loadingOlder` on the way out so the pill never stays spinning. Failures take the same path.

Ordinary appends deliberately do not bump it, so a page still lands over messages that arrived while it was on the wire — that is the common case, and it is one of the two new tests: `drops a page that was in flight across a rewind, and stops the spinner` (the rewound thread keeps the rows it held, `hasMore` is untouched, `loadingOlder` is empty) and `still lands a page over messages that arrived while it was on the wire`.

You are right that the id dedupe could not have caught this: the rows an abandoned branch would reintroduce are exactly the ones this client does not hold.

Local: `typecheck`, `oxlint --deny-warnings server src` and `vitest run src/state` clean.

On CI: `vitest (windows-latest, shard 2/4)` is red on `server/browser-runtime.test.ts > retires an idle MCP client without killing its browser descendant` with `kill ESRCH`, for both the EOF true and false cases. This branch does not touch the browser runtime or process handling, so I read it as a Windows flake rather than a regression from here — happy to look into it separately if it is not already known.

PR #1536, opened 2026-09-18
fix(android): recover from event streams that close before hello

aivsomkar:
## What changed

When an HTTP 200 event stream closes before its initial `hello`, Android now tries the next authorized route after its normal backoff. Previously it retried the same empty stream indefinitely, leaving “Lost the connection” visible even when an advertised protected fallback worked.

A stream that already connected continues reopening its working route with its cursor, including when a later chunk is truncated. Credential protection, deliberate disconnect and unauthorized handling retain their existing policy.

Relates to [MOCA-179](https://linear.app/supamaus/issue/MOCA-179/android-app-frequently-loses-connection). This fixes a reproduced persistent-offline path; the reporter's approximately hourly trigger has not been reproduced on a physical phone, so this does not claim to resolve every connection loss in that report.

## Verification

- Reproduced on base `0c327bca`: the new virtual-time fallback assertion failed, and a real HTTP/SSE loopback fixture timed out without reaching the healthy fallback.
- All **546 Android core tests and 914 app tests passed**, with no failures or skips.
- The five added regressions cover early closure, normal post-hello reconnect, post-hello truncation, protected fallback selection and recovery through the production Session/HTTP/SSE stack without restarting.
- Preview APK built; its v2 signature verified.
- Push-time lint, typecheck and locale checks passed; independent review found no remaining actionable findings.
- Fixtures use disposable servers, in-memory pairing stores and synthetic tailnet names pinned to loopback. No installed Tailscale, external network or user data is used.
- The broader repository suite was not rerun for this Android-only change. The earlier queue run at the same base found a delegate-replay failure reproduced on pristine main and an Electron proxy timeout that passed on rerun, documented in #1534. This PR does not claim a green full repository suite.

Recipe and evidence: `docs/verification/android-stream-recovery.md`. There are no UI or server changes.

## Checklist

- [x] Android core and app suites pass
- [x] Preview APK builds and signature verifies
- [x] Regression fails before the fix; isolated transport fixture recovers after it
- [ ] Full repository suite green (known baseline findings above)
- [x] No release, merge or user-data mutation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android stream recovery when a connection closes before initialization completes.
  * Automatically retries through the next authorized fallback route instead of repeatedly reconnecting to an unavailable route.
  * Preserves the active route and stream position when reconnecting after an established stream is interrupted.

* **Documentation**
  * Added verification guidance for Android stream recovery and fallback behavior.

CI follow-up, 2026-09-19: the [iOS job](https://github.com/milind-soni/OpenMausBot/actions/runs/35374449862/job/105695719931) fails the pre-existing weekend queue fixture assertion in ThreadNavigationUITests.swift:142. [PR #1519](https://github.com/milind-soni/OpenMausBot/pull/1519) supplies the missing queued-send fixture (also present in stack PR #1546). This Android patch does not alter that fixture. CI is not claimed green; the separately owned fixture correction has not been copied into this PR.


milind-soni (approved):
Reviewed updated head d73d6f74 and ran the complete Android core test suite successfully in an isolated worktree, including the real loopback HTTP/SSE recovery test. The patch distinguishes pre-hello closure from an established stream ending, preserves cursor-based reconnects on the working route, and retains protected-route downgrade restrictions. This proves the reproduced early-close failure, not every possible mobile disconnect. Required CI must pass before merge.

PR #1539, opened 2026-09-18
refactor(server): make thread and bot deletion durable and restart-safe

bradhallett:
## Problem

CodeRabbit flagged that the store's deletion paths could lose cleanup work. deleteGroup rolled back an in-memory snapshot, deleted transcripts before persisting the group removal, and never used the pending-deletion journal; deleteGroupTask and deleteTask mutated live records with no restore on failure; a retried deleteBot returned at its record guard before workspace, skill-state and folder cleanup ran; the pending-deletion journal read swallowed damaged files as empty; the message cache mutated before persistence; and pending deletions survived restart unless the same delete was re-invoked.

## Change

- deleteGroup flushes its journal key at entry, stages tombstones for owned threads, persists the group removal before deleting transcript records, and clears the journal only after every cleanup succeeds.
- deleteGroupTask and deleteTask snapshot the live fields they mutate and restore them on failure; deleteTask stages its tombstone before any call that can save.
- deleteBot keeps a durable bot-cleanup journal entry covering the workspace, skill-state and bot folder; retries with the record gone finish the cleanup and emit bot.deleted.
- setChiefOfStaff computes the next records, saves them, and only then mutates the live records, so a failed save leaves memory untouched.
- pendingThreadDeletions returns an empty map only on ENOENT and rethrows damaged-journal errors; the message cache updates only after mdb.appendMessage persists.
- The Store constructor drains every pending deletion after migrations, guarding owners that still exist.

## Verification

pnpm lint, pnpm typecheck, and the store, group-tasks and tasks suites pass (247 targeted tests; full suite green).

---
Saved head: 83071f0bc06e7c765d86c39f9b1cbfdfd28fba4c.


bradhallett:
@coderabbitai review

PR #1540, opened 2026-09-18
refactor(state): settle task writes on execution and harden routing

bradhallett:
## Problem

Three client-state gaps CodeRabbit identified: a task-write lane was deleted when its masked save chain fulfilled even though the batch execution rejected, so a later send never observed the failed settings write; a refreshed consumed-queue receipt kept its old insertion position so overflow eviction could drop a live receipt; and a notification for an unknown bot dispatched a select for an id not in state.

## Change

- persist() settles the lane on the batch execution promise: the lane clears (and reports the authoritative bot) only when every queued write fulfilled; rejected executions route through the existing reconciliation path.
- rememberConsumedQueueId deletes an existing key before re-adding it, so refreshed receipts move to the end of the eviction window.
- openNotificationTarget returns early when the bot is unknown and dispatches only a resolved bot.id.

## Verification

pnpm lint, pnpm typecheck, and the src/state recovery and store suites pass; the recovery fixtures hand each reconciliation a fresh response, matching the new two-reconcile flow.

---
Saved head: 70e8c78d2d0fd00d1d9f0397a2eb50e0d668c8cf.


bradhallett:
@coderabbitai review

PR #1543, opened 2026-09-19
refactor(server): inject the login stop grace and widen its test budgets

bradhallett:
## Problem

The server-owned Claude sign-in tests pinned a 1s startup budget and a hard-coded 1.5s stop grace. On a loaded runner a cold fixture CLI can take longer than either, so four tests failed intermittently with the startup-timeout message, the still-reports-sign-in message and an ENOENT on calls.jsonl, then passed warm.

## Change

- ClaudeLoginOptions gains stopGraceMs (default 1500, so production behavior is unchanged) and stopChild uses it instead of the literal.
- The test harness create() passes startupTimeoutMs 10s, lifetimeMs 30s, completeTimeoutMs 15s and stopGraceMs 10s; the one test that pins the startup timeout keeps a local 1s override.
- Fixture-file polls guard with existsSync so a cold boot polls instead of throwing ENOENT.

## Verification

vitest run server/drivers/claude-login-auth.test.ts: 15/15; typecheck clean.

---
Saved head: 54804bae562fb07f6b9d795857c18dd270fec350.


bradhallett:
@coderabbitai review

Issue #527, opened 2026-08-27
context usage and tracking


pandas9:
### Before filing

- [x] I searched existing issues and this isn't a duplicate

### Area

Chat UI (src/)

### What happened

100k tokens spent on 5 "test" messages, is that accurate?

<img width="1119" height="869" alt="Image" src="https://github.com/user-attachments/assets/3e1f8741-0096-4a58-a483-bf1714c528dc" />

### Version and install

OpenMausBot v0.1.37 linux

### Logs

```text

```

milind-soni:
Short answer: yes, that figure is accurate — but it needed a breakdown to be believable.

Every turn resends the **whole conversation plus the system prompt and every MCP tool schema** to the model. On a Codex bot with computer/connected-apps tools mounted that's ~15–20k tokens of reading per turn, so five "test" turns really are ~88k tokens *processed* — the model re-read the same context five times. Most of that comes from the provider's prompt cache (cheap on a metered key, free on a subscription), but it's still tokens the model read, which is what the chip counts.

What was wrong is that the chip gave you a total with no way to see this. https://github.com/milind-soni/OpenMausBot/pull/528 makes drivers report the cached share and shows the split in the tooltip and the Usage card — e.g. `88.2k in (79k cached) · 1.2k out` — with a caption saying what cached means. The headline number stays the same on purpose: hiding the cached share would understate what the model actually processed.

pandas9:
accidentally closed this one, sorry about that

PR #1557, opened 2026-09-19
fix(usage): headline the tokens actually bought, not the thread re-read each turn

aivsomkar:
## What

The usage screens headline `input + output`. Every engine's `input` **includes the
provider's cache reads**, so that sums the same system prompt and conversation once
per turn.

A six-turn Claude thread that actually cost **$0.2843** reported **248.3k tokens
"used"** — 177.4k of which was one cached prefix counted six times.

That is the number people look at to judge what a bot costs, and it is off by 3.5x.

## Fix

`freshTokens` already existed for exactly this, and its comment cites #527. Only
`ChatView` used it. This adds `headlineTokens` — fresh tokens where the engine
reported a cached share, the raw total where it did not (the only honest answer
without one) — and moves the remaining five surfaces onto it.

| | tokens shown |
|---|---|
| before | 248.3k |
| after | **70.9k** |
| actually spent | $0.2843 |

## Deliberately unchanged

The context indicator (`Context N of window`) still counts cached tokens, and must.
Caching is a **price discount, not a space saving** — the model re-reads the whole
prefix every call and it fills the window either way. Only the *cost-shaped* figure
was wrong.

## Test plan

- `pnpm exec vitest run src/lib/usage.test.ts` — 13 pass, including a new case
  written red-first against a 1.9M-read/1.86M-cached thread
- `pnpm typecheck` — clean
- `pnpm exec oxlint --deny-warnings src/` — clean
- Verified against a live `/api/usage` payload from a real six-turn Claude thread
  (built side-by-side as OMB5, isolated data dir): 248.3k -> 70.9k

## Platforms

|---|---|---|
| macOS | yes | in this PR, verified against a live usage ledger (OMB5) |
| Windows | yes | in this PR (desktop UI, no platform gate touched); NOT built on Windows |
| iOS | n/a | no usage screen or token total anywhere in `ios/` |
| Android | n/a | no usage total anywhere in `android/` |
| Companion | n/a | no route changed; no API surface in the diff |

## Summary by CodeRabbit

* **Bug Fixes**
  * Token usage displays now prioritize newly purchased tokens when cached usage is known, providing more accurate totals.
  * Updated token counts across chat, task selection, usage history, and bot settings.
  * Bot sorting now uses the updated token metric when costs are equal.
  * When cached usage data is unavailable, displays continue to use combined input and output tokens.


milind-soni:
Reviewed and cleaned up at 5a92f1b9. Usage-section, task-row and task-switcher tooltips now reuse the cached-input breakdown so they explain the reduced headline rather than showing an unexplained larger total. Added a separate localized switcher template, preserving existing translation placeholders. 26 usage/TaskPicker tests, typecheck, focused lint and all locale checks pass. Normal auto-merge is enabled subject to required checks.

PR #1577, opened 2026-09-19
test: make delegation, fixture and UI waits deterministic under load

bradhallett:
Several suites failed intermittently under parallel load, both locally and on CI. Each failure below was reproduced, root-caused, and fixed at the source rather than by adding retries.

## Install and login fixture races

1. **Fake npm raced its own boot.** The fixture armed its `SIGTERM` trap only after ESM imports and a log write, so under load the 300ms TERM killed it before the trap existed and the "force-stops an install that ignores TERM" assertion saw `SIGTERM` instead of `SIGKILL`. The script is now CommonJS with the trap as its literal first statement, and the test timeout has headroom (5s) so TERM lands well after boot.
2. **Device-auth startup timeout was 1s.** Under parallel load the fixture regularly took longer than that to print its link, so "shows the link…" failed. The default is now 5000ms; the cancel test pins 1000ms explicitly so it still exercises the timeout path.
3. **An unguarded `expect.poll` read `calls.jsonl` before the fixture created it.** `expect.poll` retries failed assertions but propagates thrown errors, so the ENOENT failed the test instead of retrying. The poll now catches a missing file, matching the neighbouring polls.

## MCP shutdown grace (Windows teardown)

`chat-mcp-tools.ts` passed `killCliTree(child, 500)` — the only sub-second call site in the codebase (`server/procs.ts` default is 5s). On a loaded Windows runner, `taskkill /T` can exceed 500ms, so teardown intermittently failed with "MCP server shutdown could not be confirmed". The grace is now the codebase default, with a comment explaining why.

## Renderer e2e races

- **thinking-timer**: clicking the "Expand Pepper threads" chevron by accessibility name intermittently failed with "2 elements are named …" — right after a mutating POST the sidebar snapshot briefly lists the chevron twice. The helper clicks the DOM node directly instead of resolving through the AX snapshot.
- **team-computers**: clicks dispatched during view transitions could land on a node the re-render replaced, leaving the panel never opened. Each step now waits for the settled view. Additionally, a deterministic break: the fixture's custom `instances` map replaces the default fleet and omitted `computer: { driver: "boxAgent" }`, so `cloudSupported` was false and "Team default" was unreachable. The verification config now registers the box computer instance when the Box fixture API is enabled.
- **bot-tools**: two `expect.poll`s had the default 1s timeout; under load the expanded usage panel took longer. Both now poll for 10s.

## delta-context delegation ordering

Two tests relied on a delegated reply landing while a specific turn was running, but under load the wake dispatch could lag past both replies. The fixture can now hold a depth-capped delegated turn's reply on a gate file (`holdDelegation`, supported by `room-handoff-agent.ts`), so each test controls exactly when the reply lands. One launch assertion also polled for the async launch record instead of reading it once.

## Other

- `cli.test.ts` Caddy cleanup read `caddy.pid` unconditionally; when the server failed before reaching the Caddy step, cleanup threw ENOENT and masked the real failure. It now guards with `existsSync`.
- `claude.test.ts` dispose test got a 60s timeout for its temp-dir teardown under load.
- chat-mcp-tools and claude-login-auth `waitFor`/`poll` timeouts raised to 10s for fixture boot under load.

Validation: lint, typecheck, and each touched suite pass in this tree; the previously flaking pairs pass back-to-back runs.

## Summary by CodeRabbit

- **Tests**
  - Improved reliability of authentication and installation scenarios under slower startup conditions.
  - Added safeguards for temporarily unavailable call logs and short-lived authentication links.
  - Expanded coverage for delayed termination, delegated replies, session handoffs, and cancellation behavior.
  - Improved browser test stability by waiting for rendered UI states before continuing interactions.
  - Added test coverage for computer-enabled verification fixtures and thread expansion scenarios.
- **Reliability**
  - Extended process-shutdown grace periods to better accommodate slower machines.


milind-soni:
Resolved conflicts with main (f6f749c6) and pushed merge commit baeb142a. Preserved main’s stronger thinking-timer readiness/cleanup and Codex auth windows; retained this PR’s deterministic delegation gates, installer startup fix, and MCP shutdown grace. GitHub now reports MERGEABLE (no conflicts).

Local macOS verification: 300 tests passed, 1 skipped across the 7 affected server suites, the shared room-handoff helper suite, and all 3 touched renderer fixtures. Lint, typecheck, locale validation, and diff checks passed. UI/server workflow checks used disposable fake-engine fixtures, not live user data.

GitHub CI is running. The Vercel preview separately requests authorization; no deployment settings were changed. PR remains open.

PR #1596, opened 2026-09-20
fix(acp): fail a turn whose agent goes silent mid-answer (OpenCode hang)

### LINKED ISSUE #1595: OpenCode (ACP) turns hang forever when the agent goes silent mid-answer
## Symptom
A turn against the OpenCode engine (`opencode-go` / "OpenCode gateway") can spin indefinitely: the thinking bubbles stream for a while, then the turn stops advancing with no result and no error. The only way out is to cancel manually.

## Root cause
`server/drivers/acp/core.ts` sends `session/prompt` with **no timeout**. In ACP a turn is one long-lived RPC that resolves exactly once, at the very end; everything before it arrives as `session/update` notifications. The protocol has no heartbeat or "still working" signal, so **total silence is ambiguous**: it can mean "the model paused mid-think" or "the agent is wedged and will never resolve the prompt." Nothing on the driver side ever gives up: the per-request `timeoutMs` is only wired to the setup calls (`initialize`, `session/new`, `session/load`, `session/set_config_option`); the long-lived prompt intentionally has none, because it can legitimately stream for minutes.

The silence itself can come from three distinct failures, indistinguishable on the wire:
1. **Not broken, just slow** — the upstream stops emitting for tens of seconds while still working. A production log shows gaps of 33s/49s/56s between thought chunks on a turn that used ~68k input tokens, took 7.5 minutes, and *did* complete normally (`stopReason: "end_turn"`).
2. **A dependency is hung** — OpenCode blocked on a tool, MCP server, subagent, or permission wait that never returns.
3. **The agent host itself is wedged** — process alive but stuck on an unresolved promise or stalled I/O; it will never resolve the prompt.

Because all three look identical without a heartbeat, the fix must be an *idle* timeout (deadline that restarts on every inbound line), not a fixed wall-clock one — a fixed deadline would false-positive on the verified 7.5-minute healthy turn above.

## Expected behavior
A turn whose agent produces nothing at all for a configurable window should fail with a clear `runtime.error`, settle as a failed turn, and close the wedged agent process so the next turn spawns a fresh one — instead of leaving the user staring at an eternal spinner.

## Notes
- Reproduction harness exists: a new `FAKE_ACP_MODE=stall-after-text` for the driver fake models the *permanent*-silence wedge (stream one chunk, then silence forever). Guard-on this fails fast with `rpc_error`; guard-off it hangs until manually interrupted.
- Mitigation: set `OPENMAUS_ACP_PROMPT_IDLE_TIMEOUT_MS` to a non-zero value once the fix ships.

johnsonAyo:
Fixes #1595

## Problem
`server/drivers/acp/core.ts` sends `session/prompt` with no timeout — the one ACP call that legitimately streams for minutes, so it has none by design. ACP turns resolve exactly once at the end; between requests there is no heartbeat, so full silence is ambiguous: it can be a slow pause mid-think or a wedged agent that will never return the prompt result (blocked on a hung tool/MCP/subagent, or stuck inside the agent host). Nothing on the driver side gives up, so a wedged agent leaves the user staring at an eternal spinner.

## Fix
Add an **idle** watchdog (not a wall-clock deadline) to the `session/prompt` request:
- The deadline restarts on every inbound line from the agent, so long streaming turns are never cut off — only total silence trips it. A verified production slow turn (7.5 min, ~68k input tokens, 30–56s think gaps) completed fine and would not trigger the default window.
- On trip, the pending prompt RPC rejects; the turn catch emits a descriptive `runtime.error`, settles `rpc_error`, **and closes the wedged child** (`closeSession(threadId, "prompt-stall")`) so the pooled session does not poison the next turn.
- Env-tunable: `OPENMAUS_ACP_PROMPT_IDLE_TIMEOUT_MS` (default `180000`, ms). Set it to `0` to restore the previous hang-forever behavior explicitly.

## Test
New `FAKE_ACP_MODE=stall-after-text` for the driver fake models the *permanent*-silence wedge (stream one `agent_message_chunk`, then silence forever). The new driver test asserts the turn completes `ok:false` with `stopReason: rpc_error`, a `runtime.error` mentioning the guard + env var, the streamed delta still reaching the UI, and the thread session settling (session closed).

Verified before/after against the same fixture: with the guard disabled (`OPENMAUS_ACP_PROMPT_IDLE_TIMEOUT_MS=0`, pre-fix behavior) the turn spins after its chunk and only settles via manual interrupt; with the guard at 150ms it fails fast with `rpc_error` and the session closes.

`pnpm vitest run server/drivers/acp/` → 222 passed (including this test). `pnpm typecheck` and `pnpm lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented conversations from hanging indefinitely when an agent stops responding mid-answer.
  * Preserved partial responses when an agent stalls after sending text.
  * Prevented idle timeouts while an agent is awaiting user approval.
  * Automatically closes stalled agent sessions so subsequent turns can start normally.
  * Added an idle timeout for silent prompt responses, with configuration support and an option to disable it.


milind-soni:
Review fixes pushed as 9ec04ede: suspend the idle watchdog while a human approval is pending, reset on the response, clear the current timer on completion/cleanup, and evict a stalled process before completion listeners can start a replacement turn. Added an approval-wait regression. All 223 ACP tests, typecheck, lint and locale checks pass. The isolated Qwen end-to-end fixture passes model routing, pooled second turns and rejected-session reload recovery. These are synthetic CLI checks, not paid-provider calls. Previous Linux CI failure was an Electron download HTTP 500; fresh CI is running.

PR #1619, opened 2026-09-21
Render Mermaid diagrams in chat messages

### LINKED ISSUE #1617: Render Mermaid diagrams in chat messages
### Problem

Models answer with ```mermaid fences all the time — architecture flows, sequence diagrams, timelines — and today ChatMarkdown renders them as an ordinary highlighted code block. The reader has to copy the source into mermaid.live to see the picture. The repo's own README leans on a Mermaid diagram for exactly this kind of explanation, so chat is the odd surface out.

### Proposal

Fenced blocks tagged `mermaid` render as diagrams inside the chat bubble:

- Lazy-load mermaid only when a `mermaid` fence actually appears in a message, so chats without diagrams pay nothing (same pattern as the existing lazy Shiki import).
- Render through `mermaid.render()` with `securityLevel: "strict"`; the returned SVG is the only thing injected, and model HTML everywhere else keeps rendering as text (no `rehype-raw`).
- Follow the active skin through `--code-color-scheme`, the same token the Shiki code blocks read, and re-render when the skin changes.
- Stream-safe: while the message is still streaming, keep showing the source; render once the block holds still; on a parse error, fall back to the source with the error above it instead of a broken bubble.
- Keep the source one click away (show/copy) so diagrams stay inspectable.

### Follow-ups (not required for the first pass)

- Rendered diagrams in exported transcripts (they can keep exporting as code, which stays readable).

bradhallett:
Closes #1617.

## Problem

Models answer with ```mermaid fences all the time — architecture flows, sequence diagrams, timelines — and ChatMarkdown rendered them as ordinary highlighted code, so the reader had to paste the source into mermaid.live to see anything.

## Change

Fenced blocks tagged `mermaid` now render as diagrams inside the chat bubble through a new `MermaidDiagram` component beside the existing `CodeBlock`:

- **Lazy by default.** `mermaid` is dynamically imported only when a `mermaid` fence appears, the same pattern as the lazy Shiki import. The production build splits it into on-demand chunks (`mermaid.core`, per-diagram chunks) that never load for diagram-free chats.
- **Renderer security posture unchanged.** Rendering goes through `mermaid.render()` with `securityLevel: "strict"` (labels sanitized, click callbacks dropped outside `loose`), `startOnLoad: false` and `suppressErrorRendering: true`. The returned SVG is the only injected markup; model HTML everywhere else still renders as text — no `rehype-raw`.
- **Skin-aware.** The theme reads the nearest skin's `--code-color-scheme` (the same token the Shiki blocks use) and re-renders when `data-skin` changes; both schemes stay cached per content hash, like Shiki's dual palette.
- **Stream-safe.** A streaming block keeps showing its source until the content holds still (the same `STREAM_SETTLE_MS` debounce highlighting uses), so a half-arrived diagram never flashes a parse error mid-stream.
- **Failure keeps the text.** A diagram that fails to parse falls back to its source with the parse error above it. Every frame carries show-source and copy controls, so diagrams stay inspectable.

Dependency note (per the contribution rules): this adds one runtime dependency, `mermaid` ^12.0.0 (MIT) — the reference implementation of the language — loaded only on demand.

## Screenshots

Before (main, Midnight skin — source as an ordinary code block):

before

After — flowchart (Midnight):

after flowchart

After — sequence diagram (Midnight):

after sequence

After — broken source keeps its text with the parse error:

after error fallback

After — Atelier (light skin) follows `--code-color-scheme`:

after light

## Verification

- `pnpm exec vitest run src/components/ChatMarkdown.test.ts` — 58 passed, including 4 new tests: fence routing to the diagram frame, case-insensitive fence tag, ordinary fences staying on the highlighter path, and strict-mode initialization with repeat views served from cache.
- `pnpm typecheck`, `pnpm lint`, `pnpm i18n:check` — clean (also re-run by the pre-push hook).
- `pnpm exec vite build` — passes; mermaid splits into lazy chunks.
- Screenshots above were taken from the real component rendered in Chromium against both a dark and a light skin, including the parse-error path.

Comparison base: upstream `main` @ `e87f488`. Head: `bradhallett:feat/chat-mermaid-diagrams`.

## Summary by CodeRabbit

- **New Features**
  - Added support for rendering Mermaid diagrams in chat messages.
  - Diagrams automatically adapt to light and dark themes.
  - Diagram rendering is optimized during streaming and across repeated views.
  - Users can reveal or copy the Mermaid source.
  - Rendering errors provide a clear fallback with the original diagram source.
  - Standard code blocks continue to display with their existing formatting and controls.


bradhallett:
Also ran the full renderer suite on this branch:  — 198 files, 1842 tests, all passing.

